PR #96, opened 2026-05-05
genai: add `gen_ai.token.cache` and `gen_ai.token.reasoning` metric attributes

### LINKED ISSUE #76: GenAI: Add metrics for detailed token usage (cache, reasoning)
Should these be recorded in metrics as well? Assuming yes, what should the gen_ai.token.type be?

_Originally posted by @stephentoub in https://github.com/open-telemetry/semantic-conventions/pull/3163#discussion_r2732218612_

Nik-Reddy:
## Description

Adds two attributes to `gen_ai.client.token.usage` to break down token counts:

- `gen_ai.token.cache` (enum: `read`, `creation`, `uncached`) – cache status of input tokens
- `gen_ai.token.reasoning` (boolean) – whether output tokens were used for reasoning

When a provider gives a breakdown, instrumentations partition token counts so data points sum to the total. When no breakdown is available, bare `input`/`output` data points are recorded as before.

Metric attributes map to span attributes on the same request:

| Metric attribute            | Equals (per request)                                              |
|-----------------------------|-------------------------------------------------------------------|
| cache=read                  | gen_ai.usage.cache_read.input_tokens                              |
| cache=creation              | gen_ai.usage.cache_creation.input_tokens                          |
| cache=uncached              | gen_ai.usage.input_tokens − cache_read − cache_creation           |
| reasoning=true              | gen_ai.usage.reasoning.output_tokens                              |
| reasoning=false             | gen_ai.usage.output_tokens − gen_ai.usage.reasoning.output_tokens |

Ported from open-telemetry/semantic-conventions#3624 which was auto-closed when GenAI conventions moved to this repo.

Fixes #76

## Motivation

**User journey:** cost allocation and capacity planning.

- `gen_ai.token.cache` – are we saving money from prompt caching? What fraction of input tokens hit the cache vs fresh computation?
- `gen_ai.token.reasoning` – how much of our output token spend goes to reasoning overhead vs actual content?

Right now teams only see input/output totals and have to check provider dashboards for the split.

**Prior art:**
- OpenAI: `usage.prompt_tokens_details.cached_tokens`, `usage.completion_tokens_details.reasoning_tokens`
- Anthropic: `usage.cache_read_input_tokens`, `usage.cache_creation_input_tokens`
- DeepSeek: `usage.reasoning_tokens` separate from `usage.completion_tokens`
- Span-level attributes (`gen_ai.usage.cache_read.input_tokens`, etc.) already exist for per-request data; this extends the breakdown to the histogram metric for aggregate dashboards.

## Prototype

Cache and reasoning breakdowns come straight from existing provider response fields:

| Provider | Cache source | Reasoning source |
|---|---|---|
| OpenAI | `usage.prompt_tokens_details.cached_tokens` | `usage.completion_tokens_details.reasoning_tokens` |
| Anthropic | `usage.cache_read_input_tokens`, `usage.cache_creation_input_tokens` | n/a |
| DeepSeek | n/a | `usage.reasoning_tokens` |

No separate prototype implementation added — reference scenarios don't exercise metric histogram recording yet, so the provider field mapping above shows feasibility.

## Checklist

- [ ] Reference instrumentation and scenarios updated — n/a, this is a spec-only change; no scenario exercises metric histogram recording yet


trask:
thanks @Nik-Reddy! can you update the PR description and fill out the new [PR template](https://raw.githubusercontent.com/open-telemetry/semantic-conventions-genai/refs/heads/main/.github/PULL_REQUEST_TEMPLATE.md)?

model/gen-ai/metrics.yaml
```diff
@@ -23,13 +23,33 @@ metrics:
...
+      subsets of the total `input` or `output` token counts and MUST NOT be summed with
+      them to avoid double counting.
+
+      For example, a response with 100 input tokens (50 cached) and 200 output tokens (150 reasoning)
+      SHOULD produce:
+
+      - `gen_ai.client.token.usage{gen_ai.token.type="input"}` = 100
+      - `gen_ai.client.token.usage{gen_ai.token.type="input", gen_ai.token.cache="read"}` = 50
+      - `gen_ai.client.token.usage{gen_ai.token.type="output"}` = 200
+      - `gen_ai.client.token.usage{gen_ai.token.type="output", gen_ai.token.reasoning=true}` = 150
```
trask: I think this modeling is in conflict with https://github.com/open-telemetry/semantic-conventions/blob/main/docs/general/metrics.md#general-guidelines

> As a rule of thumb, aggregations over all the attributes of a given metric SHOULD be meaningful

alexmojaki: looks like it drifted from https://github.com/open-telemetry/semantic-conventions/pull/3624

Nik-Reddy: > I think this modeling is in conflict with https://github.com/open-telemetry/semantic-conventions/blob/main/docs/general/metrics.md#general-guidelines
> 
> > As a rule of thumb, aggregations over all the attributes of a given metric SHOULD be meaningful

@trask reverted to the partition model from the original https://github.com/open-telemetry/semantic-conventions/pull/3624 I worked previously, where data points sum to the total.

Nik-Reddy: > looks like it drifted from [open-telemetry/semantic-conventions#3624](https://github.com/open-telemetry/semantic-conventions/pull/3624)

@alexmojaki Thank you for flagging, fixed it now. The note and registry entries now precisely match #3624. Additionally, added partition example's missing uncached enum value.

model/gen-ai/metrics.yaml
```diff
@@ -23,13 +23,36 @@ metrics:
...
     stability: development
+    note: |
+      When the provider reports a detailed token breakdown, instrumentations
+      SHOULD partition the token count across the relevant attribute values
+      instead of emitting a separate total. The partitioned data points MUST
+      sum to the total token count for that token type, and a bare data point
+      with only `gen_ai.token.type` set MUST NOT be emitted alongside them.
+
+      When no breakdown is available, a single data point with only
+      `gen_ai.token.type` set SHOULD be recorded.
```
alexmojaki:
```suggestion
      When no detailed breakdown is available, two data points with only
      `gen_ai.token.type` set to `input` and `output` SHOULD be recorded.
```

model/gen-ai/metrics.yaml
```diff
@@ -23,13 +23,36 @@ metrics:
...
+
+      For example, a response with 100 input tokens (50 cache-read, 50 uncached)
+      and 200 output tokens (150 reasoning, 50 non-reasoning) SHOULD produce:
+
+      - `gen_ai.client.token.usage{gen_ai.token.type="input", gen_ai.token.cache="read"}` = 50
+      - `gen_ai.client.token.usage{gen_ai.token.type="input", gen_ai.token.cache="uncached"}` = 50
+      - `gen_ai.client.token.usage{gen_ai.token.type="output", gen_ai.token.reasoning=true}` = 150
+      - `gen_ai.client.token.usage{gen_ai.token.type="output", gen_ai.token.reasoning=false}` = 50
```
trask: I think correlating the metric attributes to the span attributes would be helpful, and may not need this example then.

| Metric attribute            | Equals (per request)                                              |
|-----------------------------|-------------------------------------------------------------------|
| cache=read                  | gen_ai.usage.cache_read.input_tokens                              |
| cache=creation              | gen_ai.usage.cache_creation.input_tokens                          |
| cache=uncached              | gen_ai.usage.input_tokens − cache_read − cache_creation           |
| reasoning=true              | gen_ai.usage.reasoning.output_tokens                              |
| reasoning=false             | gen_ai.usage.output_tokens − gen_ai.usage.reasoning.output_tokens |

Nik-Reddy: @trask @alexmojaki Thanks, both suggestions make sense. Will update the note wording and replace the example with the correlation table mapping to  span attributes.

model/gen-ai/metrics.yaml
Nik-Reddy: sounds good, will update.

model/gen-ai/metrics.yaml
Nik-Reddy: @trask updated it aligned to the span attributes

model/gen-ai/metrics.yaml
```diff
@@ -23,13 +23,39 @@ metrics:
...
       code_generation:
         metric_value_type: int
     attributes:
       - ref_group: metric_attributes.gen_ai
       - ref: gen_ai.token.type
         requirement_level: required
+      - ref: gen_ai.token.cache
+        requirement_level:
+          conditionally_required: when the provider reports a cache breakdown for input tokens.
+      - ref: gen_ai.token.reasoning
```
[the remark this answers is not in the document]
alexmojaki: Yeah this seems good to clarify

model/gen-ai/metrics.yaml
```diff
@@ -23,13 +23,39 @@ metrics:
...
+        requirement_level:
+          conditionally_required: when the provider reports reasoning token usage for output tokens.
```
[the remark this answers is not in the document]
trask: @Nik-Reddy heads up it looks like you may have applied this autofix suggestion but then force pushed over it below

Mandark-droid:
+1 on this PR — the mapping table cleanly extends what's already in the registry on the span side, and the cache / reasoning enums match every provider's API field semantics we've encountered.

## Span-side reference impl already shipping

We just released [`genai-otel-instrument` v1.3.1](https://github.com/Mandark-droid/genai_otel_instrument/releases/tag/v1.3.1) (PyPI: [`pip install genai-otel-instrument==1.3.1`](https://pypi.org/project/genai-otel-instrument/1.3.1/)) which emits all three span attributes referenced in your mapping table — names verbatim, including the `cache=uncached` and `reasoning=false` derivation rules you've documented:

```
gen_ai.usage.cache_read.input_tokens       ← Anthropic + OpenAI prompt cache hits
gen_ai.usage.cache_creation.input_tokens   ← Anthropic prompt cache writes
gen_ai.usage.reasoning.output_tokens       ← OpenAI o1/o3 reasoning tokens
```

(See commit [`1822d5f`](https://github.com/Mandark-droid/genai_otel_instrument/commit/1822d5f).) Caught a naming drift in our own implementation while reviewing this PR — we had originally used `gen_ai.usage.reasoning_tokens` (a legacy short form). Fixed to match the registry-canonical `gen_ai.usage.reasoning.output_tokens` before the v1.3.1 release tag landed, so consumers see the upstream-aligned name on day one.

The PR's mapping has been useful as a producer-side spec — it's the cleanest single document of the cache/reasoning attribute family. Two suggestions inline:

## Two small comments on the PR

### 1. Confirm `cache=uncached` is provider-derivable

The table implies the producer computes `cache=uncached` as `gen_ai.usage.input_tokens − cache_read − cache_creation`. This works for Anthropic (their `usage.input_tokens` is the post-cache count and the cache fields are separate) and OpenAI (their `prompt_tokens` is total including cache, and `prompt_tokens_details.cached_tokens` is the cache-read subset).

But: OpenAI's reported `prompt_tokens` already includes the cached portion. So `uncached = prompt_tokens − cached_tokens` (no `cache_creation` term needed there). Anthropic does it differently — their `usage.input_tokens` is the **un-cached** portion, so `total_input = input_tokens + cache_read + cache_creation` and `uncached = input_tokens` directly.

In both cases the derivation works as the PR's table describes — but only because the producer is doing the right per-provider math before partitioning. Worth a `[!NOTE]` in the spec calling that out, because a naive consumer trying to validate "sum of partitioned data points equals the bare `input` data point" will be confused.

### 2. DeepSeek `usage.reasoning_tokens`

The prior-art table cites `DeepSeek: usage.reasoning_tokens separate from usage.completion_tokens`. That's correct as a flat field, but DeepSeek's API exposes it on chat completions through their OpenAI-compatible endpoint. Instrumentations using a generic OpenAI SDK against `api.deepseek.com` will see `completion_tokens_details.reasoning_tokens` in the response, not a flat `usage.reasoning_tokens`. The mapping the PR proposes still works — just worth noting that the field name varies by access path (raw DeepSeek client vs. OpenAI-compat). We treat both the same way: the value lands under `completion_tokens_details.reasoning_tokens` on the response object and maps to `gen_ai.usage.reasoning.output_tokens` on the span.

## Library-side rollout plan

Once #96 merges, we'll do a small follow-up PR on our side that adds the `gen_ai.token.cache` and `gen_ai.token.reasoning` dimensions to the `gen_ai.client.token.usage` histogram recording in [`base.py`](https://github.com/Mandark-droid/genai_otel_instrument/blob/main/genai_otel/instrumentors/base.py) — partitioning per the table you've written here. Happy to publish a `reference/scenarios/` flow that demonstrates a partitioned histogram against the mock LLM if that helps anchor the spec.

Closes the producer-side gap for [#76](https://github.com/open-telemetry/semantic-conventions-genai/issues/76).

Nik-Reddy:
> +1 on this PR — the mapping table cleanly extends what's already in the registry on the span side, and the cache / reasoning enums match every provider's API field semantics we've encountered.
> 
> ## Span-side reference impl already shipping
> We just released [`genai-otel-instrument` v1.3.1](https://github.com/Mandark-droid/genai_otel_instrument/releases/tag/v1.3.1) (PyPI: [`pip install genai-otel-instrument==1.3.1`](https://pypi.org/project/genai-otel-instrument/1.3.1/)) which emits all three span attributes referenced in your mapping table — names verbatim, including the `cache=uncached` and `reasoning=false` derivation rules you've documented:
> 
> ```
> gen_ai.usage.cache_read.input_tokens       ← Anthropic + OpenAI prompt cache hits
> gen_ai.usage.cache_creation.input_tokens   ← Anthropic prompt cache writes
> gen_ai.usage.reasoning.output_tokens       ← OpenAI o1/o3 reasoning tokens
> ```
> 
> (See commit [`1822d5f`](https://github.com/Mandark-droid/genai_otel_instrument/commit/1822d5f).) Caught a naming drift in our own implementation while reviewing this PR — we had originally used `gen_ai.usage.reasoning_tokens` (a legacy short form). Fixed to match the registry-canonical `gen_ai.usage.reasoning.output_tokens` before the v1.3.1 release tag landed, so consumers see the upstream-aligned name on day one.
> 
> The PR's mapping has been useful as a producer-side spec — it's the cleanest single document of the cache/reasoning attribute family. Two suggestions inline:
> 
> ## Two small comments on the PR
> ### 1. Confirm `cache=uncached` is provider-derivable
> The table implies the producer computes `cache=uncached` as `gen_ai.usage.input_tokens − cache_read − cache_creation`. This works for Anthropic (their `usage.input_tokens` is the post-cache count and the cache fields are separate) and OpenAI (their `prompt_tokens` is total including cache, and `prompt_tokens_details.cached_tokens` is the cache-read subset).
> 
> But: OpenAI's reported `prompt_tokens` already includes the cached portion. So `uncached = prompt_tokens − cached_tokens` (no `cache_creation` term needed there). Anthropic does it differently — their `usage.input_tokens` is the **un-cached** portion, so `total_input = input_tokens + cache_read + cache_creation` and `uncached = input_tokens` directly.
> 
> In both cases the derivation works as the PR's table describes — but only because the producer is doing the right per-provider math before partitioning. Worth a `[!NOTE]` in the spec calling that out, because a naive consumer trying to validate "sum of partitioned data points equals the bare `input` data point" will be confused.
> 
> ### 2. DeepSeek `usage.reasoning_tokens`
> The prior-art table cites `DeepSeek: usage.reasoning_tokens separate from usage.completion_tokens`. That's correct as a flat field, but DeepSeek's API exposes it on chat completions through their OpenAI-compatible endpoint. Instrumentations using a generic OpenAI SDK against `api.deepseek.com` will see `completion_tokens_details.reasoning_tokens` in the response, not a flat `usage.reasoning_tokens`. The mapping the PR proposes still works — just worth noting that the field name varies by access path (raw DeepSeek client vs. OpenAI-compat). We treat both the same way: the value lands under `completion_tokens_details.reasoning_tokens` on the response object and maps to `gen_ai.usage.reasoning.output_tokens` on the span.
> 
> ## Library-side rollout plan
> Once #96 merges, we'll do a small follow-up PR on our side that adds the `gen_ai.token.cache` and `gen_ai.token.reasoning` dimensions to the `gen_ai.client.token.usage` histogram recording in [`base.py`](https://github.com/Mandark-droid/genai_otel_instrument/blob/main/genai_otel/instrumentors/base.py) — partitioning per the table you've written here. Happy to publish a `reference/scenarios/` flow that demonstrates a partitioned histogram against the mock LLM if that helps anchor the spec.
> 
> Closes the producer-side gap for [#76](https://github.com/open-telemetry/semantic-conventions-genai/issues/76).

@Mandark-droid Thanks for the detailed feedback and for validating the mapping against your instrumentation.  Both points are good, the provider-specific derivation nuance and the DeepSeek access path. I'll keep this PR focused on the scoping clarifications @trask and @alexmojaki flagged, and we can address the derivation note in a follow-up

model/gen-ai/metrics.yaml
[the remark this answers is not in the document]
Nik-Reddy: Thanks for the heads-up @trask, the autofix landed on the right attribute but the condition wasn't scoped to `token.type`. I've rebased  and updated both conditionally_required conditions to explicitly gate on `gen_ai.token.type` being input / output, and added corresponding MUST only be set constraints in the registry notes.

Nik-Reddy:
@alexmojaki @trask, Pushed scoping clarifications per feedback:
   - conditionally_required for `gen_ai.token.cache` now explicitly gates on `gen_ai.token.type = input`
  - conditionally_required for `gen_ai.token.reasoning` now explicitly gates on `gen_ai.token.type = output`
  - Registry notes add MUST only be set applicability constraints for both attributes and rebased onto current main
 
 Substantive changes are only in `model/gen-ai/metrics.yaml` and `model/gen-ai/registry.yaml`; remaining diffs are regenerated docs/snapshot.

model/gen-ai/metrics.yaml
```diff
@@ -23,13 +23,39 @@ metrics:
...
+      `gen_ai.token.type` set to `input` and `output` SHOULD be recorded.
+
+      Each partitioned metric attribute value corresponds to span attributes
+      on the same request:
+
+      | Metric attribute            | Equals (per request)                                              |
+      |-----------------------------|-------------------------------------------------------------------|
+      | cache=read                  | gen_ai.usage.cache_read.input_tokens                              |
+      | cache=creation              | gen_ai.usage.cache_creation.input_tokens                          |
+      | cache=uncached              | gen_ai.usage.input_tokens − cache_read − cache_creation           |
```
singankit: Uncached tokens, would this be calculated by instrumenting libs? If yes, I am guessing this could differ from provider to provider? Any provider specific guidelines would be helpful at least for the ones mentioned in PR description

model/gen-ai/registry.yaml
```diff
@@ -218,6 +218,57 @@ attributes:
...
+      sum to the total input token count. Instrumentations MUST NOT emit a
+      bare `gen_ai.token.type=input` data point alongside cache-partitioned
+      data points for the same request.
+
+      When the provider does not report a cache breakdown, omit this
+      attribute entirely and record a single `gen_ai.token.type=input`
+      data point with the total count.
+
+      Partition members with a value of zero MAY be omitted.
+    examples: ['read', 'creation', 'uncached']
```
singankit, line 252: Would appreciate any sample examples on how this breakdown can be done for say OpenAI which reports `cached_tokens` but does not provide granular breakdown or read vs creation

model/gen-ai/metrics.yaml
Nik-Reddy: @singankit Yes, uncached is derived by the instrumentation. The provider mapping is in the PR description, e.g. OpenAI: `uncached = prompt_tokens − cached_tokens`. The registry note also  says **partition members with a value of zero MAY be omitted**, so providers without a cache creation concept (like OpenAI) simply omit  `cache=creation`

model/gen-ai/registry.yaml
Nik-Reddy: For OpenAI: cache=read -> `cached_tokens`, cache=uncached -> `prompt_tokens − cached_tokens`. OpenAI doesn't report cache creation, so cache=creation is omitted per the 'MAY be  omitted' rule. Anthropic has all three: `cache_read_input_tokens`, `cache_creation_input_tokens`, and `input_tokens` (which is already the uncached portion)

docs/gen-ai/mcp.md
```diff
@@ -158,21 +158,21 @@ option to enable it.
...
 | [`gen_ai.tool.name`](/docs/registry/attributes/gen-ai.md) | ![Development](https://img.shields.io/badge/-development-blue) | `Conditionally Required` When operation is related to a specific tool. | string | Name of the tool utilized by the agent. | `Flights` |
-| [`jsonrpc.request.id`](https://github.com/open-telemetry/semantic-conventions/blob/v1.41.0/docs/registry/attributes/jsonrpc.md) | ![Development](https://img.shields.io/badge/-development-blue) | `Conditionally Required` When the client executes a request. | string | A string representation of the `id` property of the request and its corresponding response. [2] | `10`; `request-7` |
+| [`jsonrpc.request.id`](https://github.com/open-telemetry/semantic-conventions/blob/v1.41.1/docs/registry/attributes/jsonrpc.md) | ![Development](https://img.shields.io/badge/-development-blue) | `Conditionally Required` When the client executes a request. | string | A string representation of the `id` property of the request and its corresponding response. [2] | `10`; `request-7` |
```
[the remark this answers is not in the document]
Nik-Reddy: Those links are in auto-generated content from the upstream baseline, they'll update when upstream refreshes their templates. Not introduced by this PR

model/gen-ai/metrics.yaml
```diff
@@ -23,13 +23,39 @@ metrics:
     unit: "{token}"
     brief: 'Number of input and output tokens used.'
     stability: development
+    note: |
+      When the provider reports a detailed token breakdown, instrumentations
+      SHOULD partition the token count across the relevant attribute values
+      instead of emitting a separate total. The partitioned data points MUST
+      sum to the total token count for that token type, and a bare data point
+      with only `gen_ai.token.type` set MUST NOT be emitted alongside them.
```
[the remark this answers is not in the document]
Nik-Reddy: updated the note to scope the rule per token type: **MUST NOT be emitted alongside the partitioned data points for the same token type** 

A provider reporting only cache breakdown (not  reasoning) still emits a bare `token.type=output` data point as expected.

Nik-Reddy:
@trask @alexmojaki circling back on this. I addressed the scoping feedback from both of you: `conditionally_required` now gates on the specific `token.type`, and the registry notes have MUST applicability constraints for both attributes. CI is green. Would appreciate a re-review when you get a chance.

model/gen-ai/metrics.yaml
```diff
@@ -23,13 +23,42 @@ metrics:
...
     annotations:
       code_generation:
         metric_value_type: int
     attributes:
       - ref_group: metric_attributes.gen_ai
       - ref: gen_ai.token.type
         requirement_level: required
+      - ref: gen_ai.token.cache
+        requirement_level:
+          conditionally_required: when `gen_ai.token.type` is `input` and the provider reports a cache breakdown.
```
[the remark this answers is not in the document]
trask: I think this is ok, as along as the metric attributes are defined in terms of span attributes which are exercised

model/gen-ai/metrics.yaml
```diff
@@ -23,13 +23,42 @@ metrics:
...
+      type, instrumentations SHOULD partition the token count for that type
+      across the relevant attribute values instead of emitting a separate
+      total. The partitioned data points MUST sum to the total token count
+      for that token type, and a bare data point with only
+      `gen_ai.token.type` set MUST NOT be emitted alongside the partitioned
+      data points for the same token type.
+
+      When no detailed breakdown is available for a given token type, a
+      single data point with only `gen_ai.token.type` set SHOULD be recorded
+      for that type.
```
trask, line 37: I think this raises an interesting metric modeling question worth discussing. I've added it to this week's SIG agenda. cc @lmolkova

`gen_ai.token.type` is required, so `gen_ai.client.token.usage` is defined as a histogram of token counts for a known token type (`input` or `output`). That works well for questions like "what is the distribution of input vs output token counts?"

The cache and reasoning attributes are different because they are conditional breakdowns within one of those token types. For example, suppose a request has 100 input tokens: 40 cached and 60 uncached. Without cache partitioning, the input-token histogram records one measurement:

```text
gen_ai.token.type=input  value=100
```

With the proposed cache dimension, that becomes two measurements:

```text
gen_ai.token.type=input, gen_ai.token.cache=read      value=40
gen_ai.token.type=input, gen_ai.token.cache=uncached  value=60
```

If a backend or query aggregates away `gen_ai.token.cache`, the histogram `sum` is still `100`. But `count` is now `2`, and the buckets, min/max, and percentile views describe cache partition sizes rather than input token counts.

This is different from a histogram like HTTP client duration, where attributes classify a single duration measurement. Here, the optional attribute partitions one token count into multiple measurements.

If this were a counter instead of a histogram, the parts would simply add back up when the optional breakdown dimension is aggregated away.

I think there are (at least) a couple of options:

1. Reconsider whether `gen_ai.client.token.usage` should be a histogram or a counter. This avoids the partitioning issue but gives up the per-request distribution view.
2. Keep `gen_ai.client.token.usage` as the histogram for the required token type split, and add a separate counter metric for additive cache/reasoning breakdowns.

alexmojaki: Well, what if i want statistics about the total number of tokens without the input/output split? Isn't that also lost? Arguably that number is quite important because it represents how close you are to filling the context window.

Nik-Reddy: Thanks @trask, the partitioning concern is fair. Hadn't reasoned about how the histogram count and bucket views behave once the optional dimension gets aggregated away.
 
 @alexmojaki +1 on context window queries. The same partitioning behavior already applies to `gen_ai.token.type` today, so any "total tokens per request" view through this histogram has the same limitation regardless of what this PR adds. `sum` still aggregates correctly across attributes (so a context window proximity gauge built from sum keeps working), but `count`, buckets, min/max, and percentiles describe per-type slices, not per-request totals. A p95 of total request tokens isn't directly query able here even today. This PR extends that pattern rather than introduces it.
 
 Both options seem reasonable. Quick observations from the implementer side:
-  Option 1 (switch `gen_ai.client.token.usage` to a counter) is the cleanest fix but is a breaking change for anyone already emitting the histogram, and loses the per-request input vs  output distribution view.
- Option 2 (keep the histogram for the `gen_ai.token.type` split, add a separate counter metric for cache/reasoning breakdowns) is backward compatible, the breakdowns stay additive, and the new counter cleanly carries the optional dimensions. Costs an extra metric, but instrumentation libraries already track these values separately, so emit cost is small.
 
One thing worth flagging given @alexmojaki's point: neither option gives a clean per-request-totals distribution view. If that's a use case the SIG wants to support, a companion per-request totals metric may be worth considering alongside whichever direction is picked.

I'd lean slightly towards option 2 for the compat reason, but either way works for me to incorporate. Will hold off on pushing changes until the SIG lands on a direction. Thanks for putting it on the agenda.

model/gen-ai/metrics.yaml
```diff
@@ -23,13 +23,42 @@ metrics:
...
+      Each partitioned metric attribute value corresponds to span attributes
+      on the same request:
+
+      | Metric attribute            | Equals (per request)                                              |
+      |-----------------------------|-------------------------------------------------------------------|
+      | cache=read                  | gen_ai.usage.cache_read.input_tokens                              |
+      | cache=creation              | gen_ai.usage.cache_creation.input_tokens                          |
+      | cache=uncached              | gen_ai.usage.input_tokens − cache_read − cache_creation           |
+      | reasoning=true              | gen_ai.usage.reasoning.output_tokens                              |
+      | reasoning=false             | gen_ai.usage.output_tokens − gen_ai.usage.reasoning.output_tokens |
```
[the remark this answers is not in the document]
Nik-Reddy: As per @trask's reply, the new metric attributes (gen_ai.token.cache, gen_ai.token.reasoning) are defined in terms of span attributes that the reference scenarios already exercise. This thread already addressed.

model/gen-ai/metrics.yaml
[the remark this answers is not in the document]
Nik-Reddy: The metric attribute values map to existing span attributes in the registry, so the condition holds. Marking as addressed. 
@trask

model/gen-ai/metrics.yaml
alexmojaki: Is the goal of using a counter to avoid confusion? It's just less information, right? Is the thinking that the histogram information is no longer useful enough when it's partitioned this much?

Is this the first time this problem arises? Arguably it applies to basically any histogram with attributes, e.g. HTTP server duration, where you can't get a histogram of durations for just a route or just a status code, only combinations. Seems like something that should be solved by e.g. histogram views.

trask: > Is this the first time this problem arises?

I couldn't find another histogram in semconv that splits a single logical value into labeled parts of the same total.

There are several examples of additive splits, but they use additive instruments:

- `container.cpu.time` and `system.cpu.time` split CPU time by `cpu.mode`
- disk/network I/O counters split bytes or operations by direction
- `system.filesystem.usage` splits bytes by filesystem state.

> Arguably it applies to basically any histogram with attributes, e.g. HTTP server duration, where you can't get a histogram of durations for just a route or just a status code, only combinations. Seems like something that should be solved by e.g. histogram views.

This is solved by histogram views already. If you aggregate away route or status code from `http.server.request.duration`, you still have a histogram of request durations. But if you aggregate away `gen_ai.token.cache`, `sum` gives total tokens while `count` and buckets describe the cache/uncached partition measurements, not original request token counts.

trask: Discussed in the general Semantic Convention SIG meeting today. The recommendation is to convert this metric from histogram to counter and create separate histogram metric(s) if needed. I'll send a separate PR to convert it to a Counter, since it's an issue today even without adding the new attributes, and then we can keep this PR cleaner. Will also still discuss in tomorrow's GenAI SIG meeting

trask: > I'll send a separate PR to convert it to a Counter

actually, maybe we should do it in this PR, since breaking, and potentially better to break once (adding new attributes which split timeseries as this PR does is considered breaking).

alexmojaki: > adding new attributes which split timeseries as this PR does is considered breaking

so maybe now is the time to also add an attribute for modality, e.g. text/audio/image?

Nik-Reddy: Thanks @trask, @alexmojaki. Pushed the rescoped branch
 - `model/gen-ai/metrics.yaml`: `gen_ai.client.token.usage` instrument changed from `histogram` to `counter`. The two new attributes  (`gen_ai.token.cache`, `gen_ai.token.reasoning`) stay as additive  dimensions on the counter.
- `docs/gen-ai/gen-ai-metrics.md`: dropped the now invalid `ExplicitBucketBoundaries` SHOULD line and updated the auto-generated row.
- `docs/gen-ai/openai.md`: refined row updated.
- `schema-snapshot/registry.yaml`: regenerated for the two `token.usage`  entries.
- `CHANGELOG.md`: breaking change entry under Unreleased.

On modality: holding `gen_ai.token.modality` out of this PR. TheSemConv SIG decision was specifically about the histogram-vs-counter shape; the modality attribute is a separate modeling question (enum scope, partition vs label, cardinality of `type x cache x reasoning x modality`), and rolling it in here risks
re-opening the conversation that just closed. Happy to file modality as its own proposal if there is appetite, and to add it in a follow-up PR once the enum lands.

 A companion per-request totals view (a separate histogram metric capturing whole-request token totals, e.g., for SLO/latency-style bucketing) is also a separate question and can be a follow-up if the SIG wants one.

alexmojaki: The SIG decision *implies* that we should consider adding modality here.

model/gen-ai/metrics.yaml
```diff
@@ -19,17 +19,46 @@ attribute_groups:
...
+      Each partitioned metric attribute value corresponds to span attributes
+      on the same request:
+
+      | Metric attribute              | Equals (per request)                                              |
+      |-------------------------------|-------------------------------------------------------------------|
+      | gen_ai.token.cache=read       | gen_ai.usage.cache_read.input_tokens                              |
+      | gen_ai.token.cache=creation   | gen_ai.usage.cache_creation.input_tokens                          |
+      | gen_ai.token.cache=uncached   | gen_ai.usage.input_tokens − cache_read − cache_creation           |
+      | gen_ai.token.reasoning=true   | gen_ai.usage.reasoning.output_tokens                              |
+      | gen_ai.token.reasoning=false  | gen_ai.usage.output_tokens − gen_ai.usage.reasoning.output_tokens |
```
trask: Thinking about this from scratch, wonder if this would be easier to understand:

Span attributes:

- `gen_ai.usage.input_token.total`
- `gen_ai.usage.input_token.cache_read`
- `gen_ai.usage.input_token.cache_creation`
- `gen_ai.usage.output_token.total`
- `gen_ai.usage.output_token.reasoning`

Metric attributes:

- `gen_ai.token.type` = `input | output` (required)
- `gen_ai.token.cache` = read | creation | uncached` (missing if unknown)
- `gen_ai.token.reasoning` = `true | false` (missing if unknown)

or maybe:

- `gen_ai.token.type` = `input | output` (required)
- `gen_ai.token.input_type` = `uncached | cache_read | cache_creation | unknown` (conditionally required when `gen_ai.token.type` = `input`)
- `gen_ai.token.output_type` = `response | reasoning | unknown` (conditionally required when `gen_ai.token.type` = `output`)

(or a variety of similar related options for metric attributes)

I do think it's worth pulling modality into this discussion, because while modality layers easily into metrics, it's unclear how to layer it into span attributes, and we may end up wanting to change token usage to a complex attribute on spans.

And the more aligned the span complex attribute and metric attributes can be the better, so the discussion of span complex attributes could help drive the metric shape (and vice versa).

model/gen-ai/metrics.yaml
Nik-Reddy: Thanks for pushing on modality. I folded it into the design discussion with @trask in the thread below (metric partition shape plus the complex-attribute span direction he raised). Holding the modality patch until that shape settles, since how it slots in depends on which way the partition goes.

model/gen-ai/metrics.yaml
Nik-Reddy: @trask Trying to separate the decisions I see in this thread:
 1. Metric partition shape
 2. Span scalar naming
 3. Whether span usage should move toward a complex attribute
 4. How modality fits across both
 
 For metrics, I see the two sketches as:
 
 **Option 1, current PR shape**
 - `gen_ai.token.type = input | output`
 - `gen_ai.token.cache = read | creation | uncached`, only for input when cache breakdown is known
 - `gen_ai.token.reasoning = true | false`, only for output when reasoning breakdown is known
 
 Keeps cache, reasoning, and eventually modality as separate dimensions. Downside: validity depends on conditional rules, since some dimensions only apply to one token type.
 
 **Option 2, input/output subtype shape** 
 - `gen_ai.token.type = input | output`
 - `gen_ai.token.input_type = uncached | cache_read | cache_creation | unknown`, only for input
 - `gen_ai.token.output_type = response | reasoning | unknown`, only  for output
 
Makes the input/output split more explicit in the attribute names and reduces the chance of invalid combinations. Tradeoff: modality still has to sit outside those subtype attributes because it crosses both input and output.                                                                                                                                                                                                                         
 
So my current read: **option 1 is better if we want independent partition dimensions, option 2 is better if we want the schema to make input and output subtypes more explicit. I do not think modality decides this on its own.**
 
 On the span side, are the names in your sketch meant as a proposed rename from the existing scalar attributes, or mainly as a conceptual shape?
 
 Existing:
 - `gen_ai.usage.cache_read.input_tokens`
 - `gen_ai.usage.cache_creation.input_tokens`
 - `gen_ai.usage.reasoning.output_tokens`
 
 Your sketch:
 - `gen_ai.usage.input_token.cache_read`
 - `gen_ai.usage.input_token.cache_creation`
 - `gen_ai.usage.output_token.reasoning`
 
**If the goal is to keep scalar span attributes, that rename can be decided separately from the metric labels. If the goal is to leave room for a complex span usage attribute, maybe we should decide that first, or at least make sure the metric labels can map cleanly into the eventual complex shape.**
 
For modality, I would avoid locking the value set in this comment. Existing message part schemas use `image | video | audio | document` plus a free-form string branch, while `gen_ai.output.type` has `text | json | image | speech`. Token metrics probably need `text`, but the enum should be aligned deliberately.
 
**I slightly prefer option 1 because it keeps the metric breakdown dimensions independent, but I am not attached to that if the span complex-attribute direction points us toward option 2 or another shape.**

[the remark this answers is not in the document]
Nik-Reddy: Fixed in f8a8295. Expanded the shorthand to the full span attribute names `gen_ai.usage.cache_read.input_tokens`, `gen_ai.usage.cache_creation.input_tokens`) and swapped the Unicode  minus for a regular hyphen on both formula rows. Widened the right  column to fit.

Nik-Reddy:
Read through the[ GenAI SIG notes from 05/19](https://docs.google.com/document/d/1EKIeDgBGXQPGehUigIRLwAUpRGa7-1kXB736EaYuJ2M/edit?tab=t.0) @trask, on the histogram-to-counter discussion. Agreed on the math here - splitting a histogram by `gen_ai.token.type` doesn't compose, and the four-value future partition (`input_cache_read`, `input_cache_write`, `output_reasoning`, `output_generate`) plus a separate modality dimension fits a counter cleanly.

Two things I want to confirm before I rework this PR:

1. Partition shape: does the SIG want a single richer `gen_ai.token.type` enum carrying all four values, or keep two attributes (`cache` + `reasoning`) that compose to the same set? The notes read like one-attribute, but the current PR is two-attribute and I don't want to flip without confirmation.

2. Modality: separate orthogonal attribute (`gen_ai.token.modality`) on the counter, or rolled into `gen_ai.token.type`? Alex raised this above and the notes list it alongside cache/reasoning, but the enum is not pinned. The repo currently has `text | json | image | speech` on `gen_ai.output.type` and `image | video | audio | document` in message part schemas, so the value set is its own discussion.

One open item from the notes worth flagging here so it doesn't get lost: **"are there cases that require histograms, and over which types of tokens."** If any token-related histogram survives, that affects whether this PR fully replaces `gen_ai.client.token.usage` as a histogram or runs a counter alongside.

Once partition shape and modality decision land, I can put up the rewrite (instrument flip, attribute consolidation, modality addition, weaver regen) as one fresh commit on top of f8a8295.

model/gen-ai/metrics.yaml
trask: I tried to sketch out a proposal in https://github.com/open-telemetry/semantic-conventions-genai/pull/197

Nik-Reddy:
Thanks for the fast turnaround on #197. The rewrite looks like it covers the histogram-to-counter pivot plus modality and the cache/phase splits we landed on last week and today. `gen_ai.token.phase` as an enum is cleaner than the boolean I had here.

I'll redirect engagement to #197 and close this once that lands. Two things I want to raise over there as a reviewer: the three-counter split vs a single counter with three partition attributes (total tokens becomes a sum across instruments), and the Anthropic + Cohere V2 mappings (`cache_creation` vs `cache_write`, tool-result token slotting) which I have the most direct context on from #4418/#4421/#56.

Could you add `Closes #96` to the #197 body so the genealogy is explicit?

trask:
Moving to draft while we work through https://github.com/open-telemetry/semantic-conventions-genai/pull/197

trask:

Please move this PR's changelog entries out of `CHANGELOG.md` and into these Towncrier fragments:

1. Create `changelog.d/96.enhancement.md` containing the change log entry, e.g.:

```md
Add `gen_ai.token.cache` and `gen_ai.token.reasoning` attributes to `gen_ai.client.token.usage` for detailed token usage breakdowns.
```

2. Create `changelog.d/96.breaking.md` containing the change log entry, e.g.:

```md
Change `gen_ai.client.token.usage` instrument type from `Histogram` to `Counter`. Splitting one logical token count across optional attributes such as `gen_ai.token.cache` and `gen_ai.token.reasoning` is incompatible with the histogram count/buckets/percentile views once those attributes are aggregated away. A Counter aggregates correctly under that operation. Per-request token distribution use cases should be served by a separate dedicated histogram in a follow-up if needed.
```

After adding the fragments, please remove this PR's direct edit to `CHANGELOG.md`. Towncrier will add the PR links from the fragment filenames when release notes are generated.

Thanks!

TsengSR:
Any news/progress on this PR? the current metrics are really limiting, considering current implementations of the gen_ai conventions.

lmolkova:
this is being solved in #374, PTAL

lmolkova:
this is being addressed in #374, please give it a review

Nik-Reddy:
Sure, @liudmila

trask:
Is this superseded by https://github.com/open-telemetry/semantic-conventions-genai/pull/374?

Issue #35, opened 2025-08-20
Semantic Conventions for Generative AI Agentic Systems (gen_ai.*)


dany-moshkovich:
### Area(s)

area:gen-ai

## Summary  

This proposal introduces semantic conventions for observability of generative AI agentic systems. The conventions define attributes for tracing **tasks, actions, agents, teams, artifacts, and memory** in OpenTelemetry, along with the relationships among them. These conventions are intended to standardize telemetry across complex AI workflows and improve traceability, reproducibility, and analysis.  

---

## Concepts and Relationships  

- **Tasks** – Tasks define the minimal trackable units of work in agentic systems, describing *what needs to be done* (the objective). They can be decomposed into subtasks, each representing a smaller unit of the overall goal. Every task or subtask is expected to be executed through one or more actions.  

- **Actions** – Actions specify *how a task is carried out*. They represent the execution mechanisms such as tool calls, LLM queries, API requests, vector database queries, human input, or workflows, and capture the nondeterministic aspects of the system. Actions fulfill tasks and subtasks and serve as the basic execution units of AI agents.  

- **Agents** – Agents are autonomous, stateful entities that plan, reason, learn, and execute tasks by orchestrating associated actions. They provide the decision-making and control layer, often expressed through workflow actions, and may operate individually or as part of collaborative teams.  

- **Teams** – Teams are dynamic groups of agents collaborating to achieve shared goals. They provide structure for distributing roles, responsibilities, and communication across multiple agents, enabling coordinated execution of complex tasks.  

- **Artifacts** – Artifacts are the tangible inputs and outputs produced or consumed during execution, such as LLM prompts, embeddings, documents, images, or code. They represent the observable products of tasks and actions, linking abstract objectives to concrete data objects that can be shared across contexts through associated memory mechanisms.  

- **Memory** – Memory provides persistent and scoped storage of knowledge and context. It stores and exposes artifacts, allowing agents, tasks, actions, and teams to reuse past information, maintain continuity, and build long-term context across sessions, agents, and tasks.  

---

## Overall Flow  

A task defines an objective, which may be decomposed into subtasks. Tasks and subtasks are executed by actions, which are carried out by agents working individually or in teams. Actions and agents produce or consume artifacts, while memory ensures that artifacts are stored and reused to provide continuity across sessions, tasks, and agents.  

---

## Motivation  

[...]

PR #97, opened 2026-05-05
gen-ai: add plan operation for agent planning spans

Krishnachaitanyakc:
## Description

Adds `plan` as a new `gen_ai.operation.name` value following the same pattern as `execute_tool` — a dedicated operation name because planning has independent duration, error status, and parent-child structure that justify a distinct span. Zero new attributes; only a new enum member reusing existing attributes.

Resubmits open-telemetry/semantic-conventions#3594 to this repository per open-telemetry/semantic-conventions#3696. Partially addresses #35.

### User journey

Incident triage / failure-mode attribution for agent runs.

An agent produces a wrong answer. Was it bad planning (wrong task decomposition) or bad execution (tool returned stale data)? The remediation is different: bad planning needs prompt changes; bad execution needs tool or retrieval fixes.

Without a `plan` span, the operator sees this:

```text
invoke_agent "research_agent"          1200ms
├── chat "gpt-4o"                       400ms  ← planning? reasoning? unclear
├── execute_tool "web_search"           350ms
├── execute_tool "summarize"            200ms
└── chat "gpt-4o"                       250ms  ← final response? another plan?
```

The first `chat` span might be planning, reasoning, or a direct answer attempt. The operator cannot distinguish these failure modes.

With a `plan` span:

```text
invoke_agent "research_agent"          1200ms
├── plan "research_agent"               400ms  ← planning duration isolated
│   └── chat "gpt-4o"                   400ms  (LLM generates the plan)
├── execute_tool "web_search"           350ms  (step 1)
├── execute_tool "summarize"            200ms  (step 2)
└── chat "gpt-4o"                       250ms  (final response)
```

Planning latency, errors, and the LLM call(s) that produced the plan are now isolated under a parent boundary. An operator can filter on `gen_ai.operation.name = plan`, set sampling rules on planning spans independently, and immediately see whether planning or execution consumed the time budget.

### Why a span, not an attribute or grouping primitive

`execute_tool` got its own span — not an attribute on `chat` — because tool execution has independent duration, error status, and parent-child structure. Planning has the same properties.

A boolean `gen_ai.planning = true` attribute on `chat` would only work if planning were always exactly one LLM call. It isn't.

**Current evidence — 1 plan : N chat in production library code paths.**

The `run_crew_planning_with_native_retry()` reference scenario (`reference/scenarios/crewai/scenario.py`) walks one CrewAI planner invocation that produces **3 child `chat` spans** under one `plan` parent, all from real library-native code paths:

1. `beta.chat.completions.parse` — planner's structured-output call carrying the planner role prompt.
2. `chat.completions.create` — fall-through after `_handle_completion` sees `parsed_object=None` ([`completion.py:1612`](https://github.com/crewAIInc/crewAI/blob/main/lib/crewai/src/crewai/llms/providers/openai/completion.py#L1612)). Same prompt, different code path.
3. `chat.completions.create` — issued by `Task._export_output → convert_to_model → handle_partial_json → convert_with_instructions → Converter.to_pydantic()` ([`converter.py`](https://github.com/crewAIInc/crewAI/blob/main/lib/crewai/src/crewai/utilities/converter.py)) carrying CrewAI's schema-conversion system prompt — a **different kind of work** (parse-then-convert), not a retry of calls 1 or 2.

These three calls belong to one planning operation but are not three retries — they are three structurally distinct LLM round-trips the framework dispatches as part of producing one plan. A boolean attribute on the third `chat` would not tell an operator that the first two were also part of the same planning effort.

**Future direction — patterns the spec should leave room for.**

Planning architectures published in the last 18 months that already exceed 1:1:

- **Plan-Critique-Replan** (Reflexion, Self-Refine) — a separate LLM call evaluates the plan and a third call replans on failure. All three contribute to one plan operation's outcome.
- **Multi-stage decomposition** — a "macro" planning call produces high-level steps, each step is then expanded by a "micro" planning call. The plan span needs to parent the full set without forcing the instrumentor to mint multiple plan operations for one logical decomposition.
- **Parallel candidate planning** — generate K plan candidates in parallel and pick the best (tree-of-thought variants). Operators want one parent boundary they can use for SLOs and sampling rules without correlating K independent siblings.

In each of these, the operator needs:

- a **single duration** that captures the whole planning effort (input construction + N LLM calls + post-processing),
- a **single error status** that rolls up failures across the calls,
- a **parent boundary** for sampling and filtering (`gen_ai.operation.name = plan`),
- and the freedom to leave the **per-call shape** to whatever the framework actually does, without having to backfill the spec each time a new pattern appears.

A span gives all of that. A boolean attribute on `chat` gives none of it once N > 1.

**Operation vs. modifier in the `gen_ai.*` namespace.**

`gen_ai.operation.name` is used for kinds of work that have their own start/end and parent-child structure (`chat`, `embeddings`, `execute_tool`, `invoke_agent`). Attributes like `stream`, `temperature`, `tool_choice` are request modifiers on a single LLM call. A planning boolean would be the only operation-as-attribute in the namespace, and the cases above show it would not survive contact with real frameworks.

### Relationship to `gen_ai.task`

A plan formulates strategy before execution; a task executes assigned work. The plan span is the parent of the planning LLM call(s) and a sibling of the task/tool spans that follow.

### Prior art

| Framework | Planning Hook | Auto-instrumentable | Source |
|-----------|---------------|:-------------------:|--------|
| CrewAI | `CrewPlanner` — explicit planning phase before task execution | Yes | [planning_handler.py](https://github.com/crewAIInc/crewAI/blob/main/lib/crewai/src/crewai/utilities/planning_handler.py) |
| LangChain (experimental) | `load_chat_planner()` — Plan-and-Execute pattern returning `LLMPlanner.plan(...)` | Yes | [chat_planner.py](https://github.com/langchain-ai/langchain-experimental/blob/main/libs/experimental/langchain_experimental/plan_and_execute/planners/chat_planner.py) |
| LlamaIndex | `SubQuestionQueryEngine` — question decomposition before sub-queries | Yes | [sub_question_query_engine.py](https://github.com/run-llama/llama_index/blob/main/llama-index-core/llama_index/core/query_engine/sub_question_query_engine.py) |
| Google ADK | `planner` agent with `plan()` method | Unverified | — |

Instrumentation SHOULD only emit `plan` when the framework exposes an explicit planning boundary (see emission rules in `spans.yaml`).

## Prototype

Three reference scenarios demonstrate the `plan` span:

1. **`reference/scenarios/crewai/scenario.py::run_crew_planning()`** — 1 plan parent, 1 child chat. Drives the real `CrewPlanner._handle_crew_planning()` path; mock server detects the `"Task Execution Planner"` role and returns `PlannerTaskPydanticOutput`-shaped JSON via `beta.chat.completions.parse`.
2. **`reference/scenarios/crewai/scenario.py::run_crew_planning_with_native_retry()`** — 1 plan parent, **3 child chats** through CrewAI's library-native parse → fall-through → converter pipeline. Demonstrates the 1:N plan:chat shape that motivates the span-vs-boolean choice (see "Why a span" section above).
3. **`reference/scenarios/langchain/scenario.py::run_plan_and_execute_reference()`** — 1 plan parent, 1 child chat. Uses `langchain_experimental.plan_and_execute.load_chat_planner()`; mock detects the `"<END_OF_PLAN>"` `SYSTEM_PROMPT` marker and returns numbered-step text that `PlanningOutputParser` splits into `Plan(steps=[...])`.

Captured telemetry is validated by Weaver and lands in each scenario's `data.json`. The `Plan` row in `reference/README.md` and `reference/reports/plan-span.md` lists the supporting libraries.

Out of scope (deferred to follow-up PRs): plan-specific attributes (`strategy`, `step.count`), reflection, and delegation.

## Checklist

- [x] Reference instrumentation and scenarios updated for affected libraries


Krishnachaitanyakc:
@trask @lmolkova can you review this? Copy of https://github.com/open-telemetry/semantic-conventions/pull/3594

model/gen-ai/spans.yaml
```diff
@@ -478,6 +478,54 @@ spans:
...
+      - ref: gen_ai.operation.name
+        requirement_level: required
+        sampling_relevant: true
+      - ref: error.type
+        requirement_level:
+          conditionally_required: "if the operation ended in an error"
+        note: |
+          The `error.type` SHOULD match the error code returned by the Generative AI provider or the client library,
+          the canonical name of exception that occurred, or another low-cardinality error identifier.
+          Instrumentations SHOULD document the list of errors they report.
```
trask: Can reuse the shared `attributes.gen_ai.error` group.

```suggestion
    attributes:
      - ref_group: attributes.gen_ai.error
      - ref: gen_ai.operation.name
        requirement_level: required
        sampling_relevant: true
```

model/gen-ai/spans.yaml
```diff
@@ -478,6 +478,54 @@ spans:
...
+    name:
+      note: >
+        **Span name** SHOULD be `plan {gen_ai.agent.name}` if `gen_ai.agent.name` is readily available.
+        When `gen_ai.agent.name` is not available, it SHOULD be `plan`.
+    brief: >
+      Represents an agent planning or task decomposition phase.
+    note: |
+      The `gen_ai.operation.name` SHOULD be `plan`.
+      **Span name** SHOULD be `plan {gen_ai.agent.name}` if `gen_ai.agent.name` is readily available.
+      When `gen_ai.agent.name` is not available, it SHOULD be `plan`.
```
trask, line 498: The `Span name` guidance is already in `name.note` above.

```suggestion
      The `gen_ai.operation.name` SHOULD be `plan`.
```

model/gen-ai/spans.yaml
```diff
@@ -478,6 +478,54 @@ spans:
...
+
+      A plan span represents the decision phase where an agent formulates
+      a strategy before executing it. The LLM call that generates the plan is a child
+      of the plan span, and the tool or task spans produced from the plan are sibling
+      operations under the same `invoke_agent` span.
```
trask: Soften to `SHOULD` / `typically` — the parent/sibling layout is the recommended shape, not something every instrumentation can guarantee.

```suggestion
      a strategy before executing it. The LLM call that generates the plan
      SHOULD be a child of the plan span, and the tool or task spans produced
      from the plan are typically sibling operations under the same
      `invoke_agent` span.
```

reference/scenarios/crewai/scenario.py
trask, line 1: can you check out the report from the [evaluate-reference/SKILL.md](https://github.com/open-telemetry/semantic-conventions-genai/blob/main/.github/skills/evaluate-reference/SKILL.md)?

# Evaluation: PR 97 — `plan` operation for GenAI agent planning

## Mirrored proposal

PR 97 introduces `gen_ai.plan.internal` (span kind `INTERNAL`, op name `plan`) with:

- **Required:** `gen_ai.operation.name`
- **Conditionally required:** `error.type`, `gen_ai.agent.id`, `gen_ai.agent.name`

The reference set targets exactly this attribute list — the new [reference/reports/plan-span.md](reference/reports/plan-span.md) and [reference/scenarios/crewai/data.json](reference/scenarios/crewai/data.json) match the model. PR target is mirrored faithfully.

## Credible reference coverage

**crewai** ([reference/scenarios/crewai/scenario.py](reference/scenarios/crewai/scenario.py))

| Attribute | Source | Class |
| --- | --- | --- |
| `gen_ai.operation.name` | constant `"plan"` set inside `CrewPlanner` flow | direct |
| `gen_ai.agent.id` | `Agent.id` (UUID owned by the agent object) | direct |
| `gen_ai.agent.name` | `Agent.role` | direct |
| `error.type` | not exercised, but trivially capturable from a planner exception | direct (gap is honest — no error path in scenario) |

CrewAI is the strongest candidate: `CrewPlanner` is a distinct, named pre-execution phase that runs an LLM call to decompose tasks. The classifier change in [reference/src/semconv_genai/classify.py](reference/src/semconv_genai/classify.py) correctly excludes `plan` from being mis-counted as `invoke_agent` (it shares the agent attribute set).

## Weak / questionable choices in the crewai scenario

1. **Wrong agent tagged on the plan span.** The scenario sets `gen_ai.agent.id` / `gen_ai.agent.name` from `researcher` — the worker agent — but `CrewPlanner` internally synthesizes its own `Agent(role="Task Execution Planner", ...)` to actually run the planning LLM call (`crewai/utilities/planning_handler.py`). A native instrumentation hooking `CrewPlanner._handle_crew_planning` would naturally see the planner agent, not a worker agent. The spec is ambiguous about which agent to identify, but the *current call boundary* in CrewAI exposes the planner agent. Recommend either (a) tag the planner agent's `id` / `role`, or (b) leave both attributes off and add a comment that CrewAI's planner agent is an internal detail not meaningful to consumers — rather than tagging an agent that did not participate in the planning call.

2. **Private API + synthetic chat span.** The scenario calls `planner._create_tasks_summary()` (private) and then *manually* fabricates the child `chat` span instead of going through `CrewPlanner._handle_crew_planning()` which would actually invoke the LLM via `planning_agent.execute_task(task)` against the mock server. This is acceptable for a reference (other scenarios fabricate too), but the comment "exercise the public construction path here so the import surface is honestly covered" overstates it: `_create_tasks_summary` is private. A more honest framing would acknowledge this is illustrative scaffolding around the real planner entry point.

3. **No `error.type` path.** Acceptable as an honest gap; consistent with other internal spans that omit error scenarios.

## Missing supporting library implementations

The spec explicitly cautions: *"SHOULD NOT be reported when the instrumentation cannot distinguish planning from generic reasoning"* — so libraries without a separable planning phase are correctly excluded. Candidates that *do* have a separable phase and are not yet covered:

- **llamaindex** — `StructuredPlannerAgent` / `Planner` produces a `Plan` object before sub-task execution. Reference scenario exists ([reference/scenarios/llamaindex/scenario.py](reference/scenarios/llamaindex/scenario.py)) but does not exercise the planner agent. Add reference for supporting library.
- **google-adk** — `PlanReActPlanner` / `BuiltInPlanner` produce an explicit planning step distinct from the model call. Scenario exists ([reference/scenarios/google-adk/scenario.py](reference/scenarios/google-adk/scenario.py)) and could be extended. Add reference for supporting library.
- **langchain** — Plan-and-Execute (`langchain-experimental`) has `BaseChatPlanner.plan()` returning a `Plan`. Lower priority because it lives in `langchain-experimental` and is less actively maintained, but still a credible direct-observability source.

Honest capture gaps (leave unchanged, no implementation needed):

- **autogen, openai-agents, openai-assistants, pydantic-ai, claude-agent-sdk** — no API boundary that distinguishes planning from generic reasoning/chat.
- **dspy** — ChainOfThought / ReAct are reasoning, not planning; spec note explicitly excludes this case.
- **haystack, anthropic, openai, aws-bedrock, cohere, groq, mistralai, vertexai, azure-ai-inference, azure-openai, google-genai, instructor, litellm** — pure inference clients; no planning phase.
- **aws-bedrock-agent, azure-ai-foundry** — managed-agent services that do not surface planning as a separate API call.

## Evaluation recommendation

- **fix implementation (crewai):** Re-tag `gen_ai.agent.id` / `gen_ai.agent.name` to reflect the agent actually running the planning LLM call (CrewPlanner's internal `Agent(role="Task Execution Planner")`), or drop both attributes if the planner-internal agent is not meaningful to consumers. Also consider replacing the `_create_tasks_summary()` + synthetic chat span with a call through `CrewPlanner._handle_crew_planning()` so the chat span wraps the real LLM round-trip against the mock server.
- **add reference for supporting library:** llamaindex (`StructuredPlannerAgent`), google-adk (`PlanReActPlanner` / `BuiltInPlanner`). Optionally langchain Plan-and-Execute.
- **leave unchanged; honest capture gap:** autogen, openai-agents, openai-assistants, pydantic-ai, claude-agent-sdk, dspy, haystack, and all pure-inference SDKs — these honestly cannot distinguish planning from reasoning at their call boundary, which is exactly what the proposal's "SHOULD NOT" clause is meant to cover.

Krishnachaitanyakc: Thanks @trask, addressed everything except:

1. _handle_crew_planning() swap — it calls planning_agent.execute_task(...) with output_pydantic=PlannerTaskPydanticOutput; the mock can't return schema-valid structured output without per-library support. Synthetic chat span is the closest honest reference.
2. llamaindex StructuredPlannerAgent — not in the pinned llama-index-core==0.14.21. The agent rewrite removed it; only BaseWorkflowAgent, FunctionAgent, CodeActAgent, ReActAgent, AgentWorkflow remain.
3. google-adk PlanReActPlanner / BuiltInPlanner — both exist in 1.32.0 but are call-site modifiers (PlanReActPlanner injects /*PLANNING*/ tags into the same call; BuiltInPlanner sets Gemini thinking_config). No pre-call plan boundary — maps to the spec's "SHOULD NOT" clause.
4. langchain Plan-and-Execute — would require pulling langchain-experimental (not in current langchain==1.2.17). BasePlanner.plan(inputs) -> Plan is clean; happy to add if you want the extra dep.

model/gen-ai/spans.yaml
Krishnachaitanyakc: Applied all three suggestions.

reference/scenarios/crewai/scenario.py
```diff
@@ -212,13 +212,102 @@ def _capture_completion(*args, **kwargs):
...
+    )
+
+    # CrewPlanner._handle_crew_planning() drives a real LLM round-trip via
+    # planning_agent.execute_task(...) with output_pydantic=PlannerTaskPydanticOutput,
+    # which the mock server cannot satisfy without library-specific schema
+    # support. The closest honest reference is therefore: build the same
+    # planner agent CrewPlanner uses internally (Agent(role="Task Execution
+    # Planner", ...)), drive _create_tasks_summary() to exercise the real
+    # input-construction path, and emit the chat span manually around the
+    # planning LLM call we cannot route through the mock here.
```
trask: it would be good to see the real api exercised, can you add support to the mock server for it?

reference/scenarios/crewai/scenario.py
trask: > langchain Plan-and-Execute — would require pulling langchain-experimental (not in current langchain==1.2.17). BasePlanner.plan(inputs) -> Plan is clean; happy to add if you want the extra dep.

yeah, I think it would be good to have a second example, even if it's an experimental one

reference/scenarios/crewai/scenario.py
Krishnachaitanyakc: Good point, added it, can you check?

reference/scenarios/crewai/scenario.py
Krishnachaitanyakc: Done in the latest commit, added run_plan_and_execute_reference() driving real langchain_experimental.LLMPlanner.plan() through the mock.

model/gen-ai/spans.yaml
```diff
@@ -478,6 +478,47 @@ spans:
...
+  - type: gen_ai.plan.internal
+    kind: internal
+    name:
+      note: >
+        **Span name** SHOULD be `plan {gen_ai.agent.name}` if `gen_ai.agent.name` is readily available.
+        When `gen_ai.agent.name` is not available, it SHOULD be `plan`.
+    brief: >
+      Represents an agent planning or task decomposition phase.
+    note: |
+      The `gen_ai.operation.name` SHOULD be `plan`.
```
trask: Sorry, I think I suggested to remove this before since it's duplicated above but realized it's not rendering into markdown from above yet.

```suggestion
      The `gen_ai.operation.name` SHOULD be `plan`.

      **Span name** SHOULD be `plan {gen_ai.agent.name}` if `gen_ai.agent.name` is readily available.
      When `gen_ai.agent.name` is not available, it SHOULD be `plan`.
```

trask, line 495: Sorry, I think I suggested to remove this before because of duplication with above, but realized now that it's not rendered into markdown.

```suggestion
      The `gen_ai.operation.name` SHOULD be `plan`.

      **Span name** SHOULD be `plan {gen_ai.agent.name}` if `gen_ai.agent.name` is readily available.
      When `gen_ai.agent.name` is not available, it SHOULD be `plan`.
```

reference/scenarios/langchain/pyproject.toml
```diff
@@ -4,6 +4,8 @@ version = "0.0.0"
 requires-python = ">=3.12"
 dependencies = [
     "langchain==1.2.17",
+    "langchain-classic==1.0.6",
```
trask: `langchain-classic` isn't imported anywhere in `scenario.py` — it's pulled in transitively by `langchain-experimental==0.4.1`. Listing it as a direct dep is redundant; `uv` will resolve it from the experimental package's metadata. Suggest removing this line unless there's a specific reason to pin it independently of what `langchain-experimental` requires.

```suggestion
    "langchain-experimental==0.4.1",
```

model/gen-ai/spans.yaml
```diff
@@ -478,6 +478,47 @@ spans:
...
+      that the operation being instrumented is planning or task decomposition, and SHOULD
+      NOT be reported when the instrumentation cannot distinguish planning from generic
+      reasoning or normal inference.
+
+      Grouping primitives or span links can correlate planning-related spans, but they
+      do not represent the planning phase itself as a timed parent span with a child
+      `chat` call.
+
+      This is distinct from task execution. A task span represents execution of assigned
+      work, while a plan span represents strategy formulation before execution.
```
trask: Do you think these two sentences are needed to avoid confusion or are covered adequately by the definition above?

model/gen-ai/spans.yaml
Krishnachaitanyakc: added the span name lines to `note:` so they render.

reference/scenarios/langchain/pyproject.toml
Krishnachaitanyakc: dropped the explicit pin. Verified transitive chain: `langchain-experimental==0.4.1` → `langchain-community>=0.4.0` (`==0.4.1` resolved) → `langchain-classic>=1.0.0,<2.0.0` (`==1.0.6` resolved). uv.lock still has `langchain-classic==1.0.6`, scenario re-run passes.

model/gen-ai/spans.yaml
Krishnachaitanyakc: You're right. Dropped both.

CHANGELOG.md
```diff
@@ -10,6 +10,8 @@
 
 ### 💡 Enhancements 💡
 
+- Add `plan` operation for GenAI agent planning/task decomposition spans.
```
trask, line 15:
```suggestion
- Add `plan` operation for GenAI agent planning/task decomposition spans.
  ([#97](https://github.com/open-telemetry/semantic-conventions-genai/pull/97))
```

model/gen-ai/spans.yaml
```diff
@@ -478,6 +478,42 @@ spans:
...
+    name:
+      note: >
+        **Span name** SHOULD be `plan {gen_ai.agent.name}` if `gen_ai.agent.name` is readily available.
+        When `gen_ai.agent.name` is not available, it SHOULD be `plan`.
+    brief: >
+      Represents an agent planning or task decomposition phase.
+    note: |
+      The `gen_ai.operation.name` SHOULD be `plan`.
+      **Span name** SHOULD be `plan {gen_ai.agent.name}` if `gen_ai.agent.name` is readily available.
+      When `gen_ai.agent.name` is not available, it SHOULD be `plan`.
```
trask, line 498:
```suggestion

      When `gen_ai.agent.name` is not available, it SHOULD be `plan`.
```

model/gen-ai/spans.yaml
```diff
@@ -478,6 +478,42 @@ spans:
       - ref: gen_ai.output.messages
         requirement_level: opt_in
 
+  - type: gen_ai.plan.internal
```
trask, line 486: > ### Why a span, not an attribute or grouping primitive
>
> `execute_tool` got its own span -- not an attribute on `chat` -- because tool execution has independent duration, error status, and parent-child structure. Planning has the same properties. A grouping attribute correlates sibling spans; a `plan` span creates a parent boundary with its own duration. You cannot model a parent as an attribute on its own child.

can you expand this, particularly to address why not a boolean "planning" attribute on the chat span? I assume it's to support multiple steps under a single planning span, but currently the library examples only support a single chat span per planning operation, so I think it deserves a bit more explanation, thanks!

Krishnachaitanyakc: 3 reasons come off the top of my mind. 

Firstly, Operation vs modifier, gen_ai.* uses operation.name for kinds of work such as chat, execute_tool, invoke_agent (each of these get a span) and attributes for request modifiers (stream, temperature). A planning boolean would be the only operation-as-attribute in the namespace. 
Secondly, The 1-chat-per-plan ratio in the references is implementation-current, not a spec invariant. CrewAI's planner runs inside a `while not AgentFinish` loop, LangChain Plan-and-Execute with OutputFixingParser issues repair calls, plan-critique-replan loops issue several. Mocks happen to succeed first try. 
Thirdly, And even at 1:1, plan duration includes input construction (_create_tasks_summary) and output parsing (Pydantic validation, regex split), not just the LLM call.

trask: > Firstly, Operation vs modifier, gen_ai.* uses operation.name for kinds of work such as chat, execute_tool, invoke_agent (each of these get a span) and attributes for request modifiers (stream, temperature). A planning boolean would be the only operation-as-attribute in the namespace.

I'm not sure this is compelling if there's really a 1-1 relationship between plan and chat spans, there doesn't seem like a need for both.

I think the case we need to make here is why it's not a 1-1 relationship.

> Secondly, The 1-chat-per-plan ratio in the references is implementation-current, not a spec invariant. CrewAI's planner runs inside a while not AgentFinish loop, LangChain Plan-and-Execute with OutputFixingParser issues repair calls, plan-critique-replan loops issue several. Mocks happen to succeed first try.

Can you show this in the reference instrumentation? that would be really helpful

> Thirdly, And even at 1:1, plan duration includes input construction (_create_tasks_summary) and output parsing (Pydantic validation, regex split), not just the LLM call.

I'm not sure this is too strong of a case either just based on what we've typically modeled in semantic conventions, let's focus on your second point and try to flush that out a bit more in the reference instrumentation or in the PR description about why it's not limited to 1:1, thanks!

Krishnachaitanyakc: Done in 7dadcec, added run_plan_with_repair_reference() in reference/scenarios/langchain/scenario.py. This wraps LLMPlanner.output_parser with OutputFixingParser; first LLM call returns malformed text, parser raises, OutputFixingParser issues a second LLM call wrapping the malformed completion in NAIVE_FIX_PROMPT. Both chat spans captured under one plan span via the same chat.completions.create patch and the output is: "planned 3 step(s) after 2 chat call(s)".

About the trigger: the default PlanningOutputParser silently returns Plan(steps=[]) on malformed input rather than raising, so the demo subclasses it to raise OutputParserException (and uses a [FORCE_MALFORMED] sentinel in the user input to route the mock's response). The OutputFixingParser machinery itself (langchain_classic/output_parsers/fix.py) is the canonical langchain repair pattern — what's contrived is the trigger, not the captured telemetry.

A more naturally occurring multi-call planning case exists in CrewAI which has a `while not isinstance(formatted_answer, AgentFinish)` retry loop that re-issues real LLM calls when the planner's response fails Pydantic validation. Happy to add a CrewAI demo that triggers this native retry path (mock returns invalid Pydantic JSON first, valid second) if the OutputFixenough on its own.

I didn't update the PR description yet, let me know if this looks good and I can update it.

trask: > This wraps LLMPlanner.output_parser with OutputFixingParser; first LLM call returns malformed text, parser raises, OutputFixingParser issues a second LLM call wrapping the malformed completion in NAIVE_FIX_PROMPT

I think one could potentially argue that this could be modeled more like a "retry mechanism" (which we have other precedent for modeling in semantic conventions).

> A more naturally occurring multi-call planning case exists in CrewAI

can you add this? I agree it sounds like a better example. Thanks!

Krishnachaitanyakc: Done in 291d5b4, added run_crew_planning_with_native_retry() in reference/scenarios/crewai/scenario.py. The [FORCE_PLANNER_RETRY] sentinel in the task description scopes the malformed-response behavior to this scenario only.

The trace contains one plan Task Execution Planner span parenting 3 chat gpt-4o-mini children, all from real CrewAI library code paths:

1. beta.chat.completions.parse — planner's structured-output call. Mock returns {message.content: null, refusal: "..."}. CrewAI's _handle_completion (crewai/llms/providers/openai/completion.py:1612) treats parsed_object=None as a fall-through trigger.
2. chat.completions.create fall-through with the planner's original prompt (no response_format key — the planner's LLM instance leaves self.response_format unset). Mock returns plain text the converter cannot validate as PlannerTaskPydanticOutput.
3. chat.completions.create issued by Task._export_output → convert_to_model → handle_partial_json → convert_with_instructions → Converter.to_pydantic() (crewai/utilities/converter.py:155-336), carrying CrewAI's schema-conversion system prompt ("Format your final answer according to the following OpenAPI schema..." from crewai/translations/en.json). Mock detects that prompt and returns valid PlannerTaskPydanticOutput JSON.

Output is planned 1 task(s) after 3 chat call(s).

The 3 chats run on different code paths (beta.parse, create fall-through, converter's create) and carry different system prompts (planner role, planner role again, schema-conversion role) — the plan span's duration includes all three plus the planner-side input construction (_create_tasks_summary) and the Pydantic validation between calls. That's the case for a span boundary that a boolean attribute on chat can't carry.

This is structurally different from langchain. Call 3 is not a retry of call 1 or 2 with the same intent but it's a different kind of work (parse-then-convert, with a different system prompt and a different input shape) that the planner agent dispatches as part of producing the plan. So even if a gen_ai.retry.attempt style attribute landed for repair-style retries, the converter call 3 wouldn't fit under it.

trask: thanks, that looks good. I'd suggest reverting 7dadcec4f2af1bc771d8cb5343c691de566e1c8b

and then update the PR description to dive a bit more into the current and future need for multiple chat operations per planning operation

Krishnachaitanyakc: Done, reverted and updated the PR description.

trask (approved):
Thanks @Krishnachaitanyakc!

reference/scenarios/crewai/scenario.py
trask, line 1: I sent a PR to your branch with a bit of refactoring that helped me follow it better at least: https://github.com/Krishnachaitanyakc/semantic-conventions-genai/pull/1

Krishnachaitanyakc: Sounds good, merged it here.

lmolkova (approved):

PR #98, opened 2026-05-05
gen-ai: model agent-to-agent handoff as execute_tool span

Krishnachaitanyakc:
## Description

Models openai-agents agent-to-agent handoff as an `execute_tool` span carrying two new attributes (`gen_ai.agent.handoff.source.name`, `gen_ai.agent.handoff.target.name`). Drops the previously-proposed `gen_ai.agent.invocation.trigger` attribute (see thread).

Resubmits open-telemetry/semantic-conventions#3614 to this repository per open-telemetry/semantic-conventions#3696. Partially addresses #35 and #24.

### Diff at a glance

- Adds `gen_ai.agent.handoff.{source,target}.name` to the registry; refs both on `gen_ai.execute_tool.internal`.
- Removes `gen_ai.agent.invocation.trigger` from the registry and from `attributes.gen_ai.invoke_agent.common`.
- Rewrites `reference/scenarios/openai-agents/scenario.py::run_agent_with_handoff()` to emit an `execute_tool transfer_to_<agent>` span sourced from `HandoffCallItem` + `HandoffOutputItem`.
- Reverts the 6 other `invoke_agent` scenarios (autogen, aws-bedrock-agent, azure-ai-foundry, google-adk, openai-assistants, pydantic-ai) to upstream main — their previous `trigger="direct"` adds were hardcoded scenario knowledge.

## Background

The PR originally added a `gen_ai.agent.invocation.trigger` enum on `invoke_agent` spans to flag delegated invocations. Per #discussion_r3262472777 + #discussion_r3262483276, that design had two issues: the trigger value had to be inferred from per-run state (unreliable when an agent was both delegated to and directly invoked in the same run), and only one of the seven reference scenarios could derive a value from typed library state. This revision pivots to the maintainer's suggested direction — modeling handoff as a tool-call execution and capturing from / to agents on it.

### User journey

Multi-agent debugging, cost attribution, and blast-radius analysis.

Today, when one agent delegates to another via a framework's typed handoff primitive, the trace shows two `invoke_agent` spans with no causal link visible at the span level. After:

```
invoke_agent triage-agent
└─ execute_tool transfer_to_billing_agent   gen_ai.agent.handoff.source.name=triage-agent
                                            gen_ai.agent.handoff.target.name=billing-agent
invoke_agent billing-agent
```

The `execute_tool` span makes the delegation a first-class, queryable event. Operators can:

- **Cost-attribute** delegated work back to the orchestrating agent.
- **Trace blast-radius** when an agent misbehaves (find every `execute_tool` whose `gen_ai.agent.handoff.target.name = "<misbehaving agent>"`).
- **Set SLOs and sampling rules** on delegation events independently of direct invocations.

### Why model handoff as `execute_tool`

The OpenAI Agents SDK already models handoff as a function tool call. `HandoffCallItem` extends `RunItemBase[ResponseFunctionToolCall]` ([items.py#L266](https://github.com/openai/openai-agents-python/blob/v0.17.0/src/agents/items.py#L266)); `convert_handoff_tool` serializes each `Handoff` as a `ChatCompletionToolParam` with `"type": "function"` ([chatcmpl_converter.py#L864](https://github.com/openai/openai-agents-python/blob/v0.17.0/src/agents/models/chatcmpl_converter.py#L864)). Reusing `execute_tool` avoids adding a new operation for what the framework already treats as a tool. The source and target agent identities come from typed `HandoffOutputItem.source_agent` / `HandoffOutputItem.target_agent` references ([turn_resolution.py#L385](https://github.com/openai/openai-agents-python/blob/v0.17.0/src/agents/run_internal/turn_resolution.py#L385)).

## Prototype

Reference scenario: [`reference/scenarios/openai-agents/scenario.py::run_agent_with_handoff()`](../blob/agent-invocation-trigger/reference/scenarios/openai-agents/scenario.py). A triage agent owns one `Handoff` to a billing specialist. The mock prefers `transfer_to_*` tools so the SDK's native handoff machinery fires. The scenario opens an `invoke_agent triage-agent` span via `RunHooks.on_agent_start`, captures the source-span context at `RunHooks.on_handoff` before closing it, then walks `result.new_items` after `Runner.run` returns to emit an `execute_tool transfer_to_<agent>` span parented to the captured source-span context. Pairs `HandoffCallItem` / `HandoffOutputItem` by `call_id` for forward-compatibility with multi-handoff flows.

Attributes on the `execute_tool` span (all sourced from typed SDK fields):

- `gen_ai.operation.name = execute_tool`
- `gen_ai.tool.name` ← `ResponseFunctionToolCall.name` (e.g., `transfer_to_billing_agent`)
- `gen_ai.tool.call.id` ← `ResponseFunctionToolCall.call_id`
- `gen_ai.tool.call.arguments` ← `ResponseFunctionToolCall.arguments`
- `gen_ai.agent.handoff.source.name` ← `HandoffOutputItem.source_agent.name`
- `gen_ai.agent.handoff.target.name` ← `HandoffOutputItem.target_agent.name`

The receiving agent gets its own `invoke_agent billing-agent` span as a sibling — handoff causality is captured by the `execute_tool` span, not by parenting the target.

Per `reference/CONTRIBUTING.md`, both new attributes are demonstrated only where the library exposes a typed handoff primitive (OpenAI Agents). Other scenarios that previously hardcoded `trigger="direct"` are reverted — those values were scenario knowledge, not library-derived.

## Out of scope

- Other frameworks' delegation primitives (CrewAI, AutoGen, LangGraph). Straightforward follow-on if they expose comparable typed signals.
- Function-tool spans within the handoff scenario (only handoff tool calls are emitted).

## Checklist

- [x] Reference scenarios updated for affected libraries (openai-agents)


Krishnachaitanyakc:
Corresponding PR for #[3614](https://github.com/open-telemetry/semantic-conventions/pull/3614) in semantic-conventions

Nik-Reddy:
@Krishnachaitanyakc This is well thought out, the prior art table and design decisions section are really helpful. One thing I noticed, the scenarios only emit trigger="direct" but never trigger="agent". 

I get that the current scenarios don't naturally exercise delegation, but would it be worth adding a minimal multi-agent scenario (even a simple A delegates to B mock) to validate the agent value e2e? Otherwise LGTM on the model and registry changes.

Krishnachaitanyakc:
> @Krishnachaitanyakc This is well thought out, the prior art table and design decisions section are really helpful. One thing I noticed, the scenarios only emit trigger="direct" but never trigger="agent".
> 
> I get that the current scenarios don't naturally exercise delegation, but would it be worth adding a minimal multi-agent scenario (even a simple A delegates to B mock) to validate the agent value e2e? Otherwise LGTM on the model and registry changes.

Thanks @Nik-Reddy, trigger="agent" is actually already demonstrated in c730ef2. run_agent_with_handoff() in reference/scenarios/openai-agents/scenario.py wires a triage agent → billing specialist via the SDK's real Handoff machinery. The delegated billing-agent span emits trigger="agent" sourced from the SDK's HandoffOutputItem type. The reports only show attribute KEY presence (not enum values emitted), which is why it isn't visible at the surface level.

reference/scenarios/autogen/scenario.py
```diff
@@ -153,6 +153,7 @@ async def _run():
                 "invoke_agent test_agent", attributes=agent_span_attributes
             ) as agent_span:
                 agent_span.set_attribute("gen_ai.agent.description", agent.description)
+                agent_span.set_attribute("gen_ai.agent.invocation.trigger", "direct")
```
lmolkova: where does this value come from? is there something in the input parameters that would contain this information? 

If it's a hard-coded value, how would instrumentations know what to set it to?

Krishnachaitanyakc: Good point, So trigger="direct" on autogen / pydantic-ai / openai-assistants / azure-ai-foundry is hardcoded scenario knowledge, not library-derived. Not honest under the reference SKILL.

Proposing fix: re-scope to single-value enum `agent` with absence-as-direct semantics, mirroring how `error.type` is `conditionally_required: if the operation ended in an error`, emit only with positive evidence. Auto-instrumentations: emit "agent" when their handoff hook fires, skip otherwise. So trigger="agent" remains on the openai-agents handoff target (only library-derivable demo today); trigger="direct" drops from all 7 other scenarios.

Fallback option I can think of is if you'd rather keep both values: change requirement from `recommended: if known` to `opt_in`, application code MAY set, auto-instrumentations MUST NOT unless library-derivable.

Leaning towards the first option (cleaner spec, every emission carries positive evidence). Let me know what you think.

MikeGoldsmith (changes requested):
Thanks @Krishnachaitanyakc - I agree we should just use `agent` and removing `direct` for the enum.

Marking as Request Changes so we know it's waiting on an update from you 👍🏻

Krishnachaitanyakc:
@MikeGoldsmith @lmolkova updated it, PTAL

eternalcuriouslearner:
@Krishnachaitanyakc Upon reading the motivation I believe the intention behind this enum is to provide visibility into delegation events. Did I get this right?

Krishnachaitanyakc:
@eternalcuriouslearner Yes, agent is set only when the instrumentation can positively detect delegation from library state. Absence doesn't mean direct, just that no delegation signal was found.

reference/scenarios/openai-agents/scenario.py
```diff
@@ -167,9 +167,244 @@ def main():
...
+                agent_span.set_attribute("gen_ai.usage.input_tokens", usage["input_tokens"])
+            if usage.get("output_tokens"):
+                agent_span.set_attribute("gen_ai.usage.output_tokens", usage["output_tokens"])
+        finish_reasons = finish_reasons_per_agent.pop(agent_name, None)
+        if finish_reasons:
+            agent_span.set_attribute("gen_ai.response.finish_reasons", finish_reasons)
+
+    class _SpanLifecycleHooks(RunHooks):
+        async def on_agent_start(self_hooks, context, agent):
+            is_delegated = agent.name in delegated_agent_names
```
lmolkova: `delegated_agent_names` seems to be global static state with no cleanup. 
the check seems to be based on the fact that this agent was ever delegated to. What if one operation delegated to that agent and another one called it directly?

So it seems this instrumentation approach is not reliable. Added some thoughts on how to model it in the https://github.com/open-telemetry/semantic-conventions-genai/pull/98#discussion_r3262483276

reference/scenarios/autogen/scenario.py
lmolkova: since it's only one agent framework (openai agents) that can populate it, I think it does not make sense to define a generic attribute.

Also, looking at how this is implemented in instrumentation, it seems that it's not part of the input parameters on the agent being triggered to know who triggered it.

I would instead look into modeling openai handoff operation as a tool call execution (since it seems this is how openai agents SDK models it) and capturing from / to agents on it.

reference/scenarios/openai-agents/scenario.py
Krishnachaitanyakc: Makes sense, `delegated_agent_names` was "ever delegated to in this run" rather than "delegated to for this invocation". An agent both delegated to and directly invoked in the same run would false-flag every subsequent direct call. Made changes to drop that inferred state entirely.

Pivoted to your suggestion, the execute_tool span now sources gen_ai.agent.handoff.{source,target}.name directly from typed HandoffOutputItem.{source_agent,target_agent}.name references on the SDK-emitted item, no per-run state tracking.

Krishnachaitanyakc:
@lmolkova can you please review this again?

model/gen-ai/registry.yaml
```diff
@@ -257,6 +257,31 @@ attributes:
...
+    examples: ["triage-agent"]
+    stability: development
+  - key: gen_ai.agent.handoff.target.name
+    type: string
+    brief: Human-readable name of the GenAI agent that received a handoff.
+    note: |
+      Set on the `execute_tool` span that represents an agent-to-agent
+      handoff when the framework exposes a typed handoff primitive that
+      names the target agent (for example, OpenAI Agents SDK
+      `HandoffOutputItem.target_agent`).
```
trask, line 490: Instead of:

```
invoke_agent triage-agent
└─ execute_tool transfer_to_billing_agent   gen_ai.agent.handoff.source.name=triage-agent
                                            gen_ai.agent.handoff.target.name=billing-agent
invoke_agent billing-agent
```

Should it look like this?

```
invoke_agent triage-agent
└─ execute_tool transfer_to_billing_agent   gen_ai.agent.handoff.source.name=triage-agent
                                            gen_ai.agent.handoff.target.name=billing-agent
   └─ invoke_agent billing-agent
```

and then this attribute wouldn't be needed either (other than "MAY be set to make the source explicit for queries that should not have to traverse parent-child links")

Krishnachaitanyakc: I don't think the nesting approach works against the SDK's actual control flow because: 

1. The SDK's run loop handles `NextStepHandoff` by mutating `current_agent` and `continue`-ing ([run_loop.py:803-815](https://github.com/openai/openai-agents-python/blob/v0.17.3/src/agents/run_internal/run_loop.py#L803-L815)). The next agent runs in subsequent iterations, not nested under `execute_handoffs` ([turn_resolution.py:510-519](https://github.com/openai/openai-agents-python/blob/v0.17.3/src/agents/run_internal/turn_resolution.py#L510-L519)). Parenting `invoke_agent billing-agent` under `execute_tool` would force `execute_tool` to stay open across the entire downstream conversation, so its duration would cover the target's LLM work rather than the handoff primitive itself — inconsistent with how every other `execute_tool` in the registry is scoped.

2. Sibling also keeps `invoke_agent <agent>` as a uniform top-level boundary regardless of how the agent was triggered; nesting would make queries like "all invocations of billing-agent" parent-shape-sensitive.

I did notice a gap though, `source.name` already carries the "MAY be set explicit for queries that should not traverse parent-child links" fallback (source IS the parent `invoke_agent`). I should add the symmetric note to `target.name` for ops who'd rather query than walk. 

Let me know what you think and I can make the change.

Krishnachaitanyakc:
@trask can you take another pass at this?

model/gen-ai/registry.yaml
```diff
@@ -269,6 +269,31 @@ attributes:
...
+  - key: gen_ai.agent.handoff.source.name
+    type: string
+    brief: Human-readable name of the GenAI agent that initiated a handoff.
+    note: |
+      Set on the `execute_tool` span that represents an agent-to-agent
+      handoff when the framework exposes a typed handoff primitive that
+      names the source agent (for example, OpenAI Agents SDK
+      `HandoffOutputItem.source_agent`).
+
+      In typical agent frameworks, the source agent is also the agent
```
MikeGoldsmith: I don't think we need this additional paragraph because `target.name` provides the context of when it is expected to be added.D

model/gen-ai/registry.yaml
MikeGoldsmith: Whichever way we decide on, child or sibling, we need to make sure to update the entry to reflect the expected hierarchy so other implementations don't do it the wrong way.

reference/scenarios/openai-agents/scenario.py
```diff
@@ -167,9 +168,370 @@ def main():
...
+            with _reference_tracer.start_as_current_span(
+                f"execute_tool {handoff_obj.tool_name}",
+                context=parent_ctx,
+                attributes=tool_span_attributes,
+            ) as tool_span:
+                tool_span.set_attribute("gen_ai.tool.call.id", tool_call.call_id)
+                tool_span.set_attribute("gen_ai.tool.call.arguments", tool_call.arguments)
+                tool_span.set_attribute("gen_ai.agent.handoff.source.name", public_agent.name)
+                new_agent = await original_on_invoke(ctx, args)
+                tool_span.set_attribute("gen_ai.agent.handoff.target.name", new_agent.name)
```
MikeGoldsmith, line 488: FYI - if `original_on_invoke` raises, `target.name` will never be set. Should we document that `target.name may` be absent if the handoff was interrupted?

MikeGoldsmith (approved):

model/gen-ai/registry.yaml
Krishnachaitanyakc: Makes sense. Going with sibling and documented on the execute_tool span definition with the rationale.

reference/scenarios/openai-agents/scenario.py
Krishnachaitanyakc: Good point, documented on target.name that it 'may' be absent if the handoff was interrupted before the target was determined.

trask:

Create `changelog.d/98.enhancement.md` containing the change log entry, e.g.:

```md
Add `gen_ai.agent.handoff.source.name` and `gen_ai.agent.handoff.target.name` attributes and model agent-to-agent handoffs as `execute_tool` spans. The handoff tool call (e.g., `transfer_to_<agent>`) is emitted as a child `execute_tool` span of the source agent's `invoke_agent` span, carrying the source and target agent names. Demonstrated in the OpenAI Agents reference scenario, which emits the source from `public_agent.name` and the target from the `on_invoke_handoff` return value at the SDK handoff invocation boundary.
```

Thanks!

Krishnachaitanyakc:
@trask can you review this when you get a chance?

AgentGymLeader:
Like the direction of folding handoff into `execute_tool`, and scoping `invocation.trigger` down to what's actually derivable from typed state makes sense.

Couple of things:

1. The delegation link lives entirely on the `execute_tool` span's `handoff.target.name`, and the receiving `invoke_agent` is a sibling, right? In a run with concurrent handoffs, or where two agents share a name, joining the edge back to the right `invoke_agent` span comes down to name matching. Did you consider a span link from `execute_tool` to the target `invoke_agent`, or a stable agent id separate from the name? The blast-radius query in the motivation feels like it'd get ambiguous on name collisions too.
2. Since the link sits in attributes rather than parentage, if that span gets dropped by sampling you keep both `invoke_agent` spans but lose the edge between them. Is that intended?

Both are edge-case-y, so feel free to punt if it's out of scope.

MikeGoldsmith (approved):

Krishnachaitanyakc:
@trask @lmolkova can we add this to the merge queue? Anything else for me to address?

Krishnachaitanyakc:
@trask @lmolkova gentle bump on this, can we add this to the merge queue? Anything else for me to address?

Krishnachaitanyakc:
@trask @lmolkova gentle bump on this, can we add this to the merge queue? Anything else for me to address?

@MikeGoldsmith can you help me get this merged?

nikhilNava:
Hi @Krishnachaitanyakc - thanks for this PR.

+1 to @AgentGymLeader point in https://github.com/open-telemetry/semantic-conventions-genai/pull/98#issuecomment-4729151864 about the edge relying on name matching.

The handoff creates 3 spans:

1.  invoke_agent triage-agent 
2.  execute_tool transfer_to_billing_agent  (under 1)
3.  invoke_agent billing-agent 

Problem: #3 sits beside #2, not under it — so no parent→child line connects the handoff to the agent it handed off to.

• trace_id — shared by all spans in the run, so it can't tell which handoff went to which agent when there are several.
• parent_span_id — would be the clean link, but #3's parent is the run root, not #2. So it's useless here.

Result: the only way to connect them is matching the name — which breaks under concurrent handoffs or duplicate names.

Proposal: add  gen_ai.agent.handoff.{source,target}.id  alongside  .name mirroring  gen_ai.agent.id  on  invoke_agent , as a stable join key.

cc @singankit, @trask

Krishnachaitanyakc:
Thanks @AgentGymLeader @nikhilNava. You're both right that the name edge is ambiguous under concurrent handoffs or duplicate names, and that it's lost if the execute_tool span is sampled out.

The reason it's keyed on name today: in the one framework here with a typed handoff primitive (OpenAI Agents SDK), agents have a name and no stable id. Agent/AgentBase carry no id field, and the SDK's own handoff tracing records the edge as name strings (HandoffSpanData(from_agent: str, to_agent: str)). gen_ai.agent.id is the hosted agent resource id (Bedrock ARN, assistant id), which framework agents don't have, so there's nothing to mirror in this scenario, and synthesizing one would fail the runtime-derivable bar.

A span link is the correct primitive and the pattern this repo already uses for MCP, but it isn't derivable here either. The execute_tool handoff span ends during the current turn, before the target's invoke_agent span is created on the next run loop turn, so there's no target SpanContext to link to, and the SDK doesn't carry the triggering span into the callee (matches @lmolkova's earlier point). A resource id also wouldn't disambiguate two concurrent invocations of the same agent, only a span id would.

So name is the honest ceiling of what's derivable now. I'll add a note on the handoff attributes calling out the correlation limitation, and leave room for gen_ai.agent.handoff.{source,target}.id or a span link as a follow-up once a framework exposes a stable agent id or the triggering context. Sound reasonable?

trask:
Superseded by https://github.com/open-telemetry/semantic-conventions-genai/pull/447?

PR #107, opened 2026-05-07
Add `agent.name` sampling

lzchen:
From https://github.com/open-telemetry/opentelemetry-python-contrib/pull/4553#discussion_r3203899201

## Description

Specifies `gen_ai.agent.name` as a relevant attribute for sampling. From pr comment: From https://github.com/open-telemetry/opentelemetry-python-contrib/pull/4553#discussion_r3203899201

## Motivation

Found out this attribute to be relevant to sampling for agent invocations during implementation.

## Prototype

From [comment](https://github.com/open-telemetry/semantic-conventions-genai/pull/107#pullrequestreview-4254187378), existing reference implementations show the existence of `gen_ai.agent.name` attribute. Implementations of agent invocations should pass in the attribute for sampling decisions when creating spans.

## Checklist

- [x ] Motivation section filled in above
- [ ] Reference instrumentation and scenarios updated for affected libraries
- [x ] Changelog entry added under `Unreleased` in `CHANGELOG.md` for any change to the conventions that a consumer would care about. Editorial changes (typos, pure rewording, repo tooling) don't need an entry.

See [CONTRIBUTING.md]

PR #124, opened 2026-05-08
Capture attributes on span start where possible

trask:
Follow-up to https://github.com/open-telemetry/semantic-conventions-genai/pull/107#pullrequestreview-4254187378


.github/skills/reference/SKILL.md
```diff
@@ -42,10 +42,11 @@ If the value would have to be guessed, carried forward from an unrelated call, o
...
 - When the library's public entry point does not expose the operation directly (for example, an internal-only span boundary inside a library helper), it is acceptable to patch the library's private methods to open spans around them. The scenario itself must still invoke the **public** API. Do not call private methods directly from the scenario.
 - Emit attributes inline at the span or activity site.
 - Keep request, derived, and response attributes close together.
+- Do not delay sampler-visible request attributes with `span.set_attribute` / `activity.SetTag` when their values are already available at span or activity creation time. This includes operation name, provider, request model and parameters, server address/port, input messages, system instructions, tool definitions, agent/workflow identity, and tool call arguments. Keep response values, generated IDs, token usage, output messages, retrieval results, and tool results after the operation that produces them.
```
lmolkova: > input messages, system instructions, tool definitions, agent/workflow identity, and tool call arguments

why? those are expensive to obtain - we should not collect them before span/ activity starts for perf reasons

trask: good point 👍

trask: changed it to split based on simple / complex attribute type, do you think that distinction works? another option would be to only pass sampling relevant attributes to span start

lmolkova: > another option would be to only pass sampling relevant attributes to span start

that would be my preference

reference/scenarios/anthropic/scenario.py
```diff
@@ -30,16 +30,18 @@ def run_chat():
...
+    span_attributes = {
+        "gen_ai.operation.name": "chat",
+        "gen_ai.provider.name": "anthropic",
+        "gen_ai.request.model": request_model,
+    }
+    if host:
+        span_attributes["server.address"] = host
+    if port is not None:
+        span_attributes["server.port"] = port
+    with _reference_tracer.start_as_current_span("chat claude-sonnet-4-20250514", attributes=span_attributes) as span:
```
lmolkova, line 43:
```suggestion
    with _reference_tracer.start_as_current_span(f"chat {request_model}", attributes=span_attributes) as span:
```

lmolkova (approved):

Issue #139, opened 2026-05-11
Clarify overlap between GenAI memory records and retrieval documents


nagkumar91:
## Context

During review of `open-telemetry/semantic-conventions#3250`, reviewers noted that the proposed `gen_ai.memory.records` attribute and the existing `gen_ai.retrieval.documents` attribute have very similar structure.

Original review comment: https://github.com/open-telemetry/semantic-conventions/pull/3250#discussion_r3147458454

Relevant schemas:

- `gen_ai.memory.records`: https://github.com/open-telemetry/semantic-conventions-genai/blob/main/docs/gen-ai/gen-ai-memory-records.json
- `gen_ai.retrieval.documents`: https://github.com/open-telemetry/semantic-conventions-genai/blob/main/docs/gen-ai/gen-ai-retrieval-documents.json

## Question

Memory search spans (`gen_ai.operation.name = search_memory`) can return records that look similar to retrieval documents: content plus optional identifiers, scores, and metadata. At the same time, memory operations also include store lifecycle and record mutation/deletion operations that are not generic retrieval.

## Proposed follow-up

Clarify during stabilization whether:

1. memory search results should always use `gen_ai.memory.records`,
2. memory search results may/should reuse `gen_ai.retrieval.documents` when the memory store behaves like a retriever, or
3. the two attributes should remain separate with explicit guidance explaining the boundary.

Related migration PR/source breadcrumb: open-telemetry/semantic-conventions#3250.

Necmttn:
I would split the operation namespace from the result item shape.

Memory lifecycle operations still need memory-specific spans: create, update, delete, compact, expire. But memory search results can use the same evidence shape as retrieval results: record/document id, store id, source URI, snapshot/version, score, content reference, and redaction state.

That keeps search_memory queryable beside retrieval without duplicating schemas. The boundary becomes: retrieval-like evidence for returned items, memory-specific conventions for mutations and retention behavior.

---

_Generated with [ax](https://github.com/Necmttn/ax)._

PR #140, opened 2026-05-11
Add GenAI memory operation conventions

nagkumar91:
## Summary

Migrates the narrowed GenAI memory operations proposal from open-telemetry/semantic-conventions#3250 into this repository.

Adds:

- memory operation names: `create_memory_store`, `search_memory`, `update_memory`, `delete_memory`, `delete_memory_store`
- memory spans for memory store lifecycle and memory record CRUD/search operations
- attributes: `gen_ai.memory.store.id`, `gen_ai.memory.record.id`, `gen_ai.memory.query.text`, `gen_ai.memory.records`
- `docs/gen-ai/gen-ai-memory-records.json` with `content` intentionally allowing any JSON value
- Google ADK reference scenario coverage for `upsert_memory` and `search_memory` spans (matching ADK's `add_session_to_memory` / `search_memory` API surface); AWS Bedrock AgentCore scenario covers `create_memory_store`, `create_memory`, `update_memory`, `search_memory`, `delete_memory`, `delete_memory_store`
- follow-up issue for memory search vs retrieval documents overlap: #139

## Review feedback applied from old PR

- Aligned `gen_ai.memory.records` YAML and JSON schema so record `content` can be any JSON value.
- Kept JSON-schema property details in the JSON schema instead of duplicating them in the YAML note.
- Backticked `gen_ai.memory.store.id` in registry guidance.
- Removed the repeated explicit “do not capture by default” registry wording; span usage marks content-bearing attributes as opt-in.
- Added reference scenario/report coverage under `reference/`.

## Framework evidence

| Framework | Memory operations represented | Evidence |
| --- | --- | --- |
| Google ADK | Yes: add/search memory APIs | [`BaseMemoryService` and `InMemoryMemoryService` define `add_session_to_memory`, `add_events_to_memory`, `search_memory`](https://github.com/google/adk-python/blob/main/src/google/adk/memory/in_memory_memory_service.py) |
| AWS Bedrock AgentCore | Yes: create/retrieve/delete memory records | [AgentCore memory quickstart uses memory record CRUD APIs](https://aws.github.io/bedrock-agentcore-starter-toolkit/user-guide/memory/quickstart.html) |
| Azure AI Foundry | Yes: search/update/delete memory scopes | [Azure AI Foundry agent memory concepts describe memory search, update, and scope deletion](https://learn.microsoft.com/en-us/azure/ai-foundry/agents/concepts/what-is-memory) |
| Mem0 | Yes: add/search/delete/delete_all memories | [Mem0 docs show `add`, `search`, `delete`, and `delete_all`](https://docs.mem0.ai/) |
| CrewAI | Yes: remember/recall/forget memory model | [CrewAI memory docs describe short-term, long-term, entity, and user memory behavior](https://docs.crewai.com/concepts/memory) |
| Letta / MemGPT | Yes: archival memory insert/search and memory blocks | [Letta docs describe archival memory and blocks](https://docs.letta.com/) |

## Validation

- `make generate-all`
- `make check-policies`
- `cd reference && uv run run-scenario google-adk`
- `cd reference && uv run update-reports`
- `cd reference && uv tool run --from ruff ruff check src/semconv_genai scenarios/google-adk`
- `cd reference && uv tool run --from ruff ruff format --check src/semconv_genai scenarios/google-adk`

## Notes

This intentionally keeps the migrated scope narrow and leaves contentious overlap with `gen_ai.retrieval.documents` to follow-up issue #139.


reference/scenarios/google-adk/scenario.py
trask, line 1: I ran the PR against the [Reference Scenario Evaluation Skill](https://github.com/open-telemetry/semantic-conventions-genai/blob/main/.github/skills/evaluate-reference/SKILL.md), can you review its findings? thanks!

# PR 140 Reference Scenario Evaluation

Evaluation date: 2026-05-11

PR: <https://github.com/open-telemetry/semantic-conventions-genai/pull/140>

## Scope

PR 140 adds GenAI memory operation conventions:

- `create_memory_store`
- `search_memory`
- `update_memory`
- `delete_memory`
- `delete_memory_store`

It also adds memory attributes including `gen_ai.memory.store.id`, `gen_ai.memory.record.id`, `gen_ai.memory.query.text`, and `gen_ai.memory.records`.

The reference implementation updates the Google ADK scenario and generated reports for the new memory span types.

## Credible Reference Coverage

Google ADK credibly demonstrates `update_memory` and `search_memory` in `reference/scenarios/google-adk/scenario.py`.

The `update_memory` span wraps the public `InMemoryMemoryService.add_session_to_memory(session)` call. The emitted `gen_ai.memory.records` value is built from the current `Session` and `Event` passed into that SDK call, so the stored content and author/session metadata are directly observable at the instrumentation point.

The `search_memory` span wraps the public `InMemoryMemoryService.search_memory(app_name=..., user_id=..., query=...)` call. `gen_ai.memory.query.text` comes directly from the `query` argument, and returned memory content comes from `response.memories`, so those fields are directly observable from the current call.

The empty Google ADK coverage for `delete_memory` and `delete_memory_store` appears honest. ADK's `InMemoryMemoryService` exposes add/search APIs, but no delete-memory or delete-store public API in the scenario's installed version.

## Semantically Derivable Fields

`gen_ai.memory.store.id` on the Google ADK `update_memory` and `search_memory` spans is derivable from the ADK memory scope. The scenario uses `store_id = f"{app_name}/{user_id}"`, and the ADK in-memory implementation internally keys memory by that same app/user scope.

This mapping should be documented for ADK-specific instrumentation because it includes `user_id`, which may be high-cardinality in real telemetry.

## Too App-Specific Or Cross-Call

### Google ADK create-store id

The Google ADK `create_memory_store` span wraps `InMemoryMemoryService()` construction but emits `gen_ai.memory.store.id = "test_app/test_user"`.

That constructor has no `app_name` or `user_id` argument and does not create a specific app/user store. The emitted store id comes from future scenario variables rather than the current create-store call boundary.

Recommendation: either omit `gen_ai.memory.store.id` and name the span `create_memory_store`, or treat Google ADK as not credibly demonstrating create-store id coverage.

### Google ADK synthesized search record id

The Google ADK search span emits memory record ids using `memory.id or f"{session.id}/{index}"`.

In ADK 1.32.0, `MemoryEntry.id` exists on the type, but `InMemoryMemoryService.search_memory()` leaves it unset. The fallback id is synthesized from prior scenario state and result order, not from a stable library-owned record id in the search response.

Recommendation: include `id` in `gen_ai.memory.records` only when `memory.id` is present; otherwise omit the id field for ADK in-memory search results.

## Missing Supporting Library Implementations

### AWS Bedrock AgentCore

AWS Bedrock AgentCore looks like a missed supporting-library implementation.

The PR body lists AWS Bedrock AgentCore as framework evidence, and the existing `aws-bedrock-agent` reference environment exposes concrete Boto3 memory APIs:

- `BatchCreateMemoryRecords`
- `BatchUpdateMemoryRecords`
- `DeleteMemoryRecord`
- `RetrieveMemoryRecords`
- `CreateMemory`
- `DeleteMemory`

Those APIs can cover memory record creation/update/search/delete and memory store create/delete operations that Google ADK cannot demonstrate.

Recommendation: add an AWS Bedrock AgentCore reference scenario, or explicitly document why AgentCore memory coverage is out of scope for this PR.

### CrewAI

CrewAI is a plausible but weaker gap.

The installed `Crew` constructor has a first-class `memory` parameter, but the current scenario constructs crews without enabling memory. This may support memory behavior through public CrewAI APIs, but it needs additional investigation before treating it as required coverage.

Recommendation: leave CrewAI out only if the PR scope is intentionally limited to APIs with explicit memory CRUD/search boundaries, or add a small CrewAI memory scenario if the proposal wants framework-level memory behavior coverage.

## Evaluation Recommendation

- `fix implementation`: remove or change the Google ADK create-store `gen_ai.memory.store.id` emission, because the current constructor call does not expose that store id.
- `fix implementation`: stop synthesizing Google ADK search memory record ids when `MemoryEntry.id` is absent.
- `add reference for supporting library`: add AWS Bedrock AgentCore memory coverage, especially for delete spans and memory-store lifecycle spans.
- `leave unchanged; honest capture gap`: keep Google ADK delete-memory and delete-memory-store reports empty unless ADK exposes public delete APIs in a future version.

nagkumar91: Thanks for running the eval @trask . I've addressed the findings:

- Updated Google ADK `create_memory_store` to stop emitting a synthetic `gen_ai.memory.store.id`; the span is now named `create_memory_store`.
 - Updated Google ADK search results to include `id` only when `MemoryEntry.id` is present, instead of synthesizing a fallback id.
 - Added a separate AWS Bedrock AgentCore reference scenario covering `CreateMemory`, `BatchCreateMemoryRecords`, `BatchUpdateMemoryRecords`, `RetrieveMemoryRecords`, `DeleteMemoryRecord`, and `DeleteMemory`.
 - Regenerated the reference data and reports. ADK delete/delete-store remain honest gaps because ADK does not expose public delete APIs for those operations.

reference/scenarios/google-adk/scenario.py
```diff
@@ -288,6 +288,90 @@ async def _run():
...
+                    "content": memory_text,
+                    "id": event.id,
+                    "metadata": {"author": event.author, "session_id": session.id},
+                }
+            ]
+        )
+
+        update_span_attributes = {
+            "gen_ai.operation.name": "update_memory",
+            "gen_ai.provider.name": "gcp.gemini",
```
trask: not sure what to do about provider name for InMemoryMemoryService

model/gen-ai/spans.yaml
```diff
@@ -295,6 +303,131 @@ spans:
...
+      note: |
+        **Span name** SHOULD be `update_memory {gen_ai.memory.store.id}`
+        or `update_memory` if the store id is not available.
+    brief: >
+      Describes a memory update operation that creates new or modifies existing
+      memory records in a memory store.
+    note: |
+      The `gen_ai.operation.name` SHOULD be `update_memory`.
+
+      This operation covers both creating new and updating existing memory records.
```
trask: wondering about upsert vs create/update (or maybe all 3?).....

--------------

Research for review of [open-telemetry/semantic-conventions-genai#140](https://github.com/open-telemetry/semantic-conventions-genai/pull/140),
specifically the question of whether collapsing `create_memory` and
`update_memory` into a single `update_memory` operation loses information.

## Distinct create vs. update APIs

| Library | Create | Update | Notes |
| --- | --- | --- | --- |
| **Mem0** | `POST /v3/memories/add/` (extracts from messages) | `PUT /v1/memories/{memory_id}` (replace text/metadata on a known id) | Cleanly separate. Different verbs, different inputs, different semantics. |
| **Letta** | `POST /v1/archives/{archive_id}/passages` (create new passage; embeddings generated) | `PATCH /v1/archives/.../passages/{id}` (modify existing passage) | Cleanly separate. |
| **AWS Bedrock AgentCore** | `BatchCreateMemoryRecords` (writes new records by `requestIdentifier`) | `BatchUpdateMemoryRecords` (modifies by `memoryRecordId`) | Two distinct boto3 calls. This is exactly what PR #140's AgentCore reference scenario collapses into one `update_memory` op. |
| **Azure AI Foundry agents (memory items)** | `POST /memoryItems` | `PATCH /memoryItems/{id}` | Per the REST control-plane shape. |

## Unified upsert / single write entry point

| Library | Entry point | Notes |
| --- | --- | --- |
| **CrewAI** | `memory.remember(content, scope=…)` | Single API. Consolidation logic inside may *decide* keep / update / delete / insert_new based on similarity, but the caller never picks "create vs update". |
| **Google ADK** | `MemoryService.add_session_to_memory(session)` / `add_events_to_memory(events)` | Single API. The Vertex Memory Bank backend internally may generate or ingest, but it's one call from the consumer's POV. No update-by-id at all. |
| **LangChain memory** (chat history / vector memory) | `save_context(inputs, outputs)` or vectorstore `add_texts` | Single append; no public update-by-id semantics. |
| **LlamaIndex `ChatMemoryBuffer` / vector memory** | `put(message)` or vectorstore `add` | Append-only from the memory abstraction's POV. |
| **OpenAI Assistants v2 threads** | `messages.create` | Append-only. No update of a stored message. |

## Mixed / hybrid

| Library | Notes |
| --- | --- |
| **OpenAI "Memories" feature** (newer Responses API memory) | `POST /v1/memories` to add; no public PATCH yet — closer to append-only. |
| **LangGraph store** | `put(namespace, key, value)` is an upsert by `key` — single API, but semantically covers both create and update. This is actually a case where collapsing them into one op is the natural fit. |
| **Vertex AI Memory Bank** (under ADK) | `memories.generate` (LLM-derived create) and `memories.create` (explicit create) coexist; no public update — closer to create-only with implicit consolidation. |

model/gen-ai/spans.yaml
```diff
@@ -295,6 +303,131 @@ spans:
...
+      - ref_group: attributes.gen_ai.memory.client
+      - ref: gen_ai.provider.name
+        requirement_level: required
+      - ref: gen_ai.memory.store.id
+        requirement_level:
+          conditionally_required: if applicable
+      - ref: gen_ai.memory.query.text
+        requirement_level: opt_in
+      - ref: gen_ai.memory.records
+        requirement_level: opt_in
```
trask: I forget, did we punt on record count? would be nice to know if any records were found without having to opt-in to PII sensitive data

reference/scenarios/google-adk/scenario.py
```diff
@@ -288,6 +288,90 @@ async def _run():
...
+    from google.adk.memory.in_memory_memory_service import InMemoryMemoryService
+    from google.adk.sessions.session import Session
+    from google.genai import types
+
+    print("  [memory] ADK in-memory memory service (reference implementation)")
+
+    app_name = "test_app"
+    user_id = "test_user"
+    session_id = "session_memory_1"
+    store_id = f"{app_name}/{user_id}"
```
trask, line 306: I'm not sure whether this fits low-cardinality for being used in span names, @lmolkova?

https://opentelemetry.io/docs/specs/otel/trace/api/#span

> The span name SHOULD be the most general string that identifies a (statistically) interesting class of Spans, rather than individual Span instances while still being human-readable. That is, “get_user” is a reasonable name, while “get_user/314159”, where “314159” is a user ID, is not a good name due to its high cardinality. Generality SHOULD be prioritized over human-readability.

model/gen-ai/registry.yaml
```diff
@@ -508,6 +528,59 @@ attributes:
...
+    type: string
+    brief: The unique identifier of the memory record.
+    examples: ["mem_5j66UpCpwteGg4YSxUnt7lPY"]
+    stability: development
+  - key: gen_ai.memory.query.text
+    type: string
+    brief: The search query used to retrieve memories.
+    note: |
+      > [!Warning]
+      > This attribute may contain sensitive information.
```
trask, line 564: let's make opt-in requirement level, gated by `OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT` similar to others

model/gen-ai/registry.yaml
```diff
@@ -508,6 +528,59 @@ attributes:
...
+    type: any
+    brief: The memory records stored or retrieved in a memory operation.
+    note: |
+      Instrumentations MUST follow [Memory records JSON schema](/docs/gen-ai/gen-ai-memory-records.json).
+      When the attribute is recorded on events, it MUST be recorded in structured
+      form. When recorded on spans, it MAY be recorded as a JSON string if structured
+      format is not supported and SHOULD be recorded in structured form otherwise.
+
+      > [!Warning]
+      > This attribute may contain sensitive information including user/PII data.
```
trask, line 580: let's make opt-in requirement level, gated by `OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT` similar to others

docs/gen-ai/gen-ai-memory-records.json
```diff
@@ -0,0 +1,38 @@
+{
+  "$defs": {
+    "MemoryRecord": {
+      "additionalProperties": true,
+      "description": "Represents a single memory record stored in or retrieved from a memory store.",
+      "properties": {
+        "content": {
```
trask: should this be required?

reference/scenarios/google-adk/scenario.py
nagkumar91: Fixed in 9084ad3. `gen_ai.provider.name` is now conditionally required on memory spans only when the operation is handled by a named external GenAI provider or service. The Google ADK `InMemoryMemoryService` reference no longer sets `gen_ai.provider.name`; the AgentCore reference still sets `aws.bedrock`.

model/gen-ai/spans.yaml
nagkumar91: Fixed in 9084ad3. Added `gen_ai.memory.record.count` as a non-content attribute and made it recommended on `search_memory` and `update_memory`. The note defines search count as returned records and update count as records attempted for create/modify. Updated ADK and AgentCore reference scenarios and regenerated reports.

reference/scenarios/google-adk/scenario.py
nagkumar91: Fixed in 9084ad3. Memory span names now use constant names (`create_memory_store`, `search_memory`, `update_memory`, `delete_memory`, `delete_memory_store`) and the spec notes that `gen_ai.memory.store.id` is intentionally omitted because store identifiers may be high-cardinality. Updated the reference scenario span names accordingly.

model/gen-ai/registry.yaml
nagkumar91: Fixed in 9084ad3. Added registry guidance that `gen_ai.memory.query.text` should not be captured by default and should be gated by an explicit user opt-in, for example `OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT`.

model/gen-ai/registry.yaml
nagkumar91: Fixed in 9084ad3. Added the same explicit opt-in capture guidance for `gen_ai.memory.records`, in addition to the existing sensitive/PII warning and opt-in requirement level on spans.

model/gen-ai/spans.yaml
nagkumar91: Implemented this direction in e9d1063. The PR now has separate `create_memory`, `update_memory`, and `upsert_memory` operations: AgentCore `BatchCreateMemoryRecords` maps to `create_memory`, AgentCore `BatchUpdateMemoryRecords` maps to `update_memory`, and Google ADK `add_session_to_memory` maps to `upsert_memory` because the caller invokes a unified API that may create/update/consolidate records without choosing which. Regenerated docs, schema snapshot, reference data, and reports.

nagkumar91: Implemented and pushed the agreed option 3 split. We now have `create_memory` for explicit creation, `update_memory` for modifications of known existing records, and `upsert_memory` for unified APIs where the caller does not choose create vs update. Reference coverage now maps AgentCore create/update APIs separately and Google ADK to `upsert_memory`.

docs/gen-ai/gen-ai-memory-records.json
nagkumar91: Updated this in 8b40879. `gen_ai.memory.records` remains opt-in as the enclosing attribute, but once instrumentations opt in to emit it, each memory record object now requires `content`; `id`, `metadata`, and `score` remain optional. I also updated the non-normative model so the checked-in JSON schema stays aligned with it.

docs/gen-ai/gen-ai-spans.md
```diff
@@ -12,6 +12,11 @@ linkTitle: Spans
   - [Inference](#inference)
   - [Embeddings](#embeddings)
   - [Retrievals](#retrievals)
+  - [Create memory store](#create-memory-store)
+  - [Search memory](#search-memory)
+  - [Update memory](#update-memory)
+  - [Delete memory](#delete-memory)
+  - [Delete memory store](#delete-memory-store)
```
trask: need to add the 2 new ops Create memory and Upsert memory

CHANGELOG.md
```diff
@@ -10,6 +10,9 @@
 
 ### 💡 Enhancements 💡
 
+- Add GenAI memory operation spans and attributes for memory store lifecycle, create/update/upsert record
+  operations, and record counts.
```
trask:
```suggestion
- Add GenAI memory operation spans and attributes for memory store lifecycle (create/delete),
  memory record create/update/upsert/search/delete operations, and record counts.
```

reference/scenarios/aws-bedrock-agentcore/scenario.py
```diff
@@ -0,0 +1,363 @@
...
+        endpoint_url=MOCK_BASE_URL,
+        region_name="us-east-1",
+        aws_access_key_id="mock",
+        aws_secret_access_key="mock",
+    )
+
+
+def run_create_memory_store_reference(client, stubber):
+    """Scenario: AgentCore CreateMemory."""
+    print("  [create_memory_store] Bedrock AgentCore CreateMemory")
```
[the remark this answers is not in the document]
nagkumar91: Updated in a225a5b. The AgentCore reference output/docstrings now use the semantic operation names `create_memory_store` and `delete_memory_store` for the store lifecycle cases.

docs/gen-ai/gen-ai-spans.md
nagkumar91: Updated in a225a5b. Added Create memory and Upsert memory to the spans document TOC.

CHANGELOG.md
nagkumar91: Applied the suggested changelog wording in a225a5b.

docs/gen-ai/gen-ai-spans.md
```diff
@@ -12,6 +12,13 @@ linkTitle: Spans
   - [Inference](#inference)
   - [Embeddings](#embeddings)
   - [Retrievals](#retrievals)
+  - [Create memory store](#create-memory-store)
+  - [Search memory](#search-memory)
+  - [Create memory](#create-memory)
+  - [Update memory](#update-memory)
+  - [Upsert memory](#upsert-memory)
```
trask: I think these sections are missing below. Also, might be nice to group the memory spans hierarchically since there's a lot of them.

nagkumar91: Updated in 00956dd. The spans doc now has a parent Memory section with the memory spans grouped underneath it, and the missing generated Create memory and Upsert memory sections are included in the body.

model/gen-ai/spans.yaml
```diff
@@ -295,6 +303,198 @@ spans:
...
+      note: |
+        **Span name** SHOULD be `create_memory_store`.
+    brief: >
+      Describes creation or initialization of a memory store.
+    note: |
+      The `gen_ai.operation.name` SHOULD be `create_memory_store`.
+
+      **Span name** SHOULD be `create_memory_store`.
+      Memory operation span names intentionally omit `gen_ai.memory.store.id` because memory store
+      identifiers may be high-cardinality.
```
trask: (and same in other places)
```suggestion
```

model/gen-ai/registry.yaml
```diff
@@ -508,6 +536,76 @@ attributes:
       > This attribute may contain sensitive information.
     examples: ["What is the capital of France?", "weather in Paris"]
     stability: development
+  - key: gen_ai.memory.store.id
+    type: string
+    brief: The unique identifier of the memory store.
+    note: >
+      Instrumentations for individual memory components SHOULD pick a low-cardinality identifier and
+      SHOULD NOT set `gen_ai.memory.store.id` if no such identifier exists for this component. Semantic
```
trask: I don't think we need the low-cardinality requirement now that it's no longer part of span name.
```suggestion
      Semantic
```

model/gen-ai/spans.yaml
nagkumar91: Accepted in 9c758c1. Removed the repeated memory span note from all memory operation span definitions and regenerated the derived docs/snapshot.

model/gen-ai/registry.yaml
nagkumar91: Accepted in 9c758c1. Simplified the `gen_ai.memory.store.id` note to only ask component-specific conventions to document what the value maps to, and regenerated the derived docs/snapshot.

model/gen-ai/spans.yaml
```diff
@@ -295,6 +303,184 @@ spans:
         requirement_level:
           conditionally_required: when applicable
 
+  - type: gen_ai.create_memory_store.client
+    kind: client
```
trask, line 299: Should we add the same span-kind allowance here that `gen_ai.inference.client` has for same-process calls? ADK’s in-memory memory service seems like a case where `INTERNAL` is more accurate, but the generated docs for these new memory spans currently say only `Span kind SHOULD be CLIENT`.

docs/gen-ai/gen-ai-spans.md
```diff
@@ -12,6 +12,14 @@ linkTitle: Spans
...
   - [Embeddings](#embeddings)
   - [Retrievals](#retrievals)
+  - [Memory](#memory)
+    - [Create memory store](#create-memory-store)
+    - [Search memory](#search-memory)
+    - [Create memory](#create-memory)
+    - [Update memory](#update-memory)
+    - [Upsert memory](#upsert-memory)
+    - [Delete memory](#delete-memory)
+    - [Delete memory store](#delete-memory-store)
```
trask: @lmolkova I'd like your thoughts on this:

Alternative shape to consider: collapse the seven memory spans into one `gen_ai.memory.client` span dispatched by `gen_ai.operation.name`, matching how MCP handles JSON-RPC methods. Sketch + tradeoffs below.

```yaml
spans:
  - type: gen_ai.memory.client
    kind: client
    name:
      note: |
        **Span name** SHOULD be `{gen_ai.operation.name}`.
    brief: >
      Describes an operation against a GenAI memory store or its records.
    note: |
      Covers the memory-store lifecycle (`create_memory_store`,
      `delete_memory_store`) and memory-record operations
      (`search_memory`, `create_memory`, `update_memory`, `upsert_memory`,
      `delete_memory`). `gen_ai.operation.name` distinguishes them.
    stability: development
    attributes:
      - ref_group: attributes.gen_ai.memory.client
      - ref: gen_ai.provider.name
        requirement_level:
          conditionally_required: if the operation is handled by a named external GenAI provider or service
      - ref: gen_ai.memory.store.id
        requirement_level:
          conditionally_required: if applicable
      - ref: gen_ai.memory.record.id
        requirement_level:
          conditionally_required: when available
      - ref: gen_ai.memory.record.count
        requirement_level:
          conditionally_required: when available
      - ref: gen_ai.memory.query.text
        requirement_level: opt_in
      - ref: gen_ai.memory.records
        requirement_level: opt_in
```

**Pros**

- **Aligns with core repo guidance.** [`docs/how-to-write-conventions/README.md`](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/how-to-write-conventions/README.md) says split on different kind or "significantly different set of attributes" — these seven share kind and ~5/7 share their attribute set verbatim.
- **Matches the MCP precedent in this repo.** `mcp.client`/`mcp.server` already cover every JSON-RPC method via `mcp.method.name`, with per-method attributes (e.g. `mcp.resource.uri`) declared `conditionally_required` on the unified span — the exact pattern this proposal applies to memory.

**Cons**

- **Per-op applicability of attributes moves from structured YAML into free-text notes.** Today, "`gen_ai.memory.query.text` only applies to `search_memory`" is a structural fact (the attribute appears only on that span). Under unification, that's a sentence in the span's note. Tooling and reviewers that want per-op expected-attribute sets have to parse prose instead of reading the YAML directly.

lmolkova: I think it's a trade-off between "how granular we want spans to be" vs "how many spans we can afford to define".  E.g. in case of cloud providers it's impossible to define and maintain spans per each REST / RPC operation type.

We could evolve tooling to support formal conditions on attributes like `if operation_name == "update_memory"` then record id is required.

In this case, I like the idea of grouping things together, but on conceptual level:
- control plane (memory store operations) - in the long term these might have substantially different  set of attributes
- memory CRUD

Having said this, we didn't create this separation for DB operations and group them all together.

TL;DR: I like one span in this case and think we should be evolving tooling to support formal conditions.

nagkumar91: Updated in 03bf644. I collapsed the seven operation-specific memory span conventions into a single `gen_ai.memory.client` span keyed by `gen_ai.operation.name`, kept the operation names for dispatch, regenerated docs/schema/reference reports, and consolidated the reference coverage under one Memory report.

model/gen-ai/spans.yaml
nagkumar91: Updated in 5347476. The unified memory span now matches the inference guidance: `CLIENT` is recommended for remote or instrumented-protocol calls, and `INTERNAL` is allowed for same-process memory systems such as ADK in-memory memory.

docs/gen-ai/non-normative/models.ipynb
```diff
@@ -469,6 +469,58 @@
...
+    "\n",
+    "class MemoryRecords(RootModel[List[MemoryRecord]]):\n",
+    "    \"\"\"\n",
+    "    Represents the list of memory records stored in or retrieved from a memory store.\n",
+    "    \"\"\"\n",
+    "    pass\n",
+    "\n",
+    "# Print the JSON schema for the MemoryRecords model\n",
+    "with open(\"../gen-ai-memory-records.json\", \"w\") as file:\n",
+    "    print(json.dumps(MemoryRecords.model_json_schema(), indent=4), file=file)"
```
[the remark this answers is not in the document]
nagkumar91: Fixed in 4beb7fd. Reformatted `gen-ai-memory-records.json` with 4-space indentation so it matches the notebook generator output and won’t become dirty when regenerated.

reference/scenarios/google-adk/scenario.py
```diff
@@ -288,6 +288,85 @@ async def _run():
...
+    memory_text = "User prefers vegetarian meals and dark mode."
+    query_text = "vegetarian meals"
+
+    async def _run():
+        create_span_attributes = {
+            "gen_ai.operation.name": "create_memory_store",
+        }
+        with _reference_tracer.start_as_current_span("create_memory_store", attributes=create_span_attributes):
+            memory_service = InMemoryMemoryService()
```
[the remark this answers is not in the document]
trask: Agreed — dropped the `create_memory_store` span from the google-adk scenario in bc38c7d. ADK's `InMemoryMemoryService` has no public store-lifecycle API and the constructor produces no library-owned store identifier, so this was a capture gap rather than credible coverage. `create_memory_store` remains demonstrated by [reference/scenarios/aws-bedrock-agentcore/](reference/scenarios/aws-bedrock-agentcore/scenario.py) where the returned `memoryId` is captured directly from the SDK response.

Also annotated the surviving `store_id = f"{app_name}/{user_id}"` as `derivable` per the evaluate-reference rubric, since `(app_name, user_id)` is the tuple ADK's `InMemoryMemoryService` itself uses to key its internal store.

reference/scenarios/aws-bedrock-agentcore/scenario.py
```diff
@@ -0,0 +1,364 @@
...
+        )
+
+
+def main():
+    print("=== Reference Implementation: AWS Bedrock AgentCore Memory ===")
+    tp, lp, mp = setup_otel()
+
+    control_client = create_agentcore_control_client()
+    data_client = create_agentcore_data_client()
+    with Stubber(control_client) as control_stubber, Stubber(data_client) as data_stubber:
```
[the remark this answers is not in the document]
trask: Good point — the scenario now uses a real `bedrock_agentcore` mock-server blueprint covering both control (`CreateMemory`, `DeleteMemory`) and data plane (`BatchCreateMemoryRecords`, `BatchUpdateMemoryRecords`, `RetrieveMemoryRecords`, `DeleteMemoryRecord`) at the URL paths declared by the botocore service model. `Stubber` is gone; boto3 reaches the mock over real HTTP, so `server.address`/`server.port` are observed from actual requests. `data.json` diff is empty after the change.

Pushed in 69c3bb6.

reference/scenarios/aws-bedrock-agentcore/scenario.py
```diff
@@ -0,0 +1,268 @@
...
+from opentelemetry.trace import SpanKind
+from reference_shared import flush_and_shutdown, mock_server_host_port, reference_tracer, setup_otel
+
+MOCK_BASE_URL = os.environ["MOCK_LLM_URL"]
+
+_reference_tracer = reference_tracer()
+
+MEMORY_ID = "mem-abc123456"
+MEMORY_NAME = "customer-support-memory"
+MEMORY_RECORD_ID = "memrec-0123456789abcdef0123456789abcdef0123"
```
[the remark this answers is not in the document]
trask: Done. The `memoryRecordId` returned by `batch_create_memory_records` now flows into the update payload and the delete call; the `MEMORY_RECORD_ID` constant was removed. 9883f06.

reference/src/semconv_genai/mock_server/bedrock_agentcore.py
```diff
@@ -0,0 +1,120 @@
...
+
+# Data plane ----------------------------------------------------------------
+
+
+@bp.route("/memories/<memory_id>/memoryRecords/batchCreate", methods=["POST"])
+def batch_create_memory_records(memory_id):
+    body = request.get_json(silent=True) or {}
+    records = body.get("records", [])
+    store = _RECORDS.setdefault(memory_id, [])
+    successful = []
```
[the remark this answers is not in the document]
trask: Done. `batchCreate` no longer uses `setdefault` — it returns 404 when `memoryId` is unknown. 9883f06.

reference/src/semconv_genai/mock_server/bedrock_agentcore.py
```diff
@@ -0,0 +1,120 @@
...
+    records = body.get("records", [])
+    store = _RECORDS.setdefault(memory_id, [])
+    by_id = {r["memoryRecordId"]: r for r in store}
+    successful = []
+    for rec in records:
+        record_id = rec["memoryRecordId"]
+        if record_id in by_id:
+            by_id[record_id]["content"] = rec.get("content", by_id[record_id].get("content", {}))
+        successful.append({"memoryRecordId": record_id, "status": "SUCCEEDED"})
+    return _json({"successfulRecords": successful, "failedRecords": []})
```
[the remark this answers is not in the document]
trask: Done. Missing records now go into `failedRecords` with `errorCode=404` instead of being falsely reported as `SUCCEEDED`. 9883f06.

reference/scenarios/aws-bedrock-agentcore/scenario.py
```diff
@@ -0,0 +1,268 @@
...
+
+import boto3
+from opentelemetry.trace import SpanKind
+from reference_shared import flush_and_shutdown, mock_server_host_port, reference_tracer, setup_otel
+
+MOCK_BASE_URL = os.environ["MOCK_LLM_URL"]
+
+_reference_tracer = reference_tracer()
+
+MEMORY_ID = "mem-abc123456"
```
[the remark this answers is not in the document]
trask: Done. `create_memory`'s returned `memory.id` is now threaded through every subsequent operation; the `MEMORY_ID` constant was removed. 9883f06.

reference/src/semconv_genai/mock_server/bedrock_agentcore.py
```diff
@@ -0,0 +1,120 @@
...
+            "score": 0.93,
+            "metadata": rec.get("metadata", {"author": {"stringValue": "assistant"}}),
+        })
+    return _json({"memoryRecordSummaries": summaries})
+
+
+@bp.route("/memories/<memory_id>/memoryRecords/<record_id>", methods=["DELETE"])
+def delete_memory_record(memory_id, record_id):
+    store = _RECORDS.get(memory_id, [])
+    _RECORDS[memory_id] = [r for r in store if r["memoryRecordId"] != record_id]
```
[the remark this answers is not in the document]
trask: Done. `DeleteMemoryRecord` now returns 404 when either `memoryId` or the requested `record_id` is not present. 9883f06.

reference/src/semconv_genai/mock_server/bedrock_agentcore.py
```diff
@@ -0,0 +1,120 @@
...
+            "requestIdentifier": rec.get("requestIdentifier", ""),
+        })
+    return _json({"successfulRecords": successful, "failedRecords": []}, status=201)
+
+
+@bp.route("/memories/<memory_id>/memoryRecords/batchUpdate", methods=["POST"])
+def batch_update_memory_records(memory_id):
+    body = request.get_json(silent=True) or {}
+    records = body.get("records", [])
+    store = _RECORDS.setdefault(memory_id, [])
```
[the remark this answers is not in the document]
trask: Done. `batchUpdate` no longer uses `setdefault` — it returns 404 when `memoryId` is unknown. 9883f06.

reference/src/semconv_genai/mock_server/bedrock_agentcore.py
```diff
@@ -0,0 +1,120 @@
...
+        "updatedAt": 1735689600.0,
+    }
+    _MEMORIES[memory_id] = memory
+    _RECORDS.setdefault(memory_id, [])
+    return _json({"memory": memory}, status=201)
+
+
+@bp.route("/memories/<memory_id>/delete", methods=["DELETE"])
+def delete_memory(memory_id):
+    _MEMORIES.pop(memory_id, None)
```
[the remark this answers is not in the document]
trask: Done. `delete_memory` now returns 404 when `memoryId` is not in `_MEMORIES`. Combined with the scenario change that uses the created `memory.id`, the lifecycle is tied to real mock state. 9883f06.

reference/src/semconv_genai/mock_server/bedrock_agentcore.py
```diff
@@ -0,0 +1,138 @@
...
+    return _json({"successfulRecords": successful, "failedRecords": failed})
+
+
+@bp.route("/memories/<memory_id>/retrieve", methods=["POST"])
+def retrieve_memory_records(memory_id):
+    body = request.get_json(silent=True) or {}
+    criteria = body.get("searchCriteria", {})
+    top_k = criteria.get("topK", 5)
+    namespace = body.get("namespace", "/")
+    store = _RECORDS.get(memory_id, [])
```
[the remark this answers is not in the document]
trask: Done in b30f50b. The retrieve route now returns 404 when `memoryId` is not in `_MEMORIES`, consistent with the other data-plane endpoints.

reference/scenarios/aws-bedrock-agentcore/scenario.py
```diff
@@ -0,0 +1,269 @@
...
+        span_attributes_2["server.address"] = host
+    if port is not None:
+        span_attributes_2["server.port"] = port
+    with _reference_tracer.start_as_current_span(
+        "update_memory", kind=SpanKind.CLIENT, attributes=span_attributes_2
+    ) as span:
+        client.batch_update_memory_records(
+            memoryId=memory_id,
+            records=update_records,
+        )
```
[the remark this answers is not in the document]
trask: Declining this one. Reference scenarios under `reference/scenarios/` are not tests of the mock — they exist to demonstrate that an instrumenter could capture each `gen_ai.*` attribute from the library's public call boundary (see `reference/README.md` and `.github/instructions/reference-scenarios.instructions.md`). Asserting on `failedRecords`/`successfulRecords` would validate our own mock rather than drive any attribute emission, and real native instrumentation would still emit the span (recording the error as appropriate) rather than skip attributes on a per-record failure. The lifecycle-integrity concern is already enforced mock-side: `batchUpdate` 404s on unknown `memoryId` and reports missing record IDs in `failedRecords`, so a scenario that lost the create→update ID linkage would fail loudly via boto3 rather than silently pass.

reference/scenarios/google-adk/scenario.py
```diff
@@ -288,6 +288,89 @@ async def _run():
...
+                    "content": memory_text,
+                    "id": event.id,
+                    "metadata": {"author": event.author, "session_id": session.id},
+                }
+            ]
+        )
+
+        upsert_span_attributes = {
+            "gen_ai.operation.name": "upsert_memory",
+            "gen_ai.memory.store.id": store_id,
```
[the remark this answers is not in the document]
trask: Updated the PR description to accurately reflect the ADK scenario: it emits `upsert_memory` and `search_memory` (matching ADK's `add_session_to_memory` / `search_memory` API), and the AWS Bedrock AgentCore scenario added in this PR covers the separate `create_memory` / `update_memory` / `delete_memory` / store-lifecycle spans. `upsert_memory` is a registered operation in `model/gen-ai/registry.yaml`.

trask (approved):
Thanks @nagkumar91!

docs/gen-ai/non-normative/models.ipynb
```diff
@@ -470,6 +470,58 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "94727515ec414a9f",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "def omit_default(schema: dict[str, Any]) -> None:\n",
```
lmolkova: I don't think this is necessary, other models don't have it and they don't have any default value in the schema 

https://github.com/open-telemetry/semantic-conventions-genai/blob/5badfb2ab6b41558144a80614eaeaa067316bf58/docs/gen-ai/gen-ai-retrieval-documents.json#L12-L16

model/gen-ai/spans.yaml
```diff
@@ -295,6 +303,55 @@ spans:
...
+      **Span kind** SHOULD be `CLIENT` and MAY be set to `INTERNAL` on spans representing
+      calls to memory systems running in the same process. It's RECOMMENDED to use `CLIENT` kind
+      when the memory system being instrumented usually runs in a different process than its
+      client or when the memory call happens over instrumented protocol such as HTTP.
+
+      For `delete_memory`, lack of `gen_ai.memory.record.id` may indicate that
+      the operation intends to delete all memory records in the specified store.
+    stability: development
+    attributes:
+      - ref_group: attributes.gen_ai.memory.client
```
lmolkova: nit, readability improvement

```suggestion
      - ref: gen_ai.operation.name
        requirement_level: required
      - ref_group: attributes.gen_ai.address_and_port
      - ref_group: attributes.gen_ai.error
```

and remove `attributes.gen_ai.memory.client`

lmolkova (approved):

lmolkova (approved):

PR #142, opened 2026-05-11
gen-ai: add document modality to multimodal content parts

Mandark-droid:
## Description

Adds `document` as a value of the `Modality` enum on `BlobPart`, `FilePart`, and `UriPart` in the gen-ai input, output, and system-instructions message JSON schemas. PDFs, DOCX, and other office-document payloads currently fall through to the free-form `string` branch of the `modality` `anyOf`; this change gives them a first-class enum value alongside `image` / `video` / `audio`.

Concrete shape on the wire:

```jsonc
{
  "type": "blob",
  "modality": "document",
  "mime_type": "application/pdf",
  "content": "<base64 of raw PDF bytes>"
}
```

The Pydantic source model in `docs/gen-ai/non-normative/models.ipynb` is the source of truth; the three JSON schemas are regenerated from it.

Follow-up from the closed `open-telemetry/semantic-conventions#3673`. Closes (or significantly advances) #50. Sister PRs in this repo: [`byte_size` #143](https://github.com/open-telemetry/semantic-conventions-genai/pull/143) and [`stripped_reason` #144](https://github.com/open-telemetry/semantic-conventions-genai/pull/144). The three are independent and additive.

## Motivation

**User journey:** content auditability and cost-attribution for document-processing GenAI workloads.

A high-volume real workload that drove this: **BFSI KYC (Know Your Customer) extraction** — a customer uploads a passport or utility bill as a PDF, the GenAI app extracts identity fields, and the instrumentation needs to record what kind of content was sent. Today the only way to convey "this was a PDF document, not an image" is either (a) leak the mime type alone and ask consumers to classify, or (b) fall through to the free-form `string` branch of `modality`, fragmenting how producers describe documents.

Both options block:

- **Cost attribution by content type** — operators want per-tenant per-modality token-usage rollups; `document` is a distinct class with different per-token cost characteristics (PDFs are commonly OCR'd or embedded by the model, which has cost implications different from raw images).
- **Compliance dashboards** that need to answer "how many document-class payloads were captured this quarter?" — a stable, normative enum value makes this a single-attribute query rather than a mime-type classification expression in every consumer.

**Prior art:**
- Anthropic's `messages` API has a first-class `document` content-block type today.
- OpenAI Chat Completions consumes PDFs via a `file` content block with inline `file_data` (data URI) or `file_id`.
- AWS Bedrock Converse has a native `DocumentBlock` (`format` + `source.bytes` + `name`).
[...]

PR #143, opened 2026-05-11
gen-ai: add optional byte_size to multimodal content parts

Mandark-droid:
## Description

Adds an optional `byte_size` field on `BlobPart`, `FilePart`, and `UriPart` in the gen-ai input, output, and system-instructions message JSON schemas. `byte_size` captures the size, in bytes, of the captured or referenced media payload as observed by the instrumentation at the time of capture.

Concretely each part type carries its own focused description:

| Part | `byte_size` description |
|---|---|
| `BlobPart` | Size of the captured payload in bytes — the raw decoded byte length, not the length of the base64-encoded `content` string on the wire. |
| `FilePart` | Size of the referenced file in bytes, as observed by the instrumentation at capture time. |
| `UriPart` | Size of the URI-referenced payload in bytes, as observed by the instrumentation at capture time. |

The Pydantic source models in `docs/gen-ai/non-normative/models.ipynb` declare `byte_size: Optional[int] = Field(default=None, ge=0, ...)` so the regenerated JSON schemas carry `"minimum": 0` inside the `anyOf` integer branch — negative values are rejected by validators rather than silently accepted.

Follow-up from the closed PR open-telemetry/semantic-conventions#3673 (split into three small PRs per the new repo's "keep PRs small" guidance: this one, the [`document` modality PR #142](https://github.com/open-telemetry/semantic-conventions-genai/pull/142), and the [`stripped_reason` PR #144](https://github.com/open-telemetry/semantic-conventions-genai/pull/144)). The three are independent and additive.

## Motivation

**User journey:** cost-of-capture telemetry, storage planning, and capacity forecasting for content-capturing instrumentation.

A single logical multimodal request may be carried inline (`BlobPart` with base64 `content`), by URI (`UriPart` pointing at S3 / GCS / MinIO / similar), or by provider-side file id (`FilePart`). Without a uniform `byte_size`, an operator who wants to answer "how much media data are we ingesting per minute / per tenant / per modality?" has to:

- For `BlobPart`: base64-decode `content` and count bytes (and decide whether to do that per-span, which is expensive).
- For `UriPart`: dereference the URI to a backing store and HEAD it (involves an extra request and trust in the store).
- For `FilePart`: ask the provider's files API for size metadata (extra API call, rate-limited, provider-specific).

Putting an optional `byte_size` directly on the captured part lets the producing instrumentation record what it already knows (it just saw the bytes or just got back a `content-length` header / file metadata) so consumers don't have to re-derive it later.

**Prior art:**
- AWS X-Ray / Bedrock spans record request and response sizes as flat attributes.
- OpenAI's Files API exposes `bytes` on every file object — instrumentation that's about to upload via `Files.create()` already has this number.
- The reference implementation in [`genai-otel-instrument`](https://github.com/Mandark-droid/genai_otel_instrument) (v1.1.1) already populates an equivalent inside its dual-emission JSON when `OTEL_SEMCONV_STABILITY_OPT_IN=gen_ai` — this PR upstreams the convention.

## Prototype

This PR ships the schema change in `docs/gen-ai/`. The repo's reference scenarios under `reference/scenarios/<library>/` are all currently text-only — none emit `BlobPart`/`FilePart`/`UriPart` today — so there is no obvious existing scenario to add `byte_size` to without first introducing a multimodal scenario.

I flagged this as an open question in the original PR body (and on the closed-and-refiled open-telemetry/semantic-conventions#3673): would maintainers prefer (a) extending an existing `<library>` scenario (e.g. `openai`) with a multimodal call that emits `byte_size`, or (b) a dedicated `openai-multimodal` (or similar) scenario, given the SDK call shape changes? Happy to do either as a follow-up — leaving this checkbox unchecked rather than guessing the shape. Cross-referenced from open-telemetry/semantic-conventions-genai#142, where the document-modality counterpart of this question was raised and approved without a scenario.

End-to-end producer verification: the [`genai-otel-instrument`](https://github.com/Mandark-droid/genai_otel_instrument) library emits `byte_size` on URI parts captured by its OpenAI / Anthropic / Bedrock instrumentors, validated through an OTel collector → OpenSearch pipeline against a MinIO-backed media store; the actual size value observed at capture time round-trips into the `gen_ai.input.messages` span attribute as the JSON schema in this PR describes.

## Checklist

- [ ] Reference scenarios updated for affected libraries — *deferred pending maintainer guidance on shape; see Prototype section*

### DIFF
--- a/docs/gen-ai/non-normative/examples-llm-calls.md
+++ b/docs/gen-ai/non-normative/examples-llm-calls.md
@@ -239,7 +239,11 @@ See the [normative JSON schema](/model/gen-ai/gen-ai-input-messages.json) for mo
         "mime_type": "image/png",
         "uri": "https://raw.githubusercontent.com/open-telemetry/opentelemetry.io/refs/heads/main/static/img/logos/opentelemetry-horizontal-color.png"
       },
-      // A video with a vendor specific URI
+      // A video with a vendor specific URI. `byte_size` is omitted here because
+      // the size of a pre-existing remote URI is generally not observable to the
+      // instrumentation without downloading the content. Set `byte_size` on a
+      // `uri` part only when the instrumentation observed the bytes before they
+      // became a URI (e.g. an inline payload uploaded to an object store).
       {
         "type": "uri",
         "modality": "video",
@@ -257,11 +261,13 @@ See the [normative JSON schema](/model/gen-ai/gen-ai-input-messages.json) for mo
         "modality": "image",
         "file_id": "provider_fileid_123"
       },
-      // An inline image
+      // An inline image. `byte_size` is unambiguous here because the
+      // instrumentation observed the payload directly.
       {
         "type": "blob",
         "modality": "image",
         "mime_type": "image/png",
+        "byte_size": 37,
         "content": "aGVsbG8gd29ybGQgaW1hZ2luZSB0aGlzIGlzIGFuIGltYWdlCg=="
       },
       // Inline audio
--- a/docs/gen-ai/non-normative/models.py
+++ b/docs/gen-ai/non-normative/models.py
@@ -204,6 +204,15 @@ class BlobPart(BaseModel):
     modality: Union[Modality, str] = Field(
         description="The general modality of the data if it is known. Instrumentations SHOULD also set the mimeType field if the specific type is known."
     )
+    byte_size: Optional[int] = Field(
+        default=None,
+        ge=0,
+        description=(
+            "Size of the captured payload in bytes — the raw decoded byte length, not the length "
+            "of the base64-encoded `content` string on the wire. Useful for cost-of-capture "
+            "telemetry and storage planning."
+        ),
+    )
     content: bytes = Field(
         description="Raw bytes of the attached data. This field SHOULD be encoded as a base64 string when serialized to JSON."
     )
@@ -220,6 +229,14 @@ class FilePart(BaseModel):
     )
     modality: Union[Modality, str] = Field(
         description="The general modality of the data if it is known. Instrumentations SHOULD also set the mimeType field if the specific type is known."
+    )
+    byte_size: Optional[int] = Field(
+        default=None,
+        ge=0,
+        description=(
+            "Size of the referenced file in bytes, as observed by the instrumentation at capture "
+            "time. Useful for cost-of-capture telemetry and storage planning."
+        ),
     )
     file_id: str = Field(
         description="An identifier referencing a file that was pre-uploaded to the provider."
@@ -239,6 +256,14 @@ class UriPart(BaseModel):
     )
     modality: Union[Modality, str] = Field(
         description="The general modality of the data if it is known. Instrumentations SHOULD also set the mimeType field if the specific type is known."
+    )
+    byte_size: Optional[int] = Field(
+        default=None,
+        ge=0,
+        description=(
+            "Size of the URI-referenced payload in bytes, as observed by the instrumentation at "
+            "capture time. Useful for cost-of-capture telemetry and storage planning."
+        ),
     )
     uri: str = Field(
         description=(
--- a/model/gen-ai/gen-ai-input-messages.json
+++ b/model/gen-ai/gen-ai-input-messages.json
@@ -34,6 +34,20 @@
                     "description": "The general modality of the data if it is known. Instrumentations SHOULD also set the mimeType field if the specific type is known.",
                     "title": "Modality"
                 },
+                "byte_size": {
+                    "anyOf": [
+                        {
+                            "minimum": 0,
+                            "type": "integer"
+                        },
+                        {
+                            "type": "null"
+                        }
+                    ],
+                    "default": null,
+                    "description": "Size of the captured payload in bytes \u2014 the raw decoded byte length, not the length of the base64-encoded `content` string on the wire. Useful for cost-of-capture telemetry and storage planning.",
+                    "title": "Byte Size"
+                },
                 "content": {
                     "description": "Raw bytes of the attached data. This field SHOULD be encoded as a base64 string when serialized to JSON.",
                     "format": "binary",
@@ -204,6 +218,20 @@
                     ],
                     "description": "The general modality of the data if it is known. Instrumentations SHOULD also set the mimeType field if the specific type is known.",
                     "title": "Modality"
+                },
+                "byte_size": {
+                    "anyOf": [
+                        {
+                            "minimum": 0,
+                            "type": "integer"
+                        },
+                        {
+                            "type": "null"
+                        }
+                    ],
+                    "default": null,
+                    "description": "Size of the referenced file in bytes, as observed by the instrumentation at capture time. Useful for cost-of-capture telemetry and storage planning.",
+                    "title": "Byte Size"
                 },
                 "file_id": {
                     "description": "An identifier referencing a file that was pre-uploaded to the provider.",
@@ -519,6 +547,20 @@
                     ],
                     "description": "The general modality of the data if it is known. Instrumentations SHOULD also set the mimeType field if the specific type is known.",
                     "title": "Modality"
+                },
+                "byte_size": {
+                    "anyOf": [
+                        {
+                            "minimum": 0,
+                            "type": "integer"
+                        },
+                        {
+                            "type": "null"
+                        }
+                    ],
+                    "default": null,
+                    "description": "Size of the URI-referenced payload in bytes, as observed by the instrumentation at capture time. Useful for cost-of-capture telemetry and storage planning.",
+                    "title": "Byte Size"
                 },
                 "uri": {
                     "description": "A URI referencing attached data. It should not be a base64 data URL, which should use the `blob` part instead. The URI may use a scheme known to the provider api (e.g. `gs://bucket/object.png`), or be a publicly accessible location.",
--- a/changelog.d/143.enhancement.md
+++ b/changelog.d/143.enhancement.md
@@ -1,0 +1,1 @@
+Add optional `byte_size` (non-negative integer) to `BlobPart`, `FilePart`, and `UriPart` in the GenAI input and output message JSON schemas. Lets instrumentations record the size of a captured media payload for cost-of-capture telemetry and storage planning; omitted when the size is not cheaply observable (e.g. a pre-existing remote URI).
--- a/model/gen-ai/gen-ai-output-messages.json
+++ b/model/gen-ai/gen-ai-output-messages.json
@@ -34,6 +34,20 @@
                     "description": "The general modality of the data if it is known. Instrumentations SHOULD also set the mimeType field if the specific type is known.",
                     "title": "Modality"
                 },
+                "byte_size": {
+                    "anyOf": [
+                        {
+                            "minimum": 0,
+                            "type": "integer"
+                        },
+                        {
+                            "type": "null"
+                        }
+                    ],
+                    "default": null,
+                    "description": "Size of the captured payload in bytes \u2014 the raw decoded byte length, not the length of the base64-encoded `content` string on the wire. Useful for cost-of-capture telemetry and storage planning.",
+                    "title": "Byte Size"
+                },
                 "content": {
                     "description": "Raw bytes of the attached data. This field SHOULD be encoded as a base64 string when serialized to JSON.",
                     "format": "binary",
@@ -126,6 +140,20 @@
                     ],
                     "description": "The general modality of the data if it is known. Instrumentations SHOULD also set the mimeType field if the specific type is known.",
                     "title": "Modality"
+                },
+                "byte_size": {
+                    "anyOf": [
+                        {
+                            "minimum": 0,
+                            "type": "integer"
+                        },
+                        {
+                            "type": "null"
+                        }
+                    ],
+                    "default": null,
+                    "description": "Size of the referenced file in bytes, as observed by the instrumentation at capture time. Useful for cost-of-capture telemetry and storage planning.",
+                    "title": "Byte Size"
                 },
                 "file_id": {
                     "description": "An identifier referencing a file that was pre-uploaded to the provider.",
@@ -550,6 +578,20 @@
                     ],
                     "description": "The general modality of the data if it is known. Instrumentations SHOULD also set the mimeType field if the specific type is known.",
                     "title": "Modality"
+                },
+                "byte_size": {
+                    "anyOf": [
+                        {
+                            "minimum": 0,
+                            "type": "integer"
+                        },
+                        {
+                            "type": "null"
+                        }
+                    ],
+                    "default": null,
+                    "description": "Size of the URI-referenced payload in bytes, as observed by the instrumentation at capture time. Useful for cost-of-capture telemetry and storage planning.",
+                    "title": "Byte Size"
                 },
                 "uri": {
                     "description": "A URI referencing attached data. It should not be a base64 data URL, which should use the `blob` part instead. The URI may use a scheme known to the provider api (e.g. `gs://bucket/object.png`), or be a publicly accessible location.",


Mandark-droid:
Thanks @Copilot for the review — all four items addressed in [b6930e0](https://github.com/open-telemetry/semantic-conventions-genai/pull/143/commits/b6930e0):

- **`byte_size >= 0`** — added Pydantic `ge=0` on the `byte_size` field of `BlobPart`/`FilePart`/`UriPart` in `models.ipynb`. The regenerated JSON schemas now emit `"minimum": 0` inside the `anyOf` integer branch, so consumers get consistent validation rather than silently accepting negative values. This is exactly the "via a Pydantic constraint such as `ge=0`" path you suggested.
- **`byte_size` description for `BlobPart`** — clarified that for `BlobPart`, `byte_size` is the size of the raw payload (the decoded byte length), not the length of the base64-encoded `content` string on the wire. The same description now reads identically on `FilePart` and `UriPart` since the on-wire payload there is opaque (no base64 ambiguity), keeping the field semantics uniform across the three parts.
- **"vendor specific" → "vendor-specific"** — hyphenated the compound modifier in the example.
- **"file id" → "file ID"** — capitalized in the CHANGELOG entry.

`make check-policies` and `make generate-all` re-validated locally on the updated branch.

trask:
@Mandark-droid can you fill out the new PR template? thanks! https://raw.githubusercontent.com/open-telemetry/semantic-conventions-genai/refs/heads/main/.github/PULL_REQUEST_TEMPLATE.md

Mandark-droid:
Thanks @Copilot for the second pass — good catch. Addressed in [d725a44](https://github.com/open-telemetry/semantic-conventions-genai/pull/143/commits/d725a44):

In the first fixup I broadened the description to mention BlobPart-specific base64 semantics, but applied it uniformly across all three parts. That was wrong: FilePart and UriPart have no `content` field so the base64 caveat doesn't apply there. Now each part has its own focused description:

| Part | Description |
|---|---|
| `BlobPart` | Size of the captured payload in bytes — the raw decoded byte length, not the length of the base64-encoded `content` string on the wire. |
| `FilePart` | Size of the referenced file in bytes, as observed by the instrumentation at capture time. |
| `UriPart` | Size of the URI-referenced payload in bytes, as observed by the instrumentation at capture time. |

The notebook and the three generated JSON schemas are updated in lockstep; `make check-policies` and `make generate-all` re-validated locally.

Mandark-droid:
@trask done — PR body now filled out against [the new template](https://raw.githubusercontent.com/open-telemetry/semantic-conventions-genai/refs/heads/main/.github/PULL_REQUEST_TEMPLATE.md). Description / Motivation (with user journey + prior art) / Prototype / Checklist all in place.

One checklist item left intentionally unchecked: **Reference scenarios updated for affected libraries.** I flagged this in the Prototype section — the repo's existing `reference/scenarios/<library>/` scenarios are all text-only today, so there's no obvious existing scenario to extend with a `byte_size`-bearing `BlobPart`/`FilePart`/`UriPart`. Same situation on the sister PRs #142 and #144. Two paths and I'd like your call before I do the work:

1. **Extend an existing library scenario** (most likely `openai`) with a small multimodal call — would change the SDK invocation shape there (vision input / file id input), so the scenario stops being "minimal text-only" but stays under the existing library directory.
2. **Add a dedicated `openai-multimodal/` (or `multimodal/`) scenario** that lives alongside the others — keeps the existing scenarios untouched, but adds a whole new uv-locked scenario folder + `data.json` + README regen.

If you prefer (1) or (2), I'll do it as part of this PR. If you'd rather land the schema changes first and follow up with the multimodal scenario(s) as a separate PR — also happy to do that, especially since the same scenario will want to exercise `document` modality (#142), `byte_size` (#143), and `stripped_reason` (#144) together rather than in three separate scenario additions.

Also note PR #142 was already approved by @MikeGoldsmith on a deferred-scenario basis — if the same standard applies here #143 and #144 could land first and the scenario lands in a unified follow-up. Whatever you prefer.

PR #144's body has been similarly filled out against the template; @MikeGoldsmith, flagging since you reviewed #142.

trask:
let's add the scenario to #142, get that merged, then you can update that scenario in this PR

docs/gen-ai/non-normative/examples-llm-calls.md
```diff
@@ -237,11 +237,12 @@ See the [normative JSON schema](/docs/gen-ai/gen-ai-input-messages.json) for mor
         "mime_type": "image/png",
         "uri": "https://raw.githubusercontent.com/open-telemetry/opentelemetry.io/refs/heads/main/static/img/logos/opentelemetry-horizontal-color.png"
       },
-      // A video with a vendor specific URI
+      // A video with a vendor-specific URI and known size
       {
         "type": "uri",
         "modality": "video",
         "mime_type": "video/mp4",
+        "byte_size": 5242880,
```
Cirilla-zmh: For `uri` and `file` type responses, I suspect you may not actually be able to obtain the `byte_size` value, since it is generally only determined when the content is downloaded.

Mandark-droid: Good catch, and the schema does handle this case — `byte_size` is `Optional[int]` precisely so instrumentations omit it when the value isn't cheaply observable. But it's worth saying *why* we wanted this field in the first place, because it shapes which cases legitimately have a size and which don't.

**Motivation: trace-side metadata as a join key for offline multimodal evals.**

The field exists to support a pattern we ship in `genai-otel-instrument`: optional offload of multimodal payloads (images / audio / video / documents) to an object store — MinIO in our default deployment, configurable to any S3-compatible target — while spans carry only lightweight metadata (modality, MIME, byte size, source, SHA, store URI). The bytes themselves never go on the span. The store becomes the substrate for downstream evaluation: LLM-as-judge passes over generated images, voice-quality scoring on audio responses, OCR/grounding checks on document inputs, joint text+image evals on text-to-image flows (judge both the prompt *and* the produced artifact). As the industry moves toward voice and multimodal use cases this becomes the dominant eval shape, and trace ↔ artifact correlation is what makes it tractable.

`byte_size` plays two roles in that pattern:

1. **Standalone signal when the offload sink is off.** Even with no object store wired up, byte size on the span gives operators a meaningful low-cost dimension — cost attribution (multimodal billing scales with payload size), spotting oversized or empty inputs, anomaly detection. It's a single integer; negligible on the wire.
2. **Correlation / integrity key when the sink is on.** Pairs with `mime_type` + (optionally) `sha256` to let an evaluator join the lightweight trace to the heavyweight artifact in the store, and to detect truncation/corruption between span emission and storage.

**Which UriPart / FilePart cases actually have a known size — concretely, in our reference impl:**

There are three code paths that produce a `UriPart`-shape, and `byte_size` is set on two of them:

1. **Inline bytes → uploaded to object store → UriPart** (`media/offload.py:142-147`). We started with inline `data:` URIs or raw bytes from the SDK, computed `len(payload)`, uploaded, and emit a UriPart pointing at the upload. Size is exact because we held the bytes. This is the primary "interesting" case — it's also the case where `byte_size` is most load-bearing for evals, because it lets the evaluator verify the artifact in the store matches what the trace says was sent.
2. **Inline bytes too large to upload → stripped UriPart-shape** (`media/offload.py:114-120`). We set `media_byte_size = len(part.data)` *before* dropping the buffer, then emit a stripped part with `stripped_reason: "size_exceeded"`. Size is the entire reason this branch exists — without it, consumers can't tell *why* the part was stripped or estimate what was lost.
3. **Caller-supplied external URL → UriPart** (`media/offload.py:95-99`). User passed an already-remote URL (e.g. OpenAI `image_url` with `https://…`). We record `media_uri = part.external_url` and **do not set `media_byte_size`**. The canonical writer at `media/canonical.py:66-67` only emits `byte_size` when it's not `None`, so this UriPart goes out without it — exactly what you'd want.

The same logic applies to `FilePart`: when an app references a provider-side file (OpenAI Files API returns `bytes`, Anthropic Files returns `size_bytes`, Gemini `File.sizeBytes`), the SDK response carries the size, so it's known without a download. When it isn't carried, the field stays unset.

**On the example you flagged** (`examples-llm-calls.md:245`, "vendor-specific URI" video): you're right that as written it implies an instrumentation magicked the size out of a remote URL. In our impl's terms that's the external_url path, where `byte_size` is *not* set. I'll push an example fix that drops `byte_size` from that specific UriPart entry and adds a brief note explaining that `byte_size` on UriPart is only populated when the instrumentation observed the bytes before they became a URI — typically the offload-to-object-store flow this field is designed to support. Happy to instead replace with an inline-then-offloaded UriPart (where size is genuinely known) plus a sibling URI example without `byte_size` if you'd prefer the canonical "size on UriPart is legitimate" case demonstrated explicitly. LMK.

model/gen-ai/gen-ai-input-messages.json
```diff
@@ -159,6 +173,20 @@
                     "description": "The general modality of the data if it is known. Instrumentations SHOULD also set the mimeType field if the specific type is known.",
                     "title": "Modality"
                 },
+                "byte_size": {
```
trask, line 222: how would instrumentation capture this on `FilePart` and `UriPart`?

model/gen-ai/gen-ai-input-messages.json
```diff
@@ -34,6 +34,20 @@
...
```
trask, line 37: on BlobPart, it's a bit redundant with content since easy to calculate from the base64 length

I think would be ok to capture on BlobPart for consistency with others, but I'm not sure about the others (see other comment), which is why I mention

docs/gen-ai/gen-ai-input-messages.json
```diff
@@ -225,7 +253,8 @@
             "enum": [
                 "image",
                 "video",
-                "audio"
+                "audio",
+                "document"
```
trask, line 303: 👍

model/gen-ai/gen-ai-input-messages.json
trask: oh, I see, this would be useful with #144 which makes content optional

model/gen-ai/gen-ai-input-messages.json
Mandark-droid: The answer differs by part:

**FilePart** — the size comes back in the SDK response, no download needed. When an app references a provider-side file, the file-create / file-get call already returns the size: OpenAI Files `bytes`, Anthropic Files `size_bytes`, Gemini `File.size_bytes`. The instrumentation reads it off the same object it's already inspecting to build the part. When a provider doesn't surface it, the field stays unset (`Optional`).

**UriPart** — two sub-cases:
- *Instrumentation-produced URI* (the case the field is really for): the instrumentation held the bytes inline, computed `len(payload)`, uploaded them to an object store, and emitted a UriPart pointing at the upload. Size is exact because the bytes were in hand before they became a URI. This is the offload-to-object-store flow `byte_size` is designed for — it lets an offline evaluator join the lightweight span to the heavyweight artifact and detect truncation.
- *Caller-supplied pre-existing remote URL* (e.g. an `image_url` that's already `https://…`): the size generally isn't observable without downloading, so `byte_size` is **not** set. That's the misleading example you and @Cirilla-zmh flagged; it's corrected in [77d45f8](https://github.com/open-telemetry/semantic-conventions-genai/commit/77d45f8) — the remote-URI example no longer carries `byte_size`, with a note that it's only populated when the bytes were observed before becoming a URI, and the inline-blob example gains it where the value is unambiguous.

**BlobPart** — agreed it's derivable from the base64 length, so on its own it's redundant. The reason to still allow it is uniformity across the three part types, and — as you noted — it's the only surviving size signal once #144 lets `content` be omitted: `{"type":"blob","modality":"image","stripped_reason":"size_exceeded","byte_size":2000000}` tells a consumer how much was dropped. So `byte_size` and #144 are complementary by design.

If you'd rather land this narrowly, I can take the same approach as #144 — scope this PR to `byte_size` on input-messages `BlobPart` only and replicate to FilePart/UriPart plus the other two schemas in a follow-up.

Mandark-droid:
Rebased onto `main` now that #142 has landed — the merged document-modality commits drop out, so this is down to a single commit (`8d5a161`) adding optional `byte_size` to `BlobPart`/`FilePart`/`UriPart` across the input/output/system-instructions schemas, plus the reference model and examples. The earlier example clarification (remote-URI size is not observable without a download; only set `byte_size` when the bytes were seen before becoming a URI) is folded into this commit.

Reference-scenario emission is deferred to the follow-up that will carry document-modality, `byte_size`, and `stripped_reason` (#144) together in one multimodal scenario.

model/gen-ai/gen-ai-input-messages.json
trask: can you add reference scenarios for FilePart and UriPart?

trask:

Create `changelog.d/143.enhancement.md` containing the change log entry, e.g.:

```md
Add optional `byte_size` (non-negative integer) to `BlobPart`, `FilePart`, and `UriPart` in the GenAI input/output/system-instructions message JSON schemas. Lets instrumentations record the size of a captured media payload for cost-of-capture telemetry and storage planning; omitted when the size is not cheaply observable (e.g. a pre-existing remote URI).
```

Thanks!

Mandark-droid:
@trask done — the changelog entry now lives in `changelog.d/143.enhancement.md` and the direct `CHANGELOG.md` edit is removed (commit be20d01).

While rebasing onto current `main` I also ported the PR to the #226 notebook-to-Python conversion: `byte_size` is now declared on `BlobPart`/`FilePart`/`UriPart` in `docs/gen-ai/non-normative/models.py`, and the three committed message schemas are regenerated via `make generate-json-schemas`. The locked generation environment reproduces `main`'s committed output byte-for-byte for untouched schemas, so the JSON diff is unchanged from what was previously reviewed — still just the three `byte_size` property additions.

model/gen-ai/gen-ai-input-messages.json
Mandark-droid: Picking this back up — sorry for the delay.

Both your comments are really one question: can generic instrumentation observe this at runtime? Working through it honestly, the answer differs per part type, and it is narrower than this PR currently implies.

**`file` — yes, generically.** The size arrives in the same response as the identifier: OpenAI `bytes`, Anthropic `size_bytes`, Gemini `sizeBytes`. No extra call and no download. This is the strongest case for the field.

**`uri` — only when the instrumentation minted the URI.** Two sub-cases:

- The caller supplied a remote URL. Not observable, and the instrumentation must not download to find out, so `byte_size` is omitted.
- The instrumentation received the bytes inline, offloaded them to an object store, and emitted a reference in their place. The size is exact, because it measured what it uploaded.

The second case is real, but it needs an instrumentation that offloads — it is not something every instrumentation does. So I don't think `byte_size` on `uri` can be stated as generally capturable, and the PR shouldn't imply it is. What I think *is* defensible is the narrower rule: set it only when the instrumentation observed the bytes before they became a URI, and omit it otherwise.

If that scoping works for you, I'll add a reference scenario under `reference/scenarios/openai/` that covers both directions in one run — a Files API part where the size comes back with the id, and an `image_url` part where it is correctly absent — and narrow the normative wording to match.

Does that scope work, or would you rather `byte_size` be restricted to `file` (and possibly `blob`) only, leaving `uri` out entirely? Happy to go either way; I'd just rather agree the shape before writing the scenario.

model/gen-ai/gen-ai-input-messages.json
Mandark-droid: Agreed — on `blob` it is derivable from the base64 length, so it carries no information the consumer doesn't already have.

I kept it for consistency, so a consumer can read one field across all three part types instead of special-casing `blob`. That is a convenience argument, not a capturability one, and I don't feel strongly about it: happy to drop `byte_size` from `BlobPart` if you'd rather the field mean strictly "a size the instrumentation could not otherwise express".

Flagging one interaction before we decide — with #144, `content` on `BlobPart` becomes optional, and a part whose bytes were observed and then deliberately not captured has no base64 length left to derive from. If we keep the field on `blob`, that is the case it earns its place in. If we drop it, that case would need to express the size some other way. Either is fine, but the two PRs should agree.

See the sibling thread for the `file` / `uri` capturability answer.

docs/gen-ai/non-normative/models.py
```diff
@@ -140,6 +140,7 @@ class BlobPart(BaseModel):
     modality: Union[Modality, str] = Field(
         description="The general modality of the data if it is known. Instrumentations SHOULD also set the mimeType field if the specific type is known."
     )
+    byte_size: Optional[int] = Field(default=None, ge=0, description="Size of the captured payload in bytes — the raw decoded byte length, not the length of the base64-encoded `content` string on the wire. Useful for cost-of-capture telemetry and storage planning.")
```
trask: Nothing in this PR demonstrates `byte_size` in a reference scenario. Since #142 merged, `run_chat_with_document_input()` in `reference/scenarios/anthropic/scenario.py` and `run_converse_with_document_input_reference()` in `reference/scenarios/aws-bedrock/scenario.py` both emit a `BlobPart` whose `len(pdf_bytes)` is available at the SDK call boundary, so `byte_size` can be set there now. The PR body still says every reference scenario is text-only.

PR #144, opened 2026-05-11
gen-ai: make input-messages BlobPart content optional and add stripped_reason

Mandark-droid:
## Description

Makes `BlobPart.content` optional and adds an optional free-form `stripped_reason` field on the `gen_ai.input.messages` `BlobPart`. This enables fail-closed observability for inline blobs: an instrumentation that detected a blob but intentionally did not capture its payload bytes can record that fact while preserving the original part `type` and `modality`.

```json
{
  "type": "blob",
  "modality": "video",
  "mime_type": "video/mp4",
  "stripped_reason": "size_exceeded"
}
```

A top-level `anyOf` on `BlobPart` enforces that an instance must carry **either** a non-null `content` **or** a non-null `stripped_reason` — both absent (or both `null`) is rejected, so structurally-empty parts cannot validate.

### Scope

This PR is intentionally narrowed to `BlobPart` on `gen_ai.input.messages` only, per @trask's review request. If the shape is accepted, follow-up PRs replicate it to:

1. `FilePart` (`file_id` optional + `stripped_reason`) and `UriPart` (`uri` optional + `stripped_reason`) on input messages — the `stripped_reason` wording there refers to the file reference / external URI rather than payload bytes; see the discussion on the capture feasibility of those parts in #143.
2. The `gen_ai.output.messages` and `gen_ai.system_instructions` schemas, which reuse the same part definitions.

Follow-up from the closed PR open-telemetry/semantic-conventions#3673. Sister PRs: [`document` modality #142](https://github.com/open-telemetry/semantic-conventions-genai/pull/142) (merged) and [`byte_size` #143](https://github.com/open-telemetry/semantic-conventions-genai/pull/143). All independent and additive.

## Motivation

**User journey:** compliance auditing and capture-pipeline diagnostics for content-capturing instrumentations.

Today the gen-ai message schemas have no way to record a part the instrumentation observed but intentionally did not capture. The instrumentation can either:

1. Emit the full part with its payload (`content`). This works only when the part is captureable — under the configured size cap, permitted by the configured modality allow-list, approved by the configured redactor (if any), and the configured offload-to-store path succeeded.
2. Emit nothing at all when any of the above fails.

In case (2) the downstream consumer sees `parts: [...]` with no media — indistinguishable from a turn that simply had none. Two consumer use cases break:

- **Compliance audits** that need to prove media was *not* exfiltrated to a logging pipeline (because the redactor or size cap held). The auditor needs telemetry that says "the instrumentation acknowledged a video here and intentionally dropped its bytes," not silence.
- **Capture-pipeline regression detection.** If a deploy starts silently dropping media (e.g. an upload-to-MinIO bucket policy broke), there is no signal to alert on — the spans look fine, the bytes just aren't there. With `stripped_reason: "upload_error"` an SRE can dashboard-count strippeds per modality / per reason and page on a step-change.

**Prior art:**
- AWS bedrock-runtime's `InvokeModel` request logs annotate "guardrail intervened" + reason rather than dropping content silently.
- HAR (HTTP Archive) format has a `_comment` + `content` shape where an inline body can be a placeholder noting "body too large; not captured" while preserving `mimeType` and `size`.
- The reference implementation in [`genai-otel-instrument`](https://github.com/Mandark-droid/genai_otel_instrument) emits this stripped-blob marker on size-cap-exceeded media.

## Prototype

This PR ships the schema change in `docs/gen-ai/gen-ai-input-messages.json` and the matching reference model in `docs/gen-ai/non-normative/models.ipynb`. As with #143, the existing `reference/scenarios/<library>/` scenarios are text-only — there is no existing scenario into which to emit a stripped blob as-is, and a new multimodal scenario is its own chunk of work I'd prefer to scope under a follow-up so it can carry #143's `byte_size` and this PR's `stripped_reason` together.

**Functional schema check:** jsonschema validation against the schema in this PR confirms the `anyOf` rejects structurally-empty parts —

| Instance | Validates? | Why |
|---|---|---|
| `{"type":"blob","modality":"image","content":null}` (no stripped_reason) | ❌ | `None` not valid under `{"type":"null"}` — both anyOf branches fail |
| `{"type":"blob","modality":"image"}` (neither) | ❌ | both branches fail |
| `{"type":"blob","modality":"image","stripped_reason":"size_exceeded"}` | ✅ | second branch matches |
| `{"type":"blob","modality":"image","content":"aGVsbG8="}` | ✅ | first branch matches |

**End-to-end producer verification:** the `genai-otel-instrument` library emits this stripped-blob shape on its OpenAI / Anthropic / Bedrock instrumentors when a configured size cap is exceeded, validated against MinIO + OTel collector + OpenSearch. The library already ships the shape proposed here (the original separate `StrippedPart` type was dropped in v1.3.0 in favour of extending the existing part with a free-form `stripped_reason` string), so the producer side is aligned with this PR rather than pending on it.

## Checklist

- [ ] Reference scenarios updated for affected libraries — *deferred pending maintainer guidance; see Prototype section*


Mandark-droid:
Thanks @Copilot for the review — both items addressed in [a9107c4](https://github.com/open-telemetry/semantic-conventions-genai/pull/144/commits/a9107c4):

- **anyOf constraint — `{content|file_id|uri}` OR `stripped_reason` required** — good catch. With the payload fields removed from `required` it was indeed possible to construct a structurally empty part, which would defeat the whole "no media" vs "media stripped" distinction this PR introduces. Fixed by declaring `json_schema_extra` with the appropriate `anyOf` on the `Config` of each of `BlobPart`/`FilePart`/`UriPart` so the notebook stays the source of truth, and applying the matching shape to the three JSON schemas. Now a part must carry at least one of the payload-bearing field or `stripped_reason` (both is also fine, e.g. for a redactor that captured a hash but not the bytes).
- **```json → ```jsonc fence for the multimodal-inputs example** — this one was actually a false-positive against the head commit: the multimodal-inputs and multimodal-output blocks under `### Multimodal inputs example` / `### Multimodal output example` are already fenced as ` ```jsonc ` in the file (lines 224 and 289). The new stripped-part example this PR adds (around line 275 of the diff) lives inside that already-jsonc block, so the `//` comments are valid in the fence. No change required.

`make check-policies` re-validated locally on the updated branch.

Mandark-droid:
Thanks @Copilot for the second pass — good catch. The first-pass `anyOf` only checked key *presence*; since the payload fields are nullable, `{..., "content": null}` would still satisfy `{required: ["content"]}` and validate even without `stripped_reason`. Exactly the structurally-empty case the constraint was supposed to forbid.

Fixed in [b00a96f](https://github.com/open-telemetry/semantic-conventions-genai/pull/144/commits/b00a96f). Each branch now also constrains the corresponding property to non-null:

```json
"anyOf": [
    {
        "properties": {"content": {"not": {"type": "null"}}},
        "required": ["content"]
    },
    {
        "properties": {"stripped_reason": {"not": {"type": "null"}}},
        "required": ["stripped_reason"]
    }
]
```

So an instance must now carry either a non-null payload (`content` / `file_id` / `uri`) or a non-null `stripped_reason` (or both).

Functional check with `jsonschema` against the regenerated schema:

| Instance | Result |
|---|---|
| `{"type": "blob", "modality": "image", "content": null}` (no stripped_reason) | ❌ rejected — `None should not be valid under {'type': 'null'}` |
| `{"type": "blob", "modality": "image", "stripped_reason": "size_exceeded"}` | ✅ accepted |
| `{"type": "blob", "modality": "image", "content": "aGVsbG8="}` | ✅ accepted |

`Config.json_schema_extra` on each of `BlobPart`/`FilePart`/`UriPart` in `models.ipynb` updated to match, so the notebook stays the source of truth. `make check-policies` re-validated.

Mandark-droid:
Thanks @Copilot for the third pass. Addressed in [107f1be](https://github.com/open-telemetry/semantic-conventions-genai/pull/144/commits/107f1be):

**1 & 2. `stripped_reason` description was BlobPart-centric.** Same issue as PR #143 round-2 (broadening wording leaked BlobPart-specific bytes terminology onto FilePart / UriPart). Now per-part:

| Part | `stripped_reason` description |
|---|---|
| `BlobPart` | "did not capture its payload bytes" (unchanged — correct here) |
| `FilePart` | "did not capture the file identifier" (the payload is a provider file id, not bytes) |
| `UriPart` | "did not capture the URI (or did not resolve it to retrieve the underlying bytes)" |

Each description also tailors the example reason list — e.g. `size_exceeded` doesn't really apply to a `file_id` capture, so it's been dropped from FilePart's example list.

**3. Notebook code comment vs Pydantic runtime.** Fair — the old comment said the model "must carry" one or the other, but `Optional[...]` with `default=None` means a Python instance with both `content=None` and `stripped_reason=None` constructs fine at the Pydantic layer; the constraint only kicks in when the schema is exported and a JSON schema validator runs on a serialized instance. Rather than add a `@model_validator` (the notebook is intended as documentation rather than a production validator), the comment is now explicit:

```python
class Config:
    # JSON-schema-level constraint: serialized instances must carry either a non-null
    # `content` or a non-null `stripped_reason` (or both). The Pydantic *runtime* model
    # still allows both fields to be None; the constraint is exported via
    # `json_schema_extra` and enforced by JSON schema validators on the consumer side.
    json_schema_extra = {
        "anyOf": [
            {"properties": {"content": {"not": {"type": "null"}}}, "required": ["content"]},
            {"properties": {"stripped_reason": {"not": {"type": "null"}}}, "required": ["stripped_reason"]},
        ]
    }
```

Same clarification applied to FilePart and UriPart's Config blocks. Happy to add a `@model_validator` instead if maintainers prefer the runtime enforcement — let me know.

PR body has also been filled out against the new PR template (per @trask's request on #143). `make check-policies` re-validated locally.

docs/gen-ai/gen-ai-input-messages.json
```diff
@@ -225,7 +317,8 @@
             "enum": [
                 "image",
                 "video",
-                "audio"
+                "audio",
+                "document"
```
trask, line 321: would you mind splitting this piece out into a separate PR?

and then reducing this PR to just cover input messages BlobPart?

You can mention in the PR description that if accepted, a follow-up PR would add replicate the change across XYZ

I think that would make it easier to get reviewers engaged on this. Thanks!

Mandark-droid: Done. Force-pushed a rebase onto `main` (the merged #142 commits drop out) and reduced this PR to just `BlobPart` on `gen_ai.input.messages`: `content` becomes optional, `stripped_reason` is added, and a top-level `anyOf` requires one or the other. The diff is now 4 files (schema + the reference model in `models.ipynb` + one example + CHANGELOG).

`FilePart`/`UriPart` and the output-messages / system-instructions schemas are pulled out — I've written the follow-up plan into the PR description so the sequencing is on record. The `FilePart`/`UriPart` capture-feasibility question is the same one on #143; I answered it there.

lmolkova (commented):
I like where it's going, but Blob part is just one of the parts that has large content problem.

I'd prefer if we

- documented a configuration option that limits size of any content in any part 
- if size is exceeded we wouldn't capture it; or, for text, we would rather trim to the size
- we could  let users know that content was there by including its size on each affected part (especially blob) as you're proposing in the https://github.com/open-telemetry/semantic-conventions-genai/pull/143, but also on the text part
- we could potentially mimic OTLP 'dropped attributes' property and record something like that on a generic part (e.g. content_dropped: true) if that's necessary (presence of non-0 size would be an indication on its own)

Having a reason seems interesting, but it's not a common approach in OTel - if something is dropped, we don't record why, it could introduce interesting problems on its own.

docs/gen-ai/non-normative/models.ipynb
```diff
@@ -156,7 +156,20 @@
     "    modality: Union[Modality, str] = Field(\n",
```
lmolkova: this file was converted to python in #226, sorry for the inconvenience!

trask:

Create `changelog.d/144.enhancement.md` containing the change log entry, e.g.:

```md
Make `BlobPart.content` optional and add a free-form `stripped_reason` marker on the `gen_ai.input.messages` `BlobPart`, so an instrumentation can record that it observed an inline blob but intentionally did not capture the payload bytes (size cap exceeded, modality not allowed, redactor failure, etc.) while preserving the original part type and modality. A schema-level `anyOf` requires either a non-null `content` or a non-null `stripped_reason`. Scoped to input-messages `BlobPart`; if accepted, follow-up PRs replicate the change to `FilePart`/`UriPart` and to the output-messages and system-instructions schemas.
```

Thanks!

Mandark-droid:
@trask done — the changelog entry now lives in `changelog.d/144.enhancement.md` and the direct `CHANGELOG.md` edit is removed (commit 6244073).

The rebase onto current `main` also ports the PR to the #226 notebook-to-Python conversion. Since `docs/gen-ai/non-normative/models.py` now generates all message schemas from one shared part union, the input-only scoping is expressed via input-messages variants of `BlobPart`/`ChatMessage` (a small namespace class in `models.py` that reuses the shared class names, so the generated `$defs` keys and the wire format stay identical across schemas). `gen-ai-input-messages.json` is regenerated via `make generate-json-schemas`; the resulting JSON diff is the same shape that was previously reviewed — top-level `anyOf`, nullable `content`, and the `stripped_reason` property, with no changes to the output-messages or system-instructions schemas.

If maintainers would rather not carry the input-only variant classes in `models.py`, the alternative is to apply the change to the shared `BlobPart` (which widens scope back to all three schemas) — happy to go either way.

Mandark-droid:
Thanks @lmolkova — I think we agree on most of this, and the pieces you list map onto the current PR stack:

- **A documented capture-size limit option** — agreed this is needed and complementary. It is instrumentation configuration rather than schema shape, so I see it as a docs follow-up that this PR's schema change enables (the marker is what an instrumentation sets when that limit fires).
- **Size on each affected part** — that is exactly #143 (`byte_size` on `BlobPart`/`FilePart`/`UriPart`). Extending it to `TextPart` is a one-line addition there; happy to fold it in if you want it.
- **Trim-to-size for text** — agreed, trimming works for text because a prefix is still useful. It does not transfer to binary parts: a truncated blob is not decodable, so for `BlobPart` the real choice is captured vs not captured. That is why this PR makes `content` omittable rather than truncatable.
- **`content_dropped: true` mimicking OTLP dropped attributes vs a reason string** — I am open to converging on the boolean form if that is the more OTel-consistent shape; the core of this PR is "content optional + explicit stripped marker", and the marker's type is a one-line change. Two things made me start with a free-form string: (a) it is deliberately non-enumerated, so there is no reason taxonomy for the spec to maintain (the failure-mode concern you mention with recording "why"); and (b) in practice operators do need to distinguish "size cap fired" from "redactor errored" without correlating collector logs — the string costs nothing when unused.

On "presence of non-0 size would be an indication on its own": `byte_size` alone cannot carry that signal, because it is optional and not always cheaply observable (e.g. a pre-existing remote URI), and absence of `content` alone is ambiguous against providers that legitimately reference blobs without inline bytes. An explicit marker keeps "instrumentation stripped this" distinguishable from "this was never inline".

If the WG preference lands on size-limit-config + `byte_size` everywhere + a boolean marker, I can reshape this PR to that in one pass — the schema mechanics (optional `content` + `anyOf` constraint) stay the same either way.

docs/gen-ai/non-normative/models.ipynb
Mandark-droid: Ported in the rebase, thanks — the BlobPart change now lives in `docs/gen-ai/non-normative/models.py` (input-messages variant classes) and `gen-ai-input-messages.json` is regenerated via `make generate-json-schemas`.

PR #148, opened 2026-05-11
[chore] add basic release workflow

lmolkova:
Initial flow:

1. we prepare release PR
   - bump version in https://github.com/open-telemetry/semantic-conventions-genai/blob/191909afdf2b2e6190da62e5090e3f7a70227790/model/manifest.yaml#L7
   - update changelog
   - create **draft** release with matching tag and notes
2. release prep PR is merged
3. we trigger Release workflow to publish release with artifacts  (manifest and resolved schema)

We should automate it further, this is automation for step 3 only

Dry run: 

https://github.com/lmolkova/semantic-conventions-genai/releases/tag/v1.42.0-dev
https://github.com/lmolkova/semantic-conventions-genai/actions/runs/25702077087/job/75464192532


singankit (approved):

JWinermaSplunk (approved):
Looks good! I might note if we wanted to guard from releasing from a non-main branch or releasing an existing tag, but those are just some guards that came to mind and could be implemented later.

.github/workflows/release-dev.yml
```diff
@@ -0,0 +1,76 @@
...
+          echo "tag=v${version}" >> "$GITHUB_OUTPUT"
+          echo "Tag: v${version}"
+
+      - name: Find draft release matching schema_url
+        id: draft
+        env:
+          GH_TOKEN: ${{ github.token }}
+          TAG: ${{ steps.tag.outputs.tag }}
+        run: |
+          info=$(gh release view "$TAG" --json databaseId,isDraft 2>/dev/null || true)
```
trask: `databaseId` isn't needed if we publish via `gh release edit` instead of `gh api PATCH` (see suggestion below).

```suggestion
          info=$(gh release view "$TAG" --json isDraft 2>/dev/null || true)
```

.github/workflows/release-dev.yml
```diff
@@ -0,0 +1,76 @@
...
+            fi
+            exit 1
+          fi
+          is_draft=$(echo "$info" | jq -r .isDraft)
+          if [ "$is_draft" != "true" ]; then
+            echo "Release '${TAG}' is already published; refusing to re-release."
+            exit 1
+          fi
+          echo "release_id=$(echo "$info" | jq -r .databaseId)" >> "$GITHUB_OUTPUT"
+          echo "Draft release '${TAG}' matches model/manifest.yaml schema_url."
```
trask, line 51: Drop the `release_id` step output — no longer needed if we key the publish step off `$TAG`.

```suggestion
          fi
          echo "Draft release '${TAG}' matches model/manifest.yaml schema_url."
```

.github/workflows/release-dev.yml
```diff
@@ -0,0 +1,76 @@
...
+      - name: Publish release
+        env:
+          GH_TOKEN: ${{ github.token }}
+          RELEASE_ID: ${{ steps.draft.outputs.release_id }}
+        run: |
+          gh api \
+            --method PATCH \
+            "/repos/${{ github.repository }}/releases/${RELEASE_ID}" \
+            -F draft=false \
+            -F target_commitish="${{ github.sha }}"
```
trask: Use `gh release edit` instead of the lower-level REST PATCH — it's the idiomatic gh CLI for this and lets us key off `$TAG` directly, avoiding the `release_id` plumbing.

```suggestion
          TAG: ${{ steps.tag.outputs.tag }}
        run: |
          gh release edit "$TAG" \
            --draft=false \
            --target "${{ github.sha }}"
```

trask (approved):
Small simplification:

MikeGoldsmith (approved):

lmolkova:
tested recent changes in https://github.com/lmolkova/semantic-conventions-genai/releases/tag/v1.42.0-dev

PR #162, opened 2026-05-14
semconv for compaction

### LINKED ISSUE #154: Conversation Compaction Signal for GenAI Spans.
### Problem

Long-running chat and agentic workloads routinely exceed model context windows. To keep going, compaction happens at two distinct layers today:                                                         
                                                                                                                                                                                                          
 - LLM provider APIs and managed services — Anthropic's memory/context-management tooling, OpenAI's Responses API with server-side conversation state, Gemini's context caching, and Bedrock-hosted  agents all summarize, truncate, or otherwise condense conversation history on the server before the model sees it. The client sends the full history (or a conversation ID) and the provider decides what actually reaches the model.                                                                                                                                                                        
 - Agent frameworks and orchestrators — LangGraph, AutoGen, CrewAI, LangChain, LlamaIndex, and IDE/coding assistants (Claude Code, Cursor, Copilot Chat) compact between steps in a multi-turn run: summarizing older turns, applying sliding windows, or hierarchical memory before issuing the next model call.                                                                                           
   
In both cases, the model effectively runs on a derived view of a longer original conversation — and today OpenTelemetry GenAI spans have no way to express that. The signal is invisible to consumers,  whether the compaction happened server-side inside a provider or client-side inside an orchestrator.                                                                                                  
                                                                                                                                                                                                          
This creates several blind spots:                                                                                                                                                                     

- Quality regressions are invisible. When answer quality degrades deep into a long session, operators can't distinguish "model got worse" from "context was dropped during compaction" — and can't tell 
  which layer dropped it.
- Token accounting misleads. gen_ai.usage.input_tokens on a post-compaction turn looks artificially small relative to the true conversation length, distorting cost and capacity analysis. Provider-side
   compaction is especially opaque: billed tokens may not reflect the conversation the user actually sent.                                                                                                
- Cross-turn correlation breaks. Two spans sharing a gen_ai.conversation.id may have wildly different effective input shapes; consumers can't tell which turns are "fresh" vs. "post-compaction" without
   inspecting payloads.                                                                                                                                                                                   
- Evaluation pipelines have no filter. Eval and replay tooling that wants to study only un-compacted turns (or only compacted ones) has no attribute to query on.                                     
                                                                                                                                                                                                          
### Solution                                                                                                                                                                                              

Add one boolean under the existing gen_ai.conversation.* namespace:                                                                                                                                     

- gen_ai.conversation.compacted (boolean) — true when the span's effective input represents a compacted view of a longer conversation, regardless of which layer performed the compaction(I am flexible to constrain the scope to server side if folks have an opinion).              
                                                                                                                                                                                                        
### How this helps                                                                                                                                                                                          
                                                                                                                                                                                                        
 - Operators can filter traces by compacted=true at either layer — agent-turn level (orchestrator compaction) or inference-call level (provider compaction) — to isolate context-loss issues from model  
  or prompt issues.
- Cost & capacity teams get a corrective lens on token usage, exposing turns where what was sent differs from what was processed — which matters especially for providers that compact server-side.     
- Eval & replay tooling gains a first-class predicate to slice traces (e.g., "show me only turns where compaction kicked in") without parsing message payloads.                                         
- Provider and framework authors — Anthropic, OpenAI, AWS Bedrock on the API side; LangGraph, and otheragent SDKs, IDE assistants on the orchestration side — get a single standard signal that works uniformly across both layers, improving interoperability. I am not sure if google gemini provides compaction or not but google's ADK provides compaction in their sdk.                                                                                                            
- Clean layering. Each span reports compaction at the layer that actually performed it. A trace where both the orchestrator and the provider compact will surface the flag on both, telling a complete story.                                                                                                                                                                                                  
- Forward-compatible. Sits cleanly next to gen_ai.conversation.id, follows established boolean patterns (gen_ai.request.stream), and leaves room for richer compaction telemetry without breaking changes.                                                                                                                                                                                                
                                                                                                                                                                                                        
Net effect: one bit of metadata, emitted at whichever layer actually performed compaction, closes a meaningful observability gap for any team running long-horizon GenAI workloads — whether the compaction lives in their agent framework, their LLM provider, or both. 

### References

1. https://platform.claude.com/docs/en/build-with-claude/compaction
2. https://developers.openai.com/api/docs/guides/compaction
3. https://docs.aws.amazon.com/bedrock/latest/userguide/claude-messages-compaction.html
4. https://learn.microsoft.com/en-us/agent-framework/agents/conversations/compaction?pivots=programming-language-csharp
5. https://www.langchain.com/blog/autonomous-context-compression.
6. https://adk.dev/context/compaction/

P.S: LLM help was used in describing the problem.

eternalcuriouslearner:
## Description

Adds `gen_ai.conversation.compacted`, a boolean GenAI conversation attribute that indicates whether the effective conversation context used for an operation is a compacted view of a prior conversation.

The attribute is added under the existing `gen_ai.conversation.*` namespace and is included as `Recommended when available` on:

- `gen_ai.inference.client`

This PR intentionally scopes the span-level attribute to inference spans. Earlier drafts considered adding the attribute to `invoke_agent` spans for orchestrator or agent-framework compaction, but that layer is left out of this PR. If we want to model framework/orchestrator compaction separately, it should be handled in a follow-up with span-specific guidance.

This PR also adds `CompactionPart` support to the GenAI input/output message schemas. When message payload capture is enabled, provider compaction items/blocks can be represented in:

- `gen_ai.output.messages[*].parts[*].type = "compaction"` when the provider emits a compaction item/block
- `gen_ai.input.messages[*].parts[*].type = "compaction"` when a prior compaction item/block is passed back on a later request

If `gen_ai.conversation.compacted` is unset, consumers should treat the compaction state as unknown.

This resolves #154.

## Motivation

Long-running chat and agentic workloads routinely exceed model context windows. To keep going, systems often compact conversation history by summarizing prior turns, pruning older messages, applying a sliding window, using hierarchical memory, or relying on provider-managed context reduction.

Today, OpenTelemetry GenAI spans can correlate turns with `gen_ai.conversation.id`, but they do not indicate whether the model operated on the full conversation history or on a compacted representation. That leaves a material observability gap for long-horizon GenAI workloads.

**User journey:** incident triage, quality evaluation, and cost analysis for long-running conversations and agent runs.

When answer quality degrades late in a conversation, an operator needs to distinguish model, prompt, retrieval, tool, and context-loss issues. If a span has `gen_ai.conversation.compacted=true`, the operator can immediately filter or group traces where context compaction may have affected the result, without inspecting sensitive message payloads.

The same signal helps token and capacity analysis. A post-compaction turn may have a much smaller `gen_ai.usage.input_tokens` value than the original conversation length would suggest. Marking compacted turns gives cost and capacity teams a first-class way to interpret token usage in long-running sessions.

Evaluation and replay tooling also benefits from a queryable predicate: evaluate only uncompacted turns, focus on compacted turns, or compare quality before and after compaction starts.

**Prior art:** provider APIs may expose compaction as part of their input/output model. OpenAI Responses API server-side compaction can emit a `ResponseCompactionItem` with `type: "compaction"` ([OpenAI compaction docs](https://developers.openai.com/api/docs/guides/compaction)). Anthropic Messages API server-side compaction can expose a compaction content block, `stop_reason: "compaction"`, or a usage iteration with `type: "compaction"` ([Anthropic compaction docs](https://platform.claude.com/docs/en/build-with-claude/compaction#pausing-after-compaction)).

`gen_ai.conversation.compacted` gives providers and SDKs one low-cardinality, non-sensitive signal for reporting that behavior consistently. `CompactionPart` complements it with payload-level fidelity for deployments that opt into recording message payloads. The boolean helps with filtering and dashboards when message payloads are disabled, redacted, truncated, sampled, or difficult to query. The message part helps debug whether a compaction item/block was emitted, where it appeared in the ordered conversation, and whether it was preserved on later turns.

Instrumentation should set `gen_ai.conversation.compacted=true` only when it can reliably determine that context compaction was applied. For example, instrumentation can derive the value from a provider-defined compaction output item/block, stop reason, usage iteration, or a compaction item/block present in the input conversation. Shared instrumentation that only sees an ordinary chat request and has no such signal should omit the attribute.

## Prototype

Reference scenarios demonstrate that the attribute can be produced from existing library-facing instrumentation surfaces and consumed in the reference support reports.

Updated scenarios:

- `reference/scenarios/openai/scenario.py`
  - emits `gen_ai.conversation.compacted=true` on inference spans
  - emits it on `gen_ai.client.inference.operation.details`
  - includes a Responses API example where the value is derived from a returned `ResponseCompactionItem` with `type: "compaction"`
  - records the returned compaction item as a `CompactionPart` in `gen_ai.output.messages`
- `reference/scenarios/anthropic/scenario.py`
  - emits `gen_ai.conversation.compacted=true` on inference spans
  - emits it on `gen_ai.client.inference.operation.details`
  - includes a beta Messages API example where the value is derived from a returned compaction block, `stop_reason: "compaction"`, compaction usage iteration, or a compaction block passed back in input messages
  - records compaction blocks as `CompactionPart` in `gen_ai.input.messages` / `gen_ai.output.messages`

The generated reports show support in:

- `reference/reports/inference-span.md`
- `reference/reports/gen-ai-client-inference-operation-details-event.md`

The scenarios were run against the local mock server and validated with Weaver live-check. The checked-in `data.json` files and reports were regenerated.

## Checklist

See [CONTRIBUTING.md](CONTRIBUTING.md).


reference/scenarios/anthropic/scenario.py
```diff
@@ -41,6 +41,7 @@ def run_chat():
     if port is not None:
         span_attributes["server.port"] = port
     with _reference_tracer.start_as_current_span("chat claude-sonnet-4-20250514", attributes=span_attributes) as span:
+        span.set_attribute("gen_ai.conversation.compacted", True)
```
lmolkova: where would this value come from in a real instrumentation? It seems to be hard-coded and it's not clear if shared app-agnostic instrumentation would be able to know whether the conversation is compacted

reference/scenarios/openai/scenario.py
```diff
@@ -28,9 +28,14 @@ def run_chat_reference(client):
...
     request_top_p = 0.9
+    previous_messages = [
+        "User asked for a friendly greeting.",
+        "Assistant should keep the answer short.",
+    ]
+    compacted_summary = "Compacted prior conversation: " + " ".join(previous_messages)
     messages = [
         {"role": "system", "content": "You are a helpful assistant."},
-        {"role": "user", "content": "Say hello."},
+        {"role": "user", "content": f"{compacted_summary}\n\nCurrent request: Say hello."},
```
eternalcuriouslearner:
```suggestion
        {"role": "user", "content": compacted_user_content},
```

reference/scenarios/anthropic/scenario.py
eternalcuriouslearner: @lmolkova good question!! this comes from server side api. We are going to add this convention only when user enabled server side compaction. But we can extend it if we want to do automatic compaction and it's out of scope for now. I wrote down a small example and it's raw but it conveys how we can capture it.

Examples:

**OpenAI Responses API**

  A request may enable compaction with `context_management`:

  ```python
  response = client.responses.create(
      model="gpt-5.4",
      input=conversation_so_far,
      context_management=[{"type": "compaction", "compact_threshold": 200000}],
      store=False,
  )

  That request option means compaction is allowed/configured. The actual signal comes from the response output. When compaction runs, the Responses API can emit a ResponseCompactionItem:

  {
    "type": "compaction",
    "id": "cmp_...",
    "encrypted_content": "...",
    "created_by": "server"
  }

  Instrumentation can derive the attribute from that returned item:

  conversation_compacted = any(
      getattr(item, "type", None) == "compaction"
      for item in response.output
  )

  span.set_attribute("gen_ai.conversation.compacted", conversation_compacted)
  ```

For streaming Responses, the same logic applies to output item events: if a streamed response.output_item.added / response.output_item.done event contains an item with type == "compaction", the span can be marked compacted.

**Anthropic Messages API**

Similarly, Anthropic supports server-side context management through context_management.edits, for example:

```python
  response = client.beta.messages.create(
      model="claude-sonnet-4-20250514",
      max_tokens=1024,
      messages=messages,
      context_management={
          "edits": [
              {
                  "type": "compact_20260112",
                  "trigger": {"type": "input_tokens", "value": 200000},
                  "pause_after_compaction": True,
              }
          ]
      },
      betas=["context-management-2026-01-12"],
  )

  Again, the request option means compaction is configured, not necessarily that it occurred. The actual signal can come from the response. Anthropic may expose compaction as:

  {
    "type": "compaction",
    "encrypted_content": "..."
  }

  or through:

  {
    "stop_reason": "compaction"
  }

  or through usage iterations:

  {
    "usage": {
      "iterations": [
        {
          "type": "compaction",
          "input_tokens": 123456,
          "output_tokens": 789
        }
      ]
    }
  }

  Instrumentation can derive the attribute from those provider-returned signals:

  conversation_compacted = (
      response.stop_reason == "compaction"
      or any(block.type == "compaction" for block in response.content)
      or any(iteration.type == "compaction" for iteration in response.usage.iterations)
  )

  span.set_attribute("gen_ai.conversation.compacted", conversation_compacted)
```

I can expand more on this if this example is ambiguous and doesn't answer your question. Lmk your thoughts.

reference/scenarios/openai/scenario.py
```diff
@@ -130,6 +154,53 @@ def run_chat_reference(client):
...
+    conversation = [
+        {
+            "type": "message",
+            "role": "user",
+            "content": "Let's continue a long implementation task.",
+        }
+    ]
+    host, port = mock_server_host_port(MOCK_BASE_URL)
+    span_attributes = {
+        "gen_ai.operation.name": "responses",
```
JWinermaSplunk: Should we use a valid `gen_ai.operation.name`?

reference/scenarios/anthropic/scenario.py
```diff
@@ -112,6 +138,62 @@ def run_chat():
...
+                        "pause_after_compaction": True,
+                    }
+                ]
+            },
+            betas=["context-management-2026-01-12"],
+        )
+
+        # Real Anthropic instrumentation can derive this from a compaction
+        # content block, stop_reason="compaction", or usage iteration with
+        # type="compaction", rather than guessing from the request option.
```
trask: is this comment outdated? is this instrumentation not doing that already? if not, can it? thx

reference/scenarios/openai/scenario.py
```diff
@@ -28,9 +45,14 @@ def run_chat_reference(client):
     request_frequency_penalty = 0.1
     request_presence_penalty = 0.2
     request_top_p = 0.9
+    previous_messages = [
+        "User asked for a friendly greeting.",
+        "Assistant should keep the answer short.",
+    ]
+    compacted_user_content, conversation_compacted = compact_conversation_history(previous_messages, "Say hello.")
```
trask: this looks essentially hardcoded (not something generic instrumentation could capture)?

bonczj:
@eternalcuriouslearner It's not quite clear to me yet what the token counts would represent on the compaction span. From your research, does the call return the before/after token counts from the compaction? If so, will those map to the existing token count attributes, or do we need to introduce new attributes for compaction?

shuwpan:
I am thinking…do we need to track how many times this conversation is being compacted as well ?

trask:
Can you mark this as resolving https://github.com/open-telemetry/semantic-conventions-genai/issues/154 (if that's the case)? thanks!

reference/src/semconv_genai/mock_server/anthropic.py
```diff
@@ -53,6 +53,35 @@
...
 
 
+MESSAGE_COMPACTION_RESPONSE = {
+    "id": "msg-mock-compaction-001",
+    "type": "message",
+    "role": "assistant",
+    "content": [
+        {
+            "type": "compaction",
+            "encrypted_content": "opaque encrypted compaction state",
```
lmolkova, line 63: Anthropic seems to support both -human-readable text and encrypted version -   https://platform.claude.com/docs/en/build-with-claude/compaction#working-with-compaction-blocks

https://github.com/anthropics/anthropic-sdk-python/blob/5db69c6d5f3e7e39023973ca4b235046e5661157/src/anthropic/types/beta/beta_compaction_block.py#L19

reference/src/semconv_genai/mock_server/openai.py
```diff
@@ -354,4 +354,16 @@ def responses():
...
+    context_management = body.get("context_management") or []
+    if any(item.get("type") == "compaction" for item in context_management if isinstance(item, dict)):
+        resp["output"].insert(
+            0,
+            {
+                "type": "compaction",
+                "id": "cmp-mock-001",
+                "encrypted_content": "opaque encrypted compaction state",
+                "created_by": "server",
+            },
```
lmolkova: since it's a part of output and also could be passed as input top the next turn, I think we should model it as message part:
- it's important to know if we properly preserved it in the chat history
- the order matters - did we drop things before it and appended next messages after. Anthropic docs  have example that shows how to add 3 latest messages after compaction item https://platform.claude.com/docs/en/build-with-claude/compaction#working-with-compaction-blocks
- the content matters (unencrypted in case of anthropic), we need to capture it. We don't need to include encrypted content now
- it's part of input and output according to model APIs and should be in the corresponding attributes.

We can't express it all with boolean flag.

Given this (we'll have part with type `compaction` which would include optional content, possibly compaction id  and maybe some other properties, would we still need boolean flag? Can we live without it?

eternalcuriouslearner:
> @eternalcuriouslearner It's not quite clear to me yet what the token counts would represent on the compaction span. From your research, does the call return the before/after token counts from the compaction? If so, will those map to the existing token count attributes, or do we need to introduce new attributes for compaction?

@bonczj  I'm assuming you're more interested in understanding token counts before and after compaction, and whether we'll get them when server-side compaction happens. Lmk if I got your question right. But overall, I don't think the call returns standardized before/after compaction token counts.

The existing usage fields should continue to represent the token counts for the effective model operation on that span. After compaction, that means `gen_ai.usage.input_tokens` reflects the compacted context used by the model, not the full historical conversation before compaction.

Consumers can still observe the before/after effect by comparing adjacent spans with the same gen_ai.conversation.id:

Prior span: input tokens have grown near the compaction threshold
Compaction span: gen_ai.conversation.compacted=true
Following/current span: input tokens reflect the compacted effective context, so they may drop/reset

I think we don't need additional attributes for token counts. Lmk your thoughts.

eternalcuriouslearner:
> I am thinking…do we need to track how many times this conversation is being compacted as well ?

@shuwpan I don't think we need a separate compaction-count attribute right now. Adding one would require our instrumentation library to maintain state and track how many times compaction occurred across a multi-turn conversation.

Each span should report whether that operation used a compacted conversation view. Consumers can derive the count by aggregating spans where `gen_ai.conversation.compacted=true`, grouped by `gen_ai.conversation.id`.

This avoids requiring instrumentation to maintain conversation-level state across turns, retries, processes, or provider/framework boundaries.

lmolkova (commented):
I realized I didn't post my comments I left a few days ago 🤦‍♀️

reference/src/semconv_genai/mock_server/openai.py
eternalcuriouslearner: @lmolkova I agree that a compaction block should be modeled as a message part when it is present in provider input/output. The block is part of the conversation state, its order is semantically important, and for providers like Anthropic it must be passed back on later turns to preserve the compacted history. A boolean alone cannot express whether the block was emitted, where it appeared, what summary/content it carried, or whether later requests preserved it.

However  the boolean is also useful for token accounting. Once a conversation is compacted, subsequent inference spans may show a much lower `gen_ai.usage.input_tokens` count because the effective model input is now a compacted representation rather than the full prior conversation. 
`gen_ai.conversation.compacted=true` lets operators distinguish a real workload-size change from a context-management event. It also makes before/after analysis straightforward: filter by `gen_ai.conversation.id`, find the spans before and after the first compacted turn, and compare token usage around that transition.

So I would model this as both:
- a `compaction` message part in `gen_ai.output.messages` when the provider emits one, and in `gen_ai.input.messages` when the client passes one back;
- `gen_ai.conversation.compacted=true` on the inference span when the effective input/output for that call involved compaction.

I believe if we only add the message part, users can reconstruct the boolean in ideal traces, but may miss it if they're not recording messages due to PII concerns. On the other hand, upon deeper reflection, if we only add the boolean, users lose the correctness/debugging view of whether the compacted state was preserved. Hence, I'm thinking of adding `CompactionPart` as:

```
  CompactionPart:
    type: object
    additionalProperties: true
    properties:
      type:
        const: compaction
      content:
        type: string
        description: The unencrypted compacted conversation summary, when available.
      id:
        type: string
        description: Provider-assigned identifier for the compaction item/block, when available.
    required:
      - type
```

Lmk your thoughts.

lmolkova: thanks for the context! 

I think boolean flag makes sense (in addition to message parts), the compelling case for me is metrics - we'll never record parts on metric but it'd still be useful to know if conversation was compacted.

We might also want to record the input params for compaction - either as individual attributes or inside a bag of extended input params along with everything else. It can be a follow up

eternalcuriouslearner: Sure I can create issue for adding metrics around compaction.

model/gen-ai/registry.yaml
```diff
@@ -241,6 +241,16 @@ attributes:
     brief: The unique identifier for a conversation (session, thread), used to store and correlate messages within this conversation.
     examples: ["conv_5j66UpCpwteGg4YSxUnt7lPY"]
     stability: development
+  - key: gen_ai.conversation.compacted
+    type: boolean
+    brief: Indicates whether the effective conversation history used for this operation is a compacted view of a longer prior conversation.
+    note: |
+      Compaction may include summarization, pruning, sliding-window selection, hierarchical memory, or
+      provider/framework-managed context reduction.
+
```
lmolkova: is this important to have in semantic conventions?

model/gen-ai/registry.yaml
```diff
@@ -241,6 +241,16 @@ attributes:
     brief: The unique identifier for a conversation (session, thread), used to store and correlate messages within this conversation.
     examples: ["conv_5j66UpCpwteGg4YSxUnt7lPY"]
     stability: development
+  - key: gen_ai.conversation.compacted
+    type: boolean
+    brief: Indicates whether the effective conversation history used for this operation is a compacted view of a longer prior conversation.
```
```suggestion
    brief: Indicates whether the effective conversation context used for this operation is a compacted view of a prior conversation.
```

nit, just slightly shorter

model/gen-ai/registry.yaml
```diff
@@ -241,6 +241,16 @@ attributes:
...
+    note: |
+      Compaction may include summarization, pruning, sliding-window selection, hierarchical memory, or
+      provider/framework-managed context reduction.
+
+      If unset, the compaction state is unknown.
```
lmolkova: do we need to define what compaction is in semconv?


```suggestion
      Instrumentations SHOULD set this attribute to `true` only when they can reliably
  determine that context compaction was applied.
```

model/gen-ai/spans.yaml
```diff
@@ -214,6 +214,14 @@ spans:
           Application developers that manage conversation history MAY add conversation id to GenAI and other
           spans or logs using custom span or log record processors or hooks provided by instrumentation
           libraries.
+      - ref: gen_ai.conversation.compacted
+        requirement_level:
+          recommended: when known
```
lmolkova: nit, I think this is one of a common incantations we use, just for consistency

```suggestion
          recommended: when available
```

model/gen-ai/spans.yaml
```diff
@@ -214,6 +214,14 @@ spans:
...
+        note: |
+          Set to `true` when the provider or client SDK compacts conversation
+          history server-side or in the client before invoking the model
+          (e.g., managed memory APIs, server-side conversation state,
+          context-window auto-truncation).
```
```suggestion
```

do we need to override note?  let's just write note on the attribute definition and reuse it unless there is a need to override.

reference/scenarios/anthropic/scenario.py
```diff
@@ -20,6 +20,44 @@
...
+def response_output_messages(response):
+    """Convert Anthropic response content blocks into OTel output messages."""
+    parts = []
+    for block in getattr(response, "content", []) or []:
+        block_type = getattr(block, "type", None)
+        if block_type == "text" and getattr(block, "text", None):
+            parts.append({"type": "text", "content": block.text})
+        elif block_type == "compaction":
+            compaction_part = {"type": "compaction"}
+            compaction_id = getattr(block, "id", None)
```
lmolkova: don't think anthropic has id in the API - https://github.com/anthropics/anthropic-sdk-python/blob/main/src/anthropic/types/beta/beta_compaction_block.py, https://platform.claude.com/docs/en/build-with-claude/compaction#working-with-compaction-blocks

reference/scenarios/anthropic/scenario.py
```diff
@@ -112,6 +141,92 @@ def run_chat():
...
+                ]
+            },
+            betas=["context-management-2026-01-12"],
+        )
+
+        # Provider-side instrumentation derives this from the live SDK
+        # response: a compaction content block, stop_reason="compaction", or
+        # usage iteration with type="compaction". The request option alone is
+        # not enough.
+        conversation_compacted = response_has_compaction_block(resp)
```
lmolkova: should we check if response or input messages had compacted item?

I.e. when I continue conversation and pass compaction item back, shouldn't we also mark that conversation as compacted?

lmolkova (commented):
Thanks for the updated! mostly minor comments now

reference/scenarios/anthropic/scenario.py
```diff
@@ -20,6 +20,82 @@
...
+            if compaction_content:
+                compaction_part["content"] = compaction_content
+            parts.append(compaction_part)
+    if not parts:
+        return []
+    return [
+        {
+            "role": "assistant",
+            "parts": parts,
+            "finish_reason": getattr(response, "stop_reason", None) or "stop",
```
lmolkova, line 94: since anthropic has 'compaction' as a stop reason and we're using it for openai as well, we should add it as a value in finish reason enum

reference/scenarios/openai/scenario.py
```diff
@@ -16,6 +16,50 @@
...
+    for item in getattr(response, "output", []) or []:
+        item_type = item.get("type") if isinstance(item, dict) else getattr(item, "type", None)
+        if item_type == "compaction":
+            compaction_part = {"type": "compaction"}
+            compaction_id = item.get("id") if isinstance(item, dict) else getattr(item, "id", None)
+            if compaction_id:
+                compaction_part["id"] = compaction_id
+            output_messages.append(
+                {
+                    "role": "assistant",
```
lmolkova: here we model it as a separate message and have artificial role and finish reason, while for anthropic case it's a part of an existing message. Can we do the same here? is there a case when it would not work?

reference/scenarios/anthropic/scenario.py
eternalcuriouslearner: Yep Liudmila, added it as a finish reason enum.

reference/scenarios/openai/scenario.py
eternalcuriouslearner: Yeah we can simulate it as we have in anthropic. I modified it. Afaik for responses.create, compaction-only output is not the primary documented shape; that is more plausible for standalone /responses/compact, while server-side compaction emits a compaction item and then continues inference in the same response. So I believe openai and anthropic have same shape.

model/gen-ai/spans.yaml
```diff
@@ -214,6 +214,9 @@ spans:
           Application developers that manage conversation history MAY add conversation id to GenAI and other
           spans or logs using custom span or log record processors or hooks provided by instrumentation
           libraries.
+      - ref: gen_ai.conversation.compacted
```
lmolkova, line 225: should we reference it on the operation duration metric? seems to be useful there

lmolkova (approved):

bonczj:
@eternalcuriouslearner sorry I wasn't able to respond earlier. No blocking notes though :) 

> > @eternalcuriouslearner It's not quite clear to me yet what the token counts would represent on the compaction span. From your research, does the call return the before/after token counts from the compaction? If so, will those map to the existing token count attributes, or do we need to introduce new attributes for compaction?
> 
> @bonczj I'm assuming you're more interested in understanding token counts before and after compaction, and whether we'll get them when server-side compaction happens. Lmk if I got your question right. But overall, I don't think the call returns standardized before/after compaction token counts.

Yes, since this is about compaction, I usually think about this as compaction of your context. If that is the case, then knowing what the context was before/after is a great way to see how effective your compaction was.

> 
> The existing usage fields should continue to represent the token counts for the effective model operation on that span. After compaction, that means `gen_ai.usage.input_tokens` reflects the compacted context used by the model, not the full historical conversation before compaction.
> 
> Consumers can still observe the before/after effect by comparing adjacent spans with the same gen_ai.conversation.id:
> 
> Prior span: input tokens have grown near the compaction threshold Compaction span: gen_ai.conversation.compacted=true Following/current span: input tokens reflect the compacted effective context, so they may drop/reset
> 
> I think we don't need additional attributes for token counts. Lmk your thoughts.

I agree we could reuse the existing input / output token attributes. I do think having them on this span would be useful rather than trying to sum up the context size from earlier calls, if the APIs return them.

reference/scenarios/openai/scenario.py
```diff
@@ -16,6 +16,51 @@
...
+        pending_parts = []
+        for block in content or []:
+            block_type = block.get("type") if isinstance(block, dict) else getattr(block, "type", None)
+            if block_type != "output_text":
+                continue
+            text = block.get("text") if isinstance(block, dict) else getattr(block, "text", None)
+            if text:
+                parts.append({"type": "text", "content": text})
+        if parts:
+            output_messages.append({"role": role or "assistant", "parts": parts})
```
JWinermaSplunk: Should we also add `finish_reason` to align with the output message schema here?

docs/gen-ai/non-normative/models.ipynb
```diff
@@ -209,6 +220,7 @@
     "    FilePart,\n",
     "    UriPart,\n",
     "    ReasoningPart,\n",
+    "    CompactionPart,\n",
```
JWinermaSplunk: This PR scopes `CompactionPart` to `gen_ai.input.messages` and `gen_ai.output.messages`, but I believe adding it to the shared `MessagePart` alias also affects `SystemInstructions` upon doc regeneration. Is this intended?

JWinermaSplunk (approved):

trask:

Create `changelog.d/162.enhancement.md` containing the change log entry, e.g.:

```md
Add `gen_ai.conversation.compacted` on GenAI inference spans and `CompactionPart` message schema support for reporting compacted GenAI conversation context.
```

Thanks!

model/gen-ai/spans.yaml
eternalcuriouslearner: Yes @lmolkova we should reference it there. I will add this to a new issue that I am creating.

model/gen-ai/registry.yaml
```diff
@@ -258,6 +258,14 @@ attributes:
     brief: The unique identifier for a conversation (session, thread), used to store and correlate messages within this conversation.
     examples: ["conv_5j66UpCpwteGg4YSxUnt7lPY"]
     stability: development
+  - key: gen_ai.conversation.compacted
+    type: boolean
+    brief: Indicates whether the effective conversation context used for this operation is a compacted view of a prior conversation.
+    note: |
+      Instrumentations SHOULD set this attribute to `true` only when they can reliably
+      determine that context compaction was applied.
```
trask: should it ever be set to `false`?

eternalcuriouslearner: @trask I am under the assumption that`gen_ai.conversation.compacted` will be emitted only when compaction is known to have occurred. If instrumentation does not know, or no compaction signal is exposed by the provider/framework, we should leave the attribute unset rather than emit false. This is because imo emitting false can be misleading because absence of an observed compaction event is not the same as knowing compaction did not happen, especially for provider-side behavior(please note only anthropic and openai are doing server side compaction currently). So the useful states are: unset = unknown/not reported, true = compaction occurred. But lmk if you think we can do it differently I am open to changing this.

trask: > absence of an observed compaction event is not the same as knowing compaction did not happen

what if you _know_ that compaction hasn't occurred? should you set it to `false` or leave it blank?

eternalcuriouslearner: I'm leaning towards leaving it blank.

trask: I don't think we have precedent for this.

Here's the closest that we have:

`gen_ai.request.stream`

> [If unset, the request is assumed to be non-streaming.](model/gen-ai/spans.yaml#L184-L186)

`messaging.destination.temporary`

> [When missing, the value is assumed to be `false`.](https://github.com/open-telemetry/semantic-conventions/blob/085cf2230203d9fb552edf94e035ee9da64ec39d/model/messaging/spans.yaml#L53-L55)

`messaging.destination.anonymous`

> [When missing, the value is assumed to be `false`.](https://github.com/open-telemetry/semantic-conventions/blob/085cf2230203d9fb552edf94e035ee9da64ec39d/model/messaging/spans.yaml#L56-L58)

`messaging.kafka.message.tombstone`

> [When missing, the value is assumed to be `false`.](https://github.com/open-telemetry/semantic-conventions/blob/085cf2230203d9fb552edf94e035ee9da64ec39d/model/messaging/spans.yaml#L160-L162)

Would this work?

Otherwise, we could consider something like this:

```suggestion
      This attribute is a positive indicator of context compaction. Instrumentations
      SHOULD set it to `true` only when they can reliably determine that context
      compaction was applied. Instrumentations SHOULD NOT set it to `false`; they
      SHOULD leave it unset otherwise.
```

eternalcuriouslearner: @trask I like your suggestion. I used it. Thanks for the help.

MikeGoldsmith (approved):

PR #164, opened 2026-05-17
Add gen_ai.server.inter_token_latency metric

Jwrede:
## Summary

Add a server-side `gen_ai.server.inter_token_latency` histogram metric that records the time elapsed between consecutive output tokens during generation.

This complements the existing `gen_ai.server.time_per_output_token` (TPOT), which is a per-request average computed after the request completes. ITL is recorded for each pair of consecutive tokens and captures the per-token timing variability caused by batching, preemption, or multi-token prediction.

## Motivation

- vLLM has deprecated TPOT in favor of ITL ([vllm-project/vllm#24015](https://github.com/vllm-project/vllm/pull/24015)) because TPOT as an average hides per-token variability, especially with multi-token prediction (MTP) where multiple tokens are generated at once.
- TGI and NIM also report ITL natively.
- The current spec has no way to express per-token-pair latency distribution, only the per-request average.

## Changes

- `model/gen-ai/metrics.yaml`: Add `gen_ai.server.inter_token_latency` metric definition
- `docs/gen-ai/gen-ai-metrics.md`: Add metric documentation section with bucket boundaries
- `CHANGELOG.md`: Add entry under Enhancements

Refs: open-telemetry/semantic-conventions-genai#232


model/gen-ai/metrics.yaml
```diff
@@ -87,6 +87,23 @@ metrics:
...
     attributes:
       - ref_group: metric_attributes.gen_ai
+  - name: gen_ai.server.inter_token_latency
+    instrument: histogram
+    unit: "s"
+    brief: 'Time between consecutive output tokens during generation for successful responses.'
+    note: >
+      Inter token latency measures the actual elapsed time between the emission of
+      consecutive output tokens. Unlike `gen_ai.server.time_per_output_token` which
+      is a per-request average, this metric is recorded for each pair of consecutive
```
lmolkova, line 97: where does the assumption on "per-request average" come from? 

Is there a need to introduce a new metric when we can just clarify the definition of the old one?

Jwrede:
@lmolkova Good questions.

**"Per-request average" framing:** This comes from vLLM's deprecation rationale in vllm-project/vllm#24015. TPOT is computed as (total_generation_time - TTFT) / (output_tokens - 1) and recorded once per request. That single value is an average over all tokens in the request.

**Why a new metric instead of clarifying TPOT:** The two metrics have fundamentally different recording cardinality:

- TPOT: 1 histogram observation per request (the averaged value)
- ITL: N-1 histogram observations per request (one per consecutive token pair)

Redefining TPOT to record per-token-pair would change its semantics and break existing consumers. The histogram distribution shape is completely different. TPOT has lower variance (smoothed by averaging); ITL captures the full distribution including spikes from batching, preemption, and multi-token prediction.

That said, if the SIG prefers a different approach (e.g. deprecating TPOT in favor of ITL, or merging the concepts), I am happy to adjust. The linked issue (open-telemetry/semantic-conventions-genai#232) left this open for discussion.

lmolkova:
@Jwrede the suggestion is to clarify description of the existing metric in OTel to align with the spirit behind ITL (which I believe it already does). OTel metrics are independent of vLLM definitions.

Cirilla-zmh:
> **"Per-request average" framing:** This comes from vLLM's deprecation rationale in [vllm-project/vllm#24015](https://github.com/vllm-project/vllm/pull/24015). TPOT is computed as (total_generation_time - TTFT) / (output_tokens - 1) and recorded once per request. That single value is an average over all tokens in the request.

TPOT can be calculated either after all requests have returned, or in the manner of ITL as you mentioned. I believe these two approaches are essentially equivalent, at least in terms of average values. I agree with Liudmila's perspective — in my view, TPOT and ITL refer to the same concept, and if you feel that TPOT is lacking in some way, perhaps the more appropriate approach would be to enrich and extend it rather than introducing a separate metric.

trask:

Create `changelog.d/164.enhancement.md` containing the change log entry, e.g.:

```md
Add `gen_ai.server.inter_token_latency` metric to track the time between consecutive output tokens during generation.
```

Thanks!

PR #184, opened 2026-05-21
gen-ai: add gen_ai.response.id to deepeval evaluation result event

hippoley:
The `deepeval` reference scenario demonstrates the `gen_ai.evaluation.result` event with the attributes that deepeval can credibly capture at the current call boundary.

### Changes

- `reference/scenarios/deepeval/scenario.py`: emit `gen_ai.evaluation.name`, `gen_ai.evaluation.score.value`, and `gen_ai.evaluation.explanation` on the `gen_ai.evaluation.result` event.
- `reference/scenarios/deepeval/data.json`: updated to reflect captured attributes.
- `reference/reports/gen-ai-evaluation-result-event.md`: regenerated.

### Honest capture gap

`gen_ai.response.id` is **not** captured here. deepeval is an evaluation framework that does not make LLM calls itself — the response id would have to come from an upstream inference span (cross-call value). Per the reference scenario guidelines, cross-call and hardcoded values are not forced into reference implementations.

In production pipelines the caller can pass the upstream `gen_ai.response.id` into the evaluation step; the reference scenario cannot demonstrate this without synthetic scaffolding.

### Verification

```bash
cd reference/
uv run run-scenario deepeval   # exit 0
uv run update-reports          # report updated
```


hippoley:
Thanks for the review feedback!

Addressed in the latest commit (0de0712):

1. **Double spaces** — fixed to single spaces in both comment blocks.
2. **Configurable mock ID** — `_MOCK_RESPONSE_ID` now reads from `MOCK_RESPONSE_ID` env var with the original value as default, keeping the scenario reproducible while allowing override in test environments.

reference/scenarios/deepeval/scenario.py
```diff
@@ -1,4 +1,13 @@
...
+"""Reference implementation for DeepEval.
+
+Changes vs upstream:
+- Added gen_ai.response.id to the evaluation result event.
+  This field correlates the evaluation back to the specific LLM completion
+  being evaluated, as required by the gen_ai.evaluation.result event spec.
+  In a real pipeline the response_id comes from the LLM call that produced
+  the output being evaluated; here we use a deterministic mock value so the
+  reference scenario stays self-contained and reproducible.
+"""
```
lmolkova: is this comment necessary here? there is a similar one on lines 21-25

lmolkova (approved):

reference/scenarios/deepeval/scenario.py
```diff
@@ -9,6 +18,12 @@
 
 _reference_tracer = reference_tracer()
 
+# Simulates the response id from the upstream LLM call whose output is
+# being evaluated. In production this would be captured from the inference
+# span (gen_ai.response.id) and passed into the evaluation step.
+# Override via MOCK_RESPONSE_ID env var when a specific value must be asserted.
+_MOCK_RESPONSE_ID = os.environ.get("MOCK_RESPONSE_ID", "chatcmpl-mockEvalRef0001")
```
trask: this looks like a hardcoded value? check out https://github.com/open-telemetry/semantic-conventions-genai/blob/main/.github/skills/reference/SKILL.md

hippoley:
Thanks for the feedback, @trask and @lmolkova!

You're right — `gen_ai.response.id` is a cross-call value for deepeval: the evaluation framework doesn't make LLM calls itself, so the response id would have to come from an upstream inference span. That makes it an **honest capture gap** per the reference scenario guidelines.

Latest commit (aa104c4) removes `_MOCK_RESPONSE_ID` and the `gen_ai.response.id` attribute entirely. The duplicate comment is also gone.

The PR now only demonstrates what deepeval can credibly capture at the current call boundary: `gen_ai.evaluation.name`, `gen_ai.evaluation.score.value`, and `gen_ai.evaluation.explanation`.

trask:
hi @hippoley, not sure what happened, but PR content appears empty now, thanks

MikeGoldsmith (changes requested):
Adding a Request Changes as changes have been lost in the last force-push

hippoley:
Closing this one rather than keeping an empty PR alive. The force-push lost the original diff, and current `main` already contains the DeepEval evaluation-result attributes that remained after review (`gen_ai.evaluation.name`, `gen_ai.evaluation.score.value`, and `gen_ai.evaluation.explanation`). The rejected cross-call `gen_ai.response.id` idea should stay rejected rather than be reconstructed with synthetic data. Thanks for the reviews that clarified that boundary.

PR #185, opened 2026-05-21
gen-ai: add evaluation operation name and gen_ai.evaluate.internal span

hippoley:
Closes / relates to: https://github.com/open-telemetry/semantic-conventions/issues/3398

### Problem

The `gen_ai.evaluation.result` event was merged in #2563 and is already used by multiple reference scenarios (deepeval, azure-ai-evaluation, dspy). However there is currently no standard **span type** to carry it, and no well-known `gen_ai.operation.name` value for evaluation operations.

This causes two concrete problems:

1. **Invisible in dashboards** — evaluation spans cannot be queried or filtered at the operation level alongside `invoke_agent`, `execute_tool`, and `chat` spans.
2. **Ecosystem fragmentation** — OpenSearch's [genai-observability-sdk-py](https://github.com/opensearch-project/opensearch-genai-observability-py) already uses `gen_ai.operation.name: "evaluation"` as a custom value. Without a standard, every instrumentation library invents its own string.

### Changes

**`model/gen-ai/registry.yaml`** — add `evaluation` enum member to `gen_ai.operation.name`:

```yaml
- id: evaluation
  value: "evaluation"
  brief: 'Evaluation of GenAI output quality, accuracy, or other characteristics'
  note: >
    Used when the operation being instrumented is an evaluation step,
    such as running an LLM-as-judge, a scoring metric, or a benchmark suite.
    The span SHOULD carry one or more gen_ai.evaluation.result events.
  stability: development
```

**`model/gen-ai/spans.yaml`** — add `gen_ai.evaluate.internal` span:

- `gen_ai.operation.name` = `evaluation` (required)
- `gen_ai.evaluation.name` (conditionally required: single named metric)
- `gen_ai.provider.name` (conditionally required: when using a judge model)
- `gen_ai.request.model` (conditionally required: when using a judge model)
- `gen_ai.conversation.id` (conditionally required: conversation-scoped eval)
- `gen_ai.agent.name` (conditionally required: agent-scoped eval)

Span name SHOULD be `evaluate {gen_ai.evaluation.name}` for a single metric, or `evaluate` when multiple metrics share one span.

### Design notes

- This is a purely additive change; no existing attributes or spans are modified.
- The span is `INTERNAL` kind, consistent with `gen_ai.execute_tool.internal` and `gen_ai.plan.internal`.
- A follow-up PR will add a reference scenario demonstrating the span in a benchmark context (deepeval / azure-ai-evaluation).

### Verification

```bash
# Weaver resolves the updated registry without new errors
cd reference/
uv run run-scenario deepeval   # exit 0, no new violations
```

### DIFF
--- a/model/gen-ai/registry.yaml
+++ b/model/gen-ai/registry.yaml
@@ -453,7 +453,7 @@ attributes:
           note: >
             Used when the operation being instrumented is an evaluation step,
             such as running an LLM-as-judge, a scoring metric, or a benchmark suite.
-            The span SHOULD carry one or more gen_ai.evaluation.result events.
+            The span SHOULD carry one or more `gen_ai.evaluation.result` events.
           stability: development
     brief: The name of the operation being performed.
     note: >
--- a/model/gen-ai/registry.yaml
+++ b/model/gen-ai/registry.yaml
@@ -447,6 +447,14 @@ attributes:
           value: "delete_memory_store"
           brief: 'Delete or deprovision a memory store'
           stability: development
+        - id: evaluation
+          value: "evaluation"
+          brief: 'Evaluation of GenAI output quality, accuracy, or other characteristics'
+          note: >
+            Used when the operation being instrumented is an evaluation step,
+            such as running an LLM-as-judge, a scoring metric, or a benchmark suite.
+            The span SHOULD carry one or more gen_ai.evaluation.result events.
+          stability: development
     brief: The name of the operation being performed.
     note: >
       If one of the predefined values applies, but specific system uses a different name it's RECOMMENDED to document it in the semantic
--- a/model/gen-ai/spans.yaml
+++ b/model/gen-ai/spans.yaml
@@ -572,6 +572,58 @@ spans:
         requirement_level:
           conditionally_required: when available
 
+  - type: gen_ai.evaluate.internal
+    kind: internal
+    name:
+      note: >
+        **Span name** SHOULD be `evaluate {gen_ai.evaluation.name}` when a
+        single named metric is being evaluated, or `evaluate` when multiple
+        metrics are evaluated within the same span.
+    brief: >
+      Represents an evaluation of GenAI output quality, accuracy, or other
+      characteristics. Carries one or more `gen_ai.evaluation.result` events.
+    note: |
+      The `gen_ai.operation.name` SHOULD be `evaluation`.
+
+      **Span name** SHOULD be `evaluate {gen_ai.evaluation.name}` when a
+      single named metric is being evaluated, or `evaluate` when multiple
+      metrics are evaluated within the same span.
+
+      This span SHOULD be the parent of `gen_ai.evaluation.result` events.
+      When evaluation is performed post-hoc (separate process from inference),
+      `gen_ai.response.id` on the event correlates the evaluation back to the
+      original inference span.
+
+      Instrumentations SHOULD report this span when they can reliably
+      determine that the operation being instrumented is an evaluation step,
+      and SHOULD NOT report it for generic inference spans that happen to
+      include self-evaluation logic.
+
+      Examples: DeepEval GEval, Azure AI Evaluation, DSPy Evaluate,
+      RAGAS, custom LLM-as-judge pipelines.
+    stability: development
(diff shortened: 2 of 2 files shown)


hippoley:
Thanks for the review feedback!

Addressed in the latest commit (9122ae0):

1. **Naming inconsistency** — renamed `gen_ai.evaluate.internal` → `gen_ai.evaluation.internal` and updated span name guidance from `evaluate ...` → `evaluation ...` to align with the operation name value (`evaluation`) and the existing `gen_ai.evaluation.result` event name.
2. **Backtick formatting** — added backticks around `gen_ai.evaluation.result` in the registry.yaml note.

hippoley:
Note on CI: the `required-status-check` failure is caused by the `claude-agent-sdk` scenario failing on this branch, which is a pre-existing issue on `main` unrelated to this PR's changes (this PR only touches `model/gen-ai/registry.yaml` and `model/gen-ai/spans.yaml`).

The root cause has been identified and fixed in PR #186: `claude-agent-sdk 0.2.x` changed the `AssistantMessage` API (fields moved from `message.message.*` to direct attributes), causing `scenario.py` to silently skip setting the response attributes. PR #186 now includes the fix.

model/gen-ai/registry.yaml
```diff
@@ -447,6 +447,14 @@ attributes:
           value: "delete_memory_store"
           brief: 'Delete or deprovision a memory store'
           stability: development
+        - id: evaluation
```
Cirilla-zmh, line 450: I think we need a verb here.

model/gen-ai/registry.yaml
```diff
@@ -447,6 +447,14 @@ attributes:
...
+          value: "evaluation"
+          brief: 'Evaluation of GenAI output quality, accuracy, or other characteristics'
+          note: >
+            Used when the operation being instrumented is an evaluation step,
```
Cirilla-zmh, line 454: I'm wondering which instrumentation should generate this kind of span? A kind of evaluation framework? Could you show a prototype for this?

singankit:
@hippoley  Idea of evaluation result as event was that it can be linked to the span via traceId, spanId for any span be it invoke_agent span , chat span etc.

model/gen-ai/spans.yaml
```diff
@@ -572,6 +572,58 @@ spans:
...
+        requirement_level: required
+        sampling_relevant: true
+      - ref: gen_ai.evaluation.name
+        requirement_level:
+          conditionally_required: when a single named metric is being evaluated
+        sampling_relevant: true
+      - ref: gen_ai.provider.name
+        requirement_level:
+          conditionally_required: when the evaluation uses a GenAI judge model
+      - ref: gen_ai.request.model
```
singankit, line 617: This can be different for different metric being evaluated, hence having this attribute for it on evaluation span does not seem relevant.

model/gen-ai/spans.yaml
```diff
@@ -572,6 +572,58 @@ spans:
...
+        requirement_level:
+          conditionally_required: when the evaluation uses a GenAI judge model
+      - ref: gen_ai.conversation.id
```
singankit, line 620: Same feedback applies to this attribute too as request.model attribute.

singankit:
Can you please help provide clarity on following questions:
- How will evaluation span be associated to operation being evaluated?
- What is evaluation span supposed to capture? The process of evaluation or evaluation results or both?
- Currently evaluation results can be association with span being evaluated. After this change will evaluation result be associated with both evaluation span and span being evaluated?

hippoley:
Closing this version instead of keeping a stale design proposal open. The reviews surfaced a real modeling ambiguity: an evaluation result can already be correlated to the span being evaluated, while a separate evaluation span needs a precise scope, parent/association model, and a concrete instrumentation prototype to justify itself. The current PR does not answer those questions well enough, and its generated-docs check has also drifted with the base.

I’d rather come back with a narrower proposal backed by a real evaluator instrumentation example than standardize a span shape first and reverse-engineer the semantics afterward. Thanks @singankit, @Cirilla-zmh and the reviewers for pushing on the boundary.

Issue #187, opened 2026-05-22
Add `invoke_node {node.name}` span for non-agent workflow nodes


RKest:
## Summary

Extend the GenAI semantic conventions with an `invoke_node` operation, a
`gen_ai.invoke_node.internal` span, a `gen_ai.node.name` attribute, and a
matching `gen_ai.client.workflow.node.invocation.details` event so that
non-agent units of work in a GenAI workflow can be instrumented uniformly
across frameworks.

## Motivation

The conventions already cover:

- `invoke_agent {agent.name}` — for agent invocations
  ([gen-ai-agent-spans.md][agent-spans]).
- `invoke_workflow {workflow.name}` — for the workflows which orchestrate
  agents and other operations
  ([open-telemetry/semantic-conventions#3249][pr-3249]).

There is no standard way to instrument the third common building block of
modern GenAI applications: a **node**, i.e. a unit of work that participates
in a workflow but is not itself a GenAI agent. Examples:

- A LangGraph graph node that runs deterministic Python before/after a model
  call.
- A Google ADK workflow node that is not a `BaseAgent` (e.g. a routing or
  formatting node). ADK already emits `invoke_node {node.name}` spans
  internally — see [`node_tracing.py`][adk-node-tracing] — but the span is
  not part of any released semantic convention. This proposal standardizes
  it.
- A LangChain chain step that does not itself invoke a model or an agent.
- A CrewAI flow can have deterministic python code run before/after agent
  invocations.

Without a convention, instrumentations diverge:

- ADK emits `invoke_node {node.name}` (already implemented but not standardized).
- `opentelemetry-instrumentation-langchain` emits `execute_task {node.name}`
  for LangGraph nodes.
- `opentelemetry-instrumentation-crewai` does not instrument its "flow" steps at all
  ([example][crewai-example]).
- Other frameworks fall back to bespoke or no spans.

Standardizing this gives operators a consistent way to:

1. **Reconstruct the traversed workflow path from a trace.** With
   `invoke_workflow` at the top, `invoke_agent` and `invoke_node` spans as the
   workflow steps, and the LLM / tool / retrieval spans as their children, the
   trace mirrors the path the workflow actually took.
2. **Localize errors to the node where they happen.** A node may run arbitrary
   user code that is not instrumented (or not fully instrumented). Any exception
   in that code surfaces on the nearest enclosing span — without an
   `invoke_node` span, that is the `invoke_workflow` root. The `invoke_node`
[...]

Issue #175, opened 2026-05-19
Add structured/key-value input part type for non-message workflow inputs


wrisa:
Motivation

  LangGraph and similar agent frameworks pass structured state dictionaries as node inputs rather than a messages list. For example, a node may receive:

`  {"user_query": "What is the weather?", "context": {"location": "NYC"}}
`
  When instrumenting such frameworks with the GenAI semantic conventions, there is currently no standard way to represent this structured input in a span or log event. The only option is to serialize the entire dict as a JSON string and wrap it
   in a text part:

`  { "role": "user", "parts": [{ "type": "text", "content": "{\"user_query\": \"What is the weather?\", ...}" }] }
`
  This loses type information, makes the data hard to query, and conflates structured state with free-text user messages.

  Proposed addition

  Add a new MessagePart variant for structured/key-value data, tentatively called structured or data:

  # example schema sketch
```
  type: structured
  data: <map of string → any>  # or an opaque JSON object
```

  This would allow instrumentors to emit:

`  { "role": "user", "parts": [{ "type": "structured", "data": {"user_query": "What is the weather?", "context": {"location": "NYC"}} }] }
`
  Use cases

  - LangGraph nodes: structured state fields passed between graph nodes
  - Workflow orchestrators: inputs that are typed schemas (Pydantic models, TypedDicts) rather than chat messages

  Alternatives considered

  - Serialize to a text part (current workaround) — loses structure, degrades observability tooling
  - Skip capturing the input entirely — loses important tracing context

  References

  - This came up while implementing LangGraph/LangChain chain instrumentation: [PR](https://github.com/open-telemetry/opentelemetry-python-genai/pull/25#discussion_r3263539924)
  - Existing MessagePart variants for reference: text, tool_call_request, tool_call_response, blob, file, uri, reasoning

trask:
Related to #187/#188?

aabmass:
> LangGraph and similar agent frameworks pass structured state dictionaries as node inputs rather than a messages list. For example, a node may receive:
> 
> ` {"user_query": "What is the weather?", "context": {"location": "NYC"}}`

In LangGraph, how does each node know how to consume this? And would the prompt and response that gets sent to the model eventually get embedded in this?

> Add a new MessagePart variant for structured/key-value data, tentatively called structured or data:

If it doesn't get sent to the LLM, I feel like we maybe shouldn't reuse the existing message parts. Is this more like generic "agent state"?

PR #188, opened 2026-05-22
Add step convention

RKest:
## Description

Implements `run_step` from #187, scoped down per the WG review:
this PR adds the span and attribute only. The matching event and the
broader `gen_ai.{input,output}.messages` reshape that earlier revisions
included have been pulled out into a separate, deferred follow-up.

Adds:

- `run_step` enum value on `gen_ai.operation.name` and a new
  `gen_ai.step.name` attribute (`model/gen-ai/registry.yaml`).
- `gen_ai.run_step.internal` span (`model/gen-ai/spans.yaml`).
- Regenerated docs and schema snapshot
  (`docs/gen-ai/gen-ai-agent-spans.md`, `docs/gen-ai/gen-ai-spans.md`,
  `docs/registry/attributes/gen-ai.md`, `schema-snapshot/registry.yaml`,
  and the per-library `docs/gen-ai/{anthropic,aws-bedrock,
  azure-ai-inference,openai,mcp}.md` enum tables).
- Reference tooling registration for the new span in
  `reference/src/semconv_genai/{semconv_model,data_files,report}.py`,
  plus regenerated `reference/reports/*.md` and `reference/README.md`.
- Reference scenarios that emit `run_step` (see Prototype).
- `Unreleased` entry in `CHANGELOG.md`.

## Motivation

See #187 for the full motivation (user journey, prior art across ADK,
LangChain, LangGraph, and CrewAI, the rationale for "step" vs. "task").
This revision tightens the definition based on review feedback:

- `run_step` exists so the execution graph of a workflow can be
  reconstructed from telemetry, particularly in graph-based workflows
  where the output of one step is the input of the next and steps can
  route control between agents, sub-workflows, or other steps.
- It MUST NOT be reported for any unit of work that is already
  instrumented by another GenAI span (no double-instrumentation of
  tool calls, LLM calls, agent invocations, etc.).
- It SHOULD NOT be reported when the framework already provides a
  well-defined agentic loop in which routing and control flow are
  expressed through already-instrumented operations (agent-transfer
  tools, agent-as-tool tool calls). In those frameworks the execution
  graph is already reconstructible from the existing spans.
- Per-framework conventions SHOULD document which framework constructs
  map to `run_step` and which do not.

## Prototype

Reference scenarios demonstrating the new span. Each one wraps a
deterministic, non-agent, non-tool, non-LLM, non-retrieval unit of work
that participates in a workflow graph -- exactly the case the new span
targets.

- **`reference/scenarios/google-adk/scenario.py`** -- emits
  `run_step validate_input` as a sibling of `invoke_agent` inside
  `invoke_workflow`. Models a deterministic input-validation step sitting
  between workflow start and an agent invocation in an ADK 2.x
  graph-based workflow.
- **`reference/scenarios/langchain/scenario.py`** -- adds
  `run_langgraph_step_reference()`, which builds a minimal LangGraph
  `StateGraph` with a deterministic `classify_query` step (emitted as
  `run_step`) and a `respond` LLM-calling step (emitted as `chat`,
  not double-wrapped in `run_step`).
- **`reference/scenarios/crewai/scenario.py`** -- adds
  `run_flow_step_reference()`, which uses CrewAI's `Flow` API
  (`@start` step) to demonstrate an *application-emitted* `run_step`
  for a deterministic Flow step. CrewAI does not instrument Flow steps
  itself; this is the case where applications are expected to emit the
  span.
- **`reference/scenarios/haystack/scenario.py`** -- adds
  `run_pipeline_step_reference()`, which models a Haystack Pipeline
  with a deterministic `query_normalizer` component (`run_step`)
  feeding into a retriever component (`retrieval`, not double-wrapped).

`data.json`, `reference/reports/*.md`, and `reference/README.md` are
regenerated to reflect each scenario's coverage.

DSPy and LlamaIndex were considered (both have step/module
abstractions) but skipped: their existing reference scenarios are
LLM-, tool-, or retrieval-centric and adding a contrived non-LLM step
purely to demonstrate `run_step` would conflict with the tightened
definition. They can be revisited in a follow-up if real upstream
patterns surface.

## Deferred from this PR

The earlier revisions of this PR also included an
`gen_ai.client.workflow.step.invocation.details` event, a broadened
brief+note for `gen_ai.{input,output}.messages`, and an appended
non-structured-payload exception across several attributes. Per WG
review feedback (we don't routinely create an event per span; the
input/output messages attribute meaning shouldn't be broadened where
the exception was misplaced), all of those changes have been removed
from this PR and held locally for a possible follow-up:

- The `gen_ai.client.workflow.step.invocation.details` event.
- The shape-by-operation brief on `gen_ai.input.messages` and
  `gen_ai.output.messages` ("this turn's input on `invoke_agent`",
  "verbatim on `run_step`").
- The "exception applies when input/output itself is not structured"
  sentence on `gen_ai.input.messages`, `gen_ai.output.messages`,
  `gen_ai.tool.definitions`, `gen_ai.retrieval.documents`, and
  `gen_ai.memory.records`.

## PR evolution

Prior, this PR dealt with "nodes". Because of naming conflicts,
the node wording was changed to "step".
See https://github.com/open-telemetry/semantic-conventions-genai/pull/188#pullrequestreview-4760268593 for more details.

## Checklist

- [x] Motivation section filled in above (and in the linked issue)
- [x] Reference scenarios updated for affected libraries (google-adk,
      langchain (LangGraph), crewai (Flow), haystack)


model/gen-ai/events.yaml
```diff
@@ -10,6 +10,32 @@ events:
...
+      independently from the corresponding `invoke_node` span.
+
+      It SHOULD be parented to the corresponding `invoke_node` span when possible.
+
+      See the notes on `gen_ai.input.messages` and `gen_ai.output.messages` for
+      how they behave on `invoke_node` versus on LLM and agent operations.
+    stability: development
+    attributes:
+      - ref: gen_ai.operation.name
+        requirement_level: required
```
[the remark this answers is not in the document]
RKest: Good catch — added `The gen_ai.operation.name SHOULD be invoke_node.` to the top of the event's note, matching the wording on the `invoke_node` span. Pushed in 2488947.

reference/scenarios/google-adk/scenario.py
```diff
@@ -171,6 +171,32 @@ async def _run():
...
+                    "gen_ai.conversation.id": session.id,
+                }
+                with _reference_tracer.start_as_current_span(
+                    f"invoke_node {node_name}", attributes=node_span_attributes
+                ) as node_span:
+                    # Per the invoke_node convention, node inputs/outputs are
+                    # recorded verbatim (as the node's actual payload) rather
+                    # than being reshaped into the chat-message schema.
+                    node_span.set_attribute(
+                        "gen_ai.input.messages", json.dumps({"text": input_text})
```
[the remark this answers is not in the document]
RKest: Agreed on the input — wrapping `input_text` in `{"text": ...}` wasn't doing any work. Changed the input to `json.dumps(input_text)` so the verbatim guidance is demonstrated literally. Kept the output as `{"valid": True, "text": input_text}` because that genuinely is the verbatim payload a validation node would return; updated the comment to call out both halves explicitly. Pushed in 2488947.

reference/reports/run-step-span.md
```diff
@@ -0,0 +1,25 @@
...
+## Required
+
+| Attribute | Supporting Libraries |
+| --- | --- |
+| gen_ai.operation.name | [google-adk] |
+
+## Conditionally Required
+
+| Attribute | Supporting Libraries |
+| --- | --- |
```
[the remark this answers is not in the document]
RKest: Not changing this one — `error.type` is intentionally filtered from every reference report by `DISPLAY_HIDDEN_ATTRS = frozenset({"error.type"})` in `reference/src/semconv_genai/report.py:39`. The Python tooling does resolve `error.type` as conditionally required on the `invoke_node` spec (it inherits from the `attributes.gen_ai.error` group, same as `invoke_workflow.internal` and the `invoke_agent` spans), it's just deliberately hidden from the rendered status pages — none of the existing reports list it either. Adding it only here would make `invoke-node-span.md` inconsistent with `invoke-workflow-span.md`, `invoke-agent-internal-span.md`, etc. Happy to file a separate issue if revisiting that repo-wide policy seems worthwhile.

reference/scenarios/google-adk/scenario.py
```diff
@@ -171,6 +171,32 @@ async def _run():
...
+                    # Per the invoke_node convention, node inputs/outputs are
+                    # recorded verbatim — as the node's actual payload — rather
+                    # than being reshaped into the chat-message schema. Here the
+                    # node's input is a plain string (the raw user prompt) and
+                    # its output is a small validation result object.
+                    node_span.set_attribute("gen_ai.input.messages", json.dumps(input_text))
+                    node_span.set_attribute(
+                        "gen_ai.output.messages",
+                        json.dumps({"valid": True, "text": input_text}),
+                    )
```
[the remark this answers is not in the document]
RKest: Good point — `json.dumps` was double-encoding the string (the stored value would have been `"\"raw prompt\""`). Switched `gen_ai.input.messages` to `input_text` directly (plain `str` is a valid OTel attribute), kept the dict-shaped output `json.dumps`'d since attributes can't carry mappings, and updated the comment to explain both halves. Pushed in 6d2d27a.

reference/README.md
```diff
@@ -40,5 +41,6 @@ Run `uv run update-reports` to regenerate.
 | Event | Libraries |
 | --- | --- |
 | [Inference Operation Details](reports/gen-ai-client-inference-operation-details-event.md) | anthropic, autogen, aws-bedrock, azure-ai-inference, cohere, dspy, google-genai, groq, instructor, litellm, llamaindex, mistralai, openai, pydantic-ai, vertexai |
+| [Workflow Node Invocation Details](reports/gen-ai-client-workflow-node-invocation-details-event.md) | (none) |
```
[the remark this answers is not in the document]
RKest: Added emitting of the events, and added google-adk as the supported library. Pushed in 6d2d27a.

reference/scenarios/google-adk/scenario.py
```diff
@@ -171,6 +177,44 @@ async def _run():
...
+                    event_name="gen_ai.client.workflow.node.invocation.details",
+                    body="Workflow node invocation details",
+                    attributes={
+                        "gen_ai.operation.name": "invoke_node",
+                        "gen_ai.node.name": node_name,
+                        "gen_ai.conversation.id": session.id,
+                        "gen_ai.input.messages": node_input,
+                        "gen_ai.output.messages": node_output,
+                    },
+                )
```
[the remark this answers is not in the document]
RKest: Fixed. Pushed in cab28d1.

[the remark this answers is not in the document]
RKest: Added a carve out for when the input/output itself is not structured. Pushed in https://github.com/open-telemetry/semantic-conventions-genai/commit/cab28d1c8e2f76844208060a0d25a1604b20d5a5.

reference/scenarios/google-adk/scenario.py
```diff
@@ -171,6 +177,47 @@ async def _run():
...
+                # applications this is emitted automatically by
+                # google.adk.telemetry.node_tracing.start_as_current_node_span for
+                # any BaseNode that is neither a BaseAgent nor a Workflow (e.g.,
+                # deterministic input validation or routing nodes). The node name
+                # comes from the node definition; here we use a static name to
+                # represent that preparatory step.
+                node_name = "validate_input"
+                node_span_attributes = {
+                    "gen_ai.operation.name": "invoke_node",
+                    "gen_ai.node.name": node_name,
```
lmolkova: where would it come from in a real instrumentation?

model/gen-ai/spans.yaml
```diff
@@ -535,6 +535,64 @@ spans:
...
+    note: |
+      The `gen_ai.operation.name` SHOULD be `invoke_node`.
+
+      **Span name** SHOULD be `invoke_node {gen_ai.node.name}` if `gen_ai.node.name` is readily
+      available. When `gen_ai.node.name` is not available, it SHOULD be `invoke_node`.
+
+      A node represents any unit of work that participates in a workflow but is not itself a
+      GenAI agent. Typical examples include deterministic pre/post-processing steps,
+      control-flow nodes, routing or branching nodes, data-transformation nodes, and other
+      framework-defined building blocks that a workflow orchestrates
```
lmolkova: > pre/post-processing steps,
      control-flow nodes, routing or branching nodes, data-transformation nodes, and other
      framework-defined building blocks that a workflow orchestrates

all of these things are short/local and don't match what OTel considers to be a span. E.g. routing  and data transformations are short local operations that should be represented as events rather than spans. 

At the same time, nodes can wrap inference calls and other continuous operations and then it's meaningful to capture them as spans.

Can we define node better? One way to approach it would be to take existing frameworks with this notion and document what node means for each of them and in which cases node span should be captured (e.g. no node span for ADK agents)

model/gen-ai/spans.yaml
```diff
@@ -535,6 +535,64 @@ spans:
...
+      `gen_ai.agent.name` set, so that an `invoke_agent` span acts as the node in the workflow.
+
+      The term "node" is used in preference to "task" because "task" is overloaded across
+      the GenAI ecosystem (for example, ADK uses `mode="task"` for a specific agent mode, and
+      A2A defines a `Task` as a stateful, long-lived object). "Node" broadly covers anything
+      that participates in a workflow and is not an agent.
+
+      Instrumentations SHOULD report this span for every workflow unit of work they wrap,
+      unless they can reliably identify that unit of work as a GenAI agent, in which case
+      an `invoke_agent` span SHOULD be reported instead. In other words, `invoke_node` is
```
lmolkova: what about tool calls? we should not report node span that wraps tool execution if some framework considers it to be a node too. I think we should add a caveat that node spans should not be reported if we know that underlying call is instrumented on its own

reference/scenarios/google-adk/scenario.py
lmolkova, line 1: there are other reference instrumentations that would be affected: lanchain and possibly others. Can you please investigate which, what terminology they use, and update their scenarios?

model/gen-ai/events.yaml
```diff
@@ -10,6 +10,34 @@ events:
     stability: development
     attributes:
       - ref_group: span.gen_ai.inference.client
+  - name: gen_ai.client.workflow.node.invocation.details
```
lmolkova: we don't routinely create events for each span, let's postpone adding this one too

model/gen-ai/registry.yaml
```diff
@@ -500,6 +506,8 @@ attributes:
       When the attribute is recorded on events, it MUST be recorded in structured
       form. When recorded on spans, it MAY be recorded as a JSON string if structured
       format is not supported and SHOULD be recorded in structured form otherwise.
+      An exception applies when input/output itself is not structured, for example
```
lmolkova: this is placed under `gen_ai.retrieval.documents` - is this intentional ?

If the intent was to put it on input / output message attributes -  
we have JSON schema that needs to be updated, but it includes generic part that should probably be used here.

model/gen-ai/registry.yaml
```diff
@@ -572,6 +580,8 @@ attributes:
...
```
lmolkova: it seems to be misplaced, see my other comment

model/gen-ai/registry.yaml
```diff
@@ -644,14 +654,28 @@ attributes:
   - key: gen_ai.input.messages
     type: any
     brief: >
-      The chat history provided to the model as an input.
+      The input provided to a GenAI operation. Shape depends on the operation:
```
lmolkova: I don't think we should update the meaning of this attribute. It means something specific today and it loses this meaning when expanded like this

lmolkova (changes requested):

hippoley:
The `invoke_node` span fills a real gap — in multi-agent workflows the current span hierarchy jumps from `invoke_workflow` directly to `invoke_agent` or `execute_tool`, with no way to represent intermediate processing nodes (validators, routers, formatters, etc.).

The design is consistent with the existing `gen_ai.execute_tool.internal` / `gen_ai.plan.internal` pattern, which is good. A few observations:

1. **Naming**: `invoke_node` reads naturally alongside `invoke_agent` and `invoke_workflow`. The `gen_ai.node.name` attribute is clean.

2. **Relationship to `gen_ai.evaluation.internal`**: I have an open PR (#185) that adds a similar `gen_ai.evaluation.internal` span for evaluation steps. Both follow the same `gen_ai.<operation>.internal` pattern. It might be worth a quick cross-check to make sure the attribute sets are consistent (both use `gen_ai.operation.name` as required, `gen_ai.agent.name` as conditionally required, etc.).

3. **`gen_ai.client.workflow.node.invocation.details` event**: the name is quite long. Is there a shorter form that still conveys the scope? Something like `gen_ai.node.invocation.details` might be easier to type and remember.

Overall this is a solid addition. The Google ADK reference scenario is a good concrete anchor.

reference/scenarios/google-adk/scenario.py
RKest: The instrumentation in Google ADK is implemented here:
https://github.com/google/adk-python/blob/main/src/google/adk/telemetry/node_tracing.py

It instruments node runner here:
https://github.com/google/adk-python/blob/7ad7994744de18f2394e4bcb961cd5c7a24afb4b/src/google/adk/workflow/_node_runner.py#L125

To give a concrete example, the following ADK sample:
https://github.com/google/adk-python/blob/main/contributing/samples/workflows/loop/agent.py

Produces the following trace:
image

model/gen-ai/spans.yaml
RKest: I'm curious how come short/local operations should not emit spans?

To me it naively makes sense to reflect which node in a workflow is preceded/succeeded by another node.

I clarified the wording in the proposal, to make it clearer what the motivation is, but on the broader picture, could you give feedback on if the example in https://github.com/open-telemetry/semantic-conventions-genai/pull/188#discussion_r3303424874 makes sense to be instrumented in this way, or do you see a more sensible OTel primitive which would be a better fit?

I considered events on the `invoke_workflow` span, but this wouldn't capture parallel execution.

model/gen-ai/spans.yaml
RKest: Good point, clarified `invoke_node` should only be emitted, if the underlying operation is not already instrumented

reference/scenarios/google-adk/scenario.py
RKest: Added (actually vibe-coded) scenarios for different frameworks.

When it comes to the terminology, https://github.com/open-telemetry/semantic-conventions-genai/issues/187 outlines in motivation what terminology is used in other frameworks.

Tl;Dr: Langgraph uses `execute_task` (I outlined in this PR and https://github.com/open-telemetry/semantic-conventions-genai/issues/187 why this not ideal terminology) and CrewAI currently doesn't instrument what they coin as "flows"

model/gen-ai/events.yaml
RKest: Makes sense, removed this event

model/gen-ai/registry.yaml
RKest: Removed the event. Will address this comment in the future PR

model/gen-ai/registry.yaml
RKest: Removed the event. Will address this comment in the future PR

model/gen-ai/registry.yaml
RKest: Removed the event. Will address this comment in the future PR

trask:
hi @RKest, can you mark this as "Resolves #187" (assuming it does)? thanks!

RKest:
> hi @RKest, can you mark this as "Resolves #187" (assuming it does)? thanks!

Hi @trask, This PR resolves #187 only partly. I still would like to write a follow-up with an event to capture node inputs/outputs. I'll write it with #175 in mind🙏

model/gen-ai/spans.yaml
```diff
@@ -535,6 +535,68 @@ spans:
...
+      agent-as-tool tool calls). In that case the execution graph is already reconstructible
+      from the existing spans and an additional `invoke_node` span would add no information.
+
+      Semantic conventions for individual GenAI frameworks SHOULD document which framework
+      constructs map to `invoke_node` and which do not.
+
+      The term "node" is used in preference to "task" because "task" is overloaded across the
+      GenAI ecosystem (for example, A2A defines a `Task` as a stateful, long-lived object).
+      "Node" broadly covers any unit of work that participates in a workflow and is not
+      already represented by another GenAI span.
```
trask: IIUC, https://github.com/open-telemetry/semantic-conventions-genai/pull/203 proposes "step" for this same concept?

model/gen-ai/spans.yaml
trask: > I'm curious how come short/local operations should not emit spans?

they generally have a low signal-to-noise ratio compared to spans defined at remote boundary operations. It doesn't mean we can't add them, but it does mean it's a much higher bar. and I think it would help me understand the proposal better if you could add the user journey to the PR description. I see that you want to reconstruct the full workflow graph from a trace, but not why? Is it for replay? Are those local operations non-deterministic? A bit more in the motivation section would help. Thanks!

> Can we define node better? One way to approach it would be to take existing frameworks with this notion and document what node means for each of them and in which cases node span should be captured (e.g. no node span for ADK agents)

👍

similar to https://github.com/open-telemetry/semantic-conventions-genai/pull/203#discussion_r3329327287:

> It would be helpful if we can define a ["node"] more concretely, even if it's a medium sized list of things that it covers.
>
> If it has to be framework specific, we may want to think about adding `gen_ai.framework.name` and framework specific conventions to define it concretely for each of the common frameworks.
> 
> It feels a bit vague at this point which will lead to lots of questions when people try to implement it.

model/gen-ai/spans.yaml
```diff
@@ -535,6 +535,68 @@ spans:
...
+      The `gen_ai.operation.name` SHOULD be `invoke_node`.
+
+      **Span name** SHOULD be `invoke_node {gen_ai.node.name}` if `gen_ai.node.name` is readily
+      available. When `gen_ai.node.name` is not available, it SHOULD be `invoke_node`.
+
+      `invoke_node` exists so that the execution graph of a workflow can be reconstructed from
+      telemetry. It is especially meaningful in graph-based workflows, where the output of one
+      node is the input of the next and nodes may route control between agents, sub-workflows,
+      or other nodes; without an explicit span for those intermediate nodes, the graph cannot
+      be reconstructed from the surrounding spans alone.
```
aabmass: This is pretty interesting. Are we able to reconstruct the graph for haystack, langchain, and crewai as well?

One concern is this doesn't help in case nodes in the graph are skipped or not chosen in a routing step. I imagine that capturing the static definition of the graph as metadata would still be needed.

reference/scenarios/haystack/scenario.py
```diff
@@ -57,6 +57,96 @@ def run_retrieval():
...
+            normalized_query = " ".join(raw_query.lower().replace("?", "").split())
+
+        # Second pipeline step: retriever component. Already mapped to a
+        # `retrieval` span; MUST NOT be wrapped in `invoke_node` on top.
+        with _reference_tracer.start_as_current_span("retrieval weather-knowledge-base") as retrieval_span:
+            retrieval_span.set_attribute("gen_ai.operation.name", "retrieval")
+            retrieval_span.set_attribute("gen_ai.data_source.id", data_source_id)
+            retrieval_span.set_attribute("gen_ai.request.top_k", request_top_k)
+            retrieval_span.set_attribute("gen_ai.retrieval.query.text", normalized_query)
+            result = retriever.run(query=normalized_query)
```
lmolkova: why do we wrap retrieval span in node span? we already have a convention for retrieval and it directly contradicts node operation definition

> Invoke a unit of work in a GenAI workflow that is not already covered by another GenAI span


I think neither workflow nor node spans should be emitted by haystack instrumentation - it's not an agentic framework, so suggesting to remove all changes in this file

model/gen-ai/spans.yaml
```diff
@@ -535,6 +535,68 @@ spans:
...
 
+  - type: gen_ai.invoke_node.internal
+    kind: internal
+    name:
+      note: >
+        **Span name** SHOULD be `invoke_node {gen_ai.node.name}` if `gen_ai.node.name` is readily
+        available. When `gen_ai.node.name` is not available, it SHOULD be `invoke_node`.
+    brief: >
+      Represents a unit of work in a GenAI workflow that is not otherwise instrumented,
+      captured so that the workflow's execution graph can be reconstructed from telemetry.
```
lmolkova: it seems we also need to record edges in formal way? How would telemetry consumer know that execution moved from node A to node B if we just capture that happens inside each of these nodes? 

or do we expect an edge to also be captured as a node with details in the attributes? Representing those as spans would be problematic - they are extremely short and local operations.

Left a comment with possible solutions on the original issue:  https://github.com/open-telemetry/semantic-conventions-genai/issues/187#issuecomment-4654925213

model/gen-ai/spans.yaml
lmolkova: did some analysis in https://github.com/open-telemetry/semantic-conventions-genai/issues/187#issuecomment-4654375405 - I have a soft preference for node

model/gen-ai/spans.yaml
```diff
@@ -535,6 +535,68 @@ spans:
...
+        requirement_level: required
+        sampling_relevant: true
+      - ref: gen_ai.node.name
+        requirement_level:
+          conditionally_required: when available
+        sampling_relevant: true
+      - ref: gen_ai.conversation.id
+        requirement_level:
+          conditionally_required: when available
+      - ref: gen_ai.input.messages
```
lmolkova: I think we should not reuse existing attributes here, motivation in the https://github.com/open-telemetry/semantic-conventions-genai/issues/187#issuecomment-4654522301

trask:

Create `changelog.d/188.enhancement.md` containing the change log entry, e.g.:

```md
Add `invoke_node` operation, `gen_ai.node.name` attribute, and `gen_ai.invoke_node.internal` span to capture units of work in a GenAI workflow that are not already covered by another GenAI span (e.g. agent transfer, tool call, LLM call, retrieval, memory operation), so that the workflow's execution graph can be reconstructed from telemetry. `invoke_node` MUST NOT be emitted for operations already instrumented by another GenAI span, and SHOULD NOT be emitted when the framework already provides a well-defined agentic loop in which routing is expressed through already-instrumented operations.
```

Thanks!

model/gen-ai/spans.yaml
RKest: I expanded the motivation and provided concrete CUJs https://github.com/open-telemetry/semantic-conventions-genai/issues/187 to convey why the `invoke_node` should be a span. The answer to the questions is below:

Two questions bundled here: (a) span vs. event, (b) defining "node". They're
linked, so feasibility first.

**The instrumentation can't classify a node's internals.** When a framework
wraps a node, it doesn't know whether the body will call an LLM, spawn children,
run for 200ms or 20s, or throw. So "events for short ops, spans for continuous
ones" isn't implementable — the framework can't tell them apart before running
user code. The only reliable decision is **structural**: what is the vertex?

- Vertex is an already-instrumented primitive (agent/tool/retriever) registered
  directly → that op's span **is** the node. Emit `invoke_agent` etc., **not**
  `invoke_node` (no double-counting).
- Vertex hands control to an **opaque user node** → emit `invoke_node`. The
  framework can't see whether it does real work or just delegates, so it emits
  unconditionally.

ADK 2.x:

```py
# vertex IS an agent → invoke_agent only, no invoke_node
root_workflow = Workflow(
    graph=(START, some_agent)
)
```
```py
# vertex is an opaque user node → invoke_node (invoke_agent is its child)
@node
def some_node(ctx, input):
    ctx.run_node(some_agent, input)

root_workflow = Workflow(
    graph=(START, some_node)
)
```

The second node is a passthrough, but the instrumentation can't know that, and
these noop wrappers are rare: a node's output is normally consumed by the next
step, so nodes are intermediate by construction, not leaves (I/O continuity,
CUJ 3 in #187). Same contract as `execute_tool`: the span marks "user code the
workflow routed to" and is its attribution parent + error boundary, without
prescribing what runs inside.

**Why a span, not an event:**
- **Parenting** — an opaque node routinely contains other instrumented ops
  (LLM/tool/retrieval/sub-node) that need a span parent. Events can't parent.
- **Error attribution** — the node runs arbitrary, possibly-uninstrumented code
  that can throw; without its own span the exception lands on `invoke_workflow`
  and you lose which step failed (CUJ 2).
- **Concurrency** — ADK runs nodes concurrently; events on the parent collapse
  to timestamps and lose per-node duration/ordering/overlap.

**Concrete definition (per-framework):** emit `invoke_node` for a vertex that
hands control to user code not already represented by another GenAI span; never
emit where routing/control flow is already expressed through instrumented ops
(e.g. ADK agent-transfer tools / agent-as-tool).

| Framework | `invoke_node` | NOT (already a GenAI span) |
| --- | --- | --- |
| ADK 2.x | `@node` / non-`BaseAgent` vertices | `BaseAgent` registered directly → `invoke_agent` |
| LangGraph | `StateGraph` nodes of user code | model-calling node → `chat` |
| CrewAI Flow | `@start` / `@listen` steps | crew agents → `invoke_agent` |
| Haystack | non-LLM/non-retriever components | retriever → `retrieval` |

The ADK trace [above](https://github.com/open-telemetry/semantic-conventions-genai/pull/188#discussion_r3303424874)
is the anchor for row 1. If this framing works I'll tighten the span definition
to the structural rule and fold these tables into the doc.

model/gen-ai/spans.yaml
RKest: The previous name for this metric was a bit of a misnomer. It's updated to `gen_ai.agent.steps` (though the PR description should be updated as well, and we may narrow the name further, because a `step` still seems too broad to me).

The metric in any case is relatively unrelated to this proposal

model/gen-ai/spans.yaml
RKest: I narrowed the wording in the original issue and this PR to the traversed graph path.

I think the reconstruction of the entire graph is out of scope/not feasible because:
1. The graph may contain dynamic nodes, which perform routing with imperative code, which would not be reflected in the static graph structure
2. It would require standardizing how to represent for example "predicate nodes", which run based on some precondition of the previous node output. These probably could be represented purely with data, but I don't think enough frameworks support it, to be able to standardize it.

reference/scenarios/haystack/scenario.py
RKest: Removed the haystack scenario. Also further specified when the `invoke_node` span should be emitted in https://github.com/open-telemetry/semantic-conventions-genai/pull/188#discussion_r3414309342

model/gen-ai/spans.yaml
RKest: Great point, completely missed the nodes executed in parallel, would not be distinguishable from short sequential node spans.

I'll add the proposal in https://github.com/open-telemetry/semantic-conventions-genai/issues/187 but will follow up the exact wording in another PR.

My idea is probably each node execution should have a separate `gen_ai.node.execution.id`, because the same node can be executed in multiple places in a workflow

model/gen-ai/spans.yaml
RKest: Removed for now. I think will be worth while to revisit this at some point around https://github.com/open-telemetry/semantic-conventions-genai/issues/175

reference/reports/invoke-workflow-span.md
```diff
@@ -6,20 +6,22 @@
...
+| gen_ai.workflow.name | [crewai], [google-adk], [haystack], [langchain] |
 
 ## Opt-In
 
 | Attribute | Supporting Libraries |
 | --- | --- |
-| gen_ai.input.messages | [crewai], [google-adk] |
-| gen_ai.output.messages | [crewai], [google-adk] |
+| gen_ai.input.messages | [crewai], [google-adk], [haystack], [langchain] |
+| gen_ai.output.messages | [crewai], [google-adk], [langchain] |
```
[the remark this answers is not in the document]
RKest: Fixed. Regenerated the reference reports

reference/scenarios/google-adk/scenario.py
```diff
@@ -171,6 +176,24 @@ async def _run():
...
+                    "gen_ai.conversation.id": session.id,
+                }
+                with _reference_tracer.start_as_current_span(
+                    f"invoke_node {node_name}", attributes=node_span_attributes
+                ):
+                    pass
```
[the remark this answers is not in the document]
RKest: Done. The `invoke_node` span now runs the deterministic input-validation work inside the span.

model/gen-ai/spans.yaml
```diff
@@ -588,6 +588,58 @@ spans:
...
+    stability: development
+    attributes:
+      - ref_group: attributes.gen_ai.error
+      - ref: gen_ai.operation.name
+        requirement_level: required
+        sampling_relevant: true
+      - ref: gen_ai.node.name
+        requirement_level:
+          conditionally_required: when available
+        sampling_relevant: true
```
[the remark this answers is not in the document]
RKest: Done

model/gen-ai/spans.yaml
```diff
@@ -588,6 +588,58 @@ spans:
...
+      - ref: gen_ai.conversation.id
+        requirement_level:
+          conditionally_required: when available
+
```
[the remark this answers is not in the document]
RKest: Done

model/gen-ai/registry.yaml
```diff
@@ -855,3 +859,23 @@ attributes:
...
+  - key: gen_ai.node.name
+    type: string
+    brief: Human-readable name of a node within a GenAI workflow.
+    note: |
+      A node is a vertex of a GenAI workflow graph that hands control to application- or
+      framework-defined code that is not itself already covered by another GenAI span (e.g.
+      not an `invoke_agent`, `execute_tool`, `chat`, `text_completion`, `embeddings`,
+      `retrieval`, `plan`, or memory operation). See the `gen_ai.invoke_node.internal` span
+      definition for the full rules on when an `invoke_node` span (and therefore this
+      attribute) should be reported.
```
```suggestion
    A node is a vertex in a GenAI workflow graph that hands control to application-
    or framework-defined code. Known operations (e.g., `invoke_workflow`,
    `execute_tool`, inference) use specific attributes like `gen_ai.workflow.name`
    to describe the node.
    
    This attribute along with the corresponding `invoke_node` operation SHOULD NOT
    be used for known operations. They SHOULD act only as a fallback for unknown
    node types.
```

model/gen-ai/registry.yaml
```diff
@@ -855,3 +859,23 @@ attributes:
...
+
+      `gen_ai.node.name` MUST have low cardinality. It is usually a static name assigned by the
```
```suggestion
      `gen_ai.node.name` SHOULD have low cardinality. It is usually a static name assigned by the
```

there is no way to enforce or guarantee it on the instrumentation side, one can always write a very special application that would have hundreds of functions annotated as nodes.

model/gen-ai/spans.yaml
```diff
@@ -588,6 +588,58 @@ spans:
...
 
+  - type: gen_ai.invoke_node.internal
+    kind: internal
+    name:
+      note: >
+        **Span name** SHOULD be `invoke_node {gen_ai.node.name}` if `gen_ai.node.name` is readily
+        available. When `gen_ai.node.name` is not available, it SHOULD be `invoke_node`.
+    brief: >
+      Represents a unit of work in a GenAI workflow that is not otherwise instrumented,
+      captured so that the workflow's execution graph can be reconstructed from telemetry.
```
lmolkova: nit:

```suggestion
      Represents a unit of work in a GenAI workflow that is not one of the known operations 
      such as `invoke_workflow`, `invoke_agent`, inference or another documented operation.
```

a little bit more precise

model/gen-ai/spans.yaml
```diff
@@ -588,6 +588,58 @@ spans:
...
+
+      **Span name** SHOULD be `invoke_node {gen_ai.node.name}` if `gen_ai.node.name` is readily
+      available, and `invoke_node` otherwise.
+
+      A node is a vertex of a workflow graph that hands control to application- or
+      framework-defined code not already represented by another GenAI span (e.g. not an
+      `invoke_agent`, `execute_tool`, `chat`, `text_completion`, `embeddings`, `retrieval`,
+      `plan`, or memory operation). Capturing it as a span lets the workflow's traversed
+      execution graph be reconstructed from telemetry, errors be attributed to the step that
+      raised them, and nested or concurrent operations be parented correctly.
```
lmolkova: nit

```suggestion
      A node is a vertex of a workflow graph that hands control to application- or
      framework-defined code not already represented by another GenAI span (e.g. not an
      `invoke_agent`, `execute_tool`, `chat`, `text_completion`, `embeddings`, `retrieval`,
      `plan`, or memory operation). 
```

removing justification from the spec

model/gen-ai/spans.yaml
```diff
@@ -588,6 +588,58 @@ spans:
...
+        vertex, that primitive's own span is the node; wrapping it would double-count.
+      - It SHOULD NOT report `invoke_node` where the framework already expresses routing and
+        control flow through instrumented operations (e.g. agent-transfer or agent-as-tool
+        calls), since the graph is then already reconstructible.
+
+      Semantic conventions for individual GenAI frameworks SHOULD document which framework
+      constructs map to `invoke_node` and which do not.
+
+      "Node" is preferred over "task", which is overloaded across the GenAI ecosystem (e.g.
+      A2A's `Task` is a stateful, long-lived object).
```
```suggestion
```

model/gen-ai/spans.yaml
```diff
@@ -588,6 +588,58 @@ spans:
       - ref: gen_ai.output.messages
         requirement_level: opt_in
 
+  - type: gen_ai.invoke_node.internal
```
lmolkova: do we also need a metric for it?

lmolkova (commented):
Looks good, just a few comments

changelog.d/188.enhancement.md
```diff
@@ -0,0 +1 @@
+Add `invoke_node` operation, `gen_ai.node.name` attribute, and `gen_ai.invoke_node.internal` span to capture units of work in a GenAI workflow that are not already covered by another GenAI span (e.g. agent transfer, tool call, LLM call, retrieval, memory operation), so that the workflow's execution graph can be reconstructed from telemetry. `invoke_node` MUST NOT be emitted for operations already instrumented by another GenAI span, and SHOULD NOT be emitted when the framework already provides a well-defined agentic loop in which routing is expressed through already-instrumented operations.
```
lmolkova: nit:

```suggestion
Add `invoke_node` operation, `gen_ai.node.name` attribute, and `gen_ai.invoke_node.internal` span to capture units of work in a GenAI workflow that are not already covered by another GenAI span (e.g. agent transfer, tool call, LLM call, retrieval, memory operation), so that the workflow's execution graph can be reconstructed from telemetry. `invoke_node` MUST NOT be emitted for operations already instrumented by another GenAI span.
```

note sure I understand what ", and SHOULD NOT be emitted when the framework already provides a well-defined agentic loop in which routing is expressed through already-instrumented operations." means and it seems it does not add much on top

lmolkova (commented):
I support the PR - it helps us cover operations that are not categorized yet.

Now bikeshed time: *node* vs *step* vs *task* 

## Agent frameworks

| lib | native term | concerns with "node" | concerns with "step" | concerns with "task" |
|---|---|---|---|---|
| langchain (LangGraph) | node | 🟢 none — native | foreign, no collision | overloaded (generic) |
| google-adk 2.0 | node + edge | 🟢 none — native | 🟢 none — ADK calls a node "a single step" | overloaded (generic) |
| pydantic-ai (pydantic-graph) | node (BaseNode) | 🟢 none — native | foreign, no collision | overloaded (generic) |
| autogen (GraphFlow) | node (= agent) | 🟢 none — native | foreign | wrong level — task = the input/overall job |
| agent-framework (MAF) | executor + edge | foreign | foreign | overloaded (generic) |
| haystack (Pipeline) | component | 🔴 deprecated v1 term | foreign | overloaded (generic) |
| llamaindex (Workflows) | step (@step) | 🔴 collides with Node/TextNode | 🟢 none — native | 🔴 collides with Task/TaskStep (core agent API) |
| crewai | step (Flow); Task (Crew) | foreign | 🟢 none — native (Flow) | 🔴 collides with Task (core class) |
| dspy | module (forward) | quite foreign | foreign | overloaded (generic) |
| openai-agents | (none — handoff) | n/a | n/a | overloaded (generic) |
| aws-bedrock-agent | action group | foreign | "step" in trace output | overloaded (generic) |

node: 4🟢/2🔴 · step: 3🟢/0🔴 · task: 0🟢/2🔴 + overloaded everywhere

## Eval libs

| eval lib | native term | concerns with "node" | concerns with "step" | concerns with "task" |
|---|---|---|---|---|
| LangSmith / AgentEvals | step | foreign | 🟢 none — native | 🔴 wrong level — task completion = whole goal |
| AgentEvals graph evaluator | step (field); node names as values | field is `steps`, not node | 🟢 none — native | 🔴 wrong level |
| DeepEval | step | foreign | 🟢 none — native | 🔴 wrong level — task completion metric |
| Strands TrajectoryEvaluator | step / action | foreign | 🟢 none — native | 🔴 wrong level |
| Langfuse | step / trajectory | foreign | 🟢 none — native | 🔴 wrong level — task completion eval |
| Arize / Phoenix | step (eval) / graph.node.name (instr.) | 🟢 native on instr. side | 🟢 native on eval side | 🔴 wrong level |

step: 6🟢/0🔴 · node: native only on Arize instr. · task: 0🟢/6🔴 (always the whole goal, never a unit)

------------
I'm proposing this:

- use `step` and `run_step` in span type and metrics
- use lib-specific terminology in the value of `gen_ai.operation.name` attribute -  we already do it for inference (chat, generate_content, text_completion)

model/gen-ai/registry.yaml
RKest: Applied with the run-step rename

model/gen-ai/registry.yaml
RKest: Done

model/gen-ai/spans.yaml
RKest: Done, applied in editor

model/gen-ai/spans.yaml
RKest: Done, applied in editor

model/gen-ai/spans.yaml
RKest: Done, applied in editor, with the `run_step` span name change

model/gen-ai/spans.yaml
RKest: Added a `recommended` duration metric to be consistent with the `invoke_node` and `invoke_workflow` spans

changelog.d/188.enhancement.md
RKest: Removed this wording. I added this wording to solve an ambiguity issue, but now that I think about it, it doesn't solve it at all. 

I'll leave my thinking process for posterity:
<details>

I wanted to distinguish ADK 1.* agentic loop, from ADK 2.* workflows:

The worry is that in ADK 1.* entrypoint, the following trace could represent a conversation with an agent transfer.
```
invoke_workflow
├── invoke_agent root
└── invoke_agent subagent
```

In that case given:
- User inputs "user message"
- Root agent outputs "root agent response"

Subagent would see the entire conversation i.e both "user message" and "root agent response" as its input.

Reason being this is a conversational agent, and `root` only hands control to `subagent`, because it believes `subagent` to be the right agent to handle user request.

OTOH in ADK 2.* the same trace as above can represent a graph-based workflow.

In that case, the subagent would only see only the "root agent response" as its input, given it follows the workflow edge.

I was thinking presence of `invoke_node` spans alongside `invoke_agent` nodes, would signify this is the 2.* case, but needless to say, it's not the best heuristic. I'll include a solution in the followup PR.

</details>

RKest:
See also: https://github.com/open-telemetry/semantic-conventions-conformance/pull/27

model/gen-ai/registry.yaml
```diff
@@ -951,3 +955,27 @@ attributes:
       what `gen_ai.workflow.name` means in the context of that framework.
     examples: ["multi_agent_rag", "customer_support_pipeline"]
     stability: development
+  - key: gen_ai.step.name
+    type: string
+    brief: Human-readable name of a step within a GenAI workflow.
```
```suggestion
    brief: Human-readable name of a step within a GenAI workflow or agent invocation.
```

lmolkova (commented):
@RKest  WDYT about the following modification to this proposal https://github.com/open-telemetry/semantic-conventions-genai/compare/main...lmolkova:semantic-conventions-genai:generic-step

It's pretty much the same, but a bit more radical - instead of defining a step, we're defining generic operation of unknown nature with fallback to `run_step`.

- Generic span (`gen_ai.operation.internal`):
  - `gen_ai.operation.name` is a static, low-cardinality step name - like node name (fetch_date, before_tool_call, wait_for_human_input).
  - no `gen_ai.step.name` - it becomes operation name. `run_step` is a fallback
-  `gen_ai.operation.duration` metric covers *ALL* internal GenAI operations. It's opt-in (debatable) with a few but wide default buckets to stay frugal.

Pros: 
- A bit more reliable graduation story: when operations specialize into dedicated span types later we can keep reporting the same metric. Span type change is not breaking today - it's not on OTLP yet, but it will be in the future; at least span does not need to change shape.
- Avoids adding generic step name attribute that would become problematic when operation specializes (e.g. guardrails today is a generic step, step.name for it is something that would be called guardrail.name in the future). 
- To define what comes into `step.name`, we need to specialize operation in some way

Cons:

- Can't group metric by `operation.name = run_step` - operation name is an open set. I don't feel it's a deal breaker though - how frequently users need to see the duration of all unspecialized operations?, also can query with != filter for known ones.


For ADK nodes, this would mean `{node name}` becomes operation name. Alternative is to invent `gen_ai.node.name` (would work for other graph-based frameworks) and when you know generic operation is a node invocation, use it along with `gen_ai.node.name` - this is allowed by the proposal.

RKest:
Thanks for the feedback @lmolkova, some thoughs:

> gen_ai.operation.duration metric covers ALL internal GenAI operations. It's opt-in (debatable) with a few but wide default buckets to stay frugal.

I worry this would run into cardinality issues. We can already squeeze very little extra cardinality to metrics, when an agent name attribute is present. Adding tool names, model names and step names to the mix, will probably make the situation untenable.

> A bit more reliable graduation story: when operations specialize into dedicated span types later we can keep reporting the same metric. Span type change is not breaking today - it's not on OTLP yet, but it will be in the future; at least span does not need to change shape.

I think most steps will be application-specific, and so they won't have a graduation story. E.g. a coding workflow, would have a `code_review` step, an e-commerce workflow would have a `detect_fraud` step.

But, if such a graduation approach, telemetry consumers would have to migrate to the specialized attribute, if they're only interested in guardrails. That's a real con.

Two ideas how to mitigate this from my side:
1. Leave out `gen_ai.operation.name` from `run_step` span, and discourage relying on the first half of the span name.
2. We could put this very wording in the semconv that "If the step is something application specific, the span name SHOULD be `run_step {step.name}`, otherwise if the step is something "gen_ai systems" specific, the name can be something more specific, for example `run_guardrail {guardrail.name}`"

PR #197, opened 2026-05-26
Add modality, cache, and phase breakdowns for token usage

### LINKED ISSUE #23: More detailed token usage span attributes and metrics
### Area(s)

area:gen-ai

### What's missing?

Models tend to have more than just 'input' and 'output' token types. The exact type matters because they have different prices.

In [Anthropic](https://docs.anthropic.com/en/api/messages#response-usage-cache-creation-input-tokens), there's `cache_creation_input_tokens` and `cache_read_input_tokens`.

[Google](https://ai.google.dev/api/generate-content#UsageMetadata) has this:

```
{
  "promptTokenCount": integer,
  "cachedContentTokenCount": integer,
  "candidatesTokenCount": integer,
  "toolUsePromptTokenCount": integer,
  "totalTokenCount": integer,
  "promptTokensDetails": [
    {
      object (ModalityTokenCount)
    }
  ],
  "cacheTokensDetails": [
    {
      object (ModalityTokenCount)
    }
  ],
  "candidatesTokensDetails": [
    {
      object (ModalityTokenCount)
    }
  ],
  "toolUsePromptTokensDetails": [
    {
      object (ModalityTokenCount)
    }
  ]
}
```

where `ModalityTokenCount` is

```
{
  "modality": enum (Modality),
  "tokenCount": integer
}
```

and `Modality` is `TEXT`, `IMAGE`, etc. Some of the token counts in the main object are just the standard input/output, and some are totals of others.

There should be optional attributes in both spans and metrics with these details. Consistency between AI providers is probably not very important, but conventions for each provider are important for backends to do things like calculate costs.

Possible span attribute names for e.g. Anthropic's `cache_creation_input_tokens`:

```
gen_ai.usage.cache_creation_input_tokens
gen_ai.usage.detailed.cache_creation_input_tokens
gen_ai.usage.anthropic.cache_creation_input_tokens
```

Google's additional dimensions make it more complicated:

```
gen_ai.usage.cachedContentTokenCount
gen_ai.usage.cached_content_token_count
gen_ai.usage.cached_content_token_count.text
gen_ai.usage.text.cached_content_token_count
gen_ai.usage.text
gen_ai.usage.modality.text
```

(I'm assuming that the `chat` operation can still involve things like image tokens and thus modality matters)

For metrics, the `gen_ai.token.type` attribute might be enough to describe the difference between 'normal' tokens and other types like 'cached'. But a separate metric name for other types (not one metric name per type) would help prevent people from adding up token types that mustn't be added, since e.g. cached tokens are a subset of input tokens. For modalities, I suggest an attribute like `gen_ai.token.modality`.

### LINKED ISSUE #76: GenAI: Add metrics for detailed token usage (cache, reasoning)

trask:
Resolves #23 
Resolves #76
Resolves #96 

## Description

Restructures GenAI token usage attributes and metrics so consumers can answer "where did the tokens go?" — by modality, by cache status, and by generation phase (response vs. reasoning) — from a single, uniform set of conventions.

**Span attributes (registry)**

- Added per-modality attributes shaped to match the existing `gen_ai.usage.cache_read.input_tokens` / `cache_write.input_tokens` / `reasoning.output_tokens` form: `gen_ai.usage.{text,image,audio,video}.input_tokens` and `gen_ai.usage.{text,image,audio}.output_tokens`.
- Added per-modality cache-read attributes `gen_ai.usage.{text,image,audio,video}.cache_read.input_tokens` for providers that report cached-token counts broken down by modality.

**Metrics**

- Replaced the `gen_ai.client.token.usage` histogram with counters that carry the breakdown dimensions:
  - `gen_ai.client.inference.tokens` (by `gen_ai.token.type` and `gen_ai.token.modality`)
  - `gen_ai.client.inference.input_tokens_by_cache` (by `gen_ai.token.cache` and `gen_ai.token.modality`)
  - `gen_ai.client.inference.output_tokens_by_phase` (by `gen_ai.token.phase`)
- Added opt-in per-operation histograms: `gen_ai.client.inference.operation.{tokens,input_tokens,output_tokens}`.

**New enum attributes**

- `gen_ai.token.modality`: `text` / `image` / `audio` / `video` / `document` / `unknown`
- `gen_ai.token.cache`: `uncached` / `read` / `write`
- `gen_ai.token.phase`: `response` / `reasoning`

## Motivation

**User journey — cost allocation and capacity planning.** Operators of GenAI applications need to attribute spend and forecast capacity by the dimensions that actually drive cost: modality (image/audio tokens often cost more than text), cache hits (cached input is billed at a discount), and reasoning output (priced separately from response output on o-series and Gemini 2.5+). Today these dimensions either don't exist in the conventions or are expressed as separate attributes that aggregations can't pivot on.

**Prior art.** All three major providers already report these breakdowns:

- OpenAI: `prompt_tokens_details.cached_tokens`, `completion_tokens_details.reasoning_tokens`.
- Anthropic: `cache_read_input_tokens`, `cache_creation_input_tokens` (the latter is captured as `gen_ai.usage.cache_write.input_tokens`).
- Google Gemini: `usageMetadata.cachedContentTokenCount`, `thoughtsTokenCount`, per-modality `promptTokensDetails` / `candidatesTokensDetails`, and per-modality `cacheTokensDetails`.

Folding these into a unified dimensional model lets a single dashboard query answer cost questions across providers.

## Prototype

Reference scenarios under `reference/scenarios/` exercise the new attributes end-to-end:

| Attribute                                  | Reference scenarios emitting it |
| ------------------------------------------ | ------------------------------- |
| `gen_ai.usage.cache_read.input_tokens`     | anthropic, openai               |
| `gen_ai.usage.cache_write.input_tokens`    | anthropic                       |
| `gen_ai.usage.reasoning.output_tokens`     | google-genai, openai            |
| `gen_ai.usage.text.input_tokens`           | google-genai, openai            |
| `gen_ai.usage.text.output_tokens`          | google-genai, openai            |
| `gen_ai.usage.image.input_tokens`          | google-genai                    |
| `gen_ai.usage.image.output_tokens`         | openai                          |
| `gen_ai.usage.audio.input_tokens`          | openai                          |
| `gen_ai.usage.audio.output_tokens`         | openai                          |
| `gen_ai.usage.video.input_tokens`          | google-genai                    |
| `gen_ai.usage.text.cache_read.input_tokens`  | google-genai                  |
| `gen_ai.usage.image.cache_read.input_tokens` | google-genai                  |
| `gen_ai.usage.video.cache_read.input_tokens` | google-genai                  |

### Why remove the `gen_ai.client.token.usage` histogram

`gen_ai.client.token.usage` records two observations per call, one with `gen_ai.token.type=input` and one with `=output`. Aggregating it without filtering on `gen_ai.token.type` doesn't give you "p95 total tokens per operation" — it gives you p95 of the **union** of all input observations and all output observations, which isn't a meaningful quantity.

Adding modality would make this worse: every call would contribute even more partition slices, and the unfiltered distribution would be even less interpretable.

Replacing it with a counter (`gen_ai.client.inference.tokens`) keeps the breakdown — sum across `gen_ai.token.type` (or `gen_ai.token.modality`) equals total tokens — without pretending each partition slice is its own distribution sample. The per-call distribution use case is preserved by the new opt-in `gen_ai.client.inference.operation.tokens` histogram, which records a single observation per call equal to `input + output` so its distribution actually answers "how big is a typical operation?". And the opt-in `...operation.input_tokens` / `...operation.output_tokens` variants keep the directional distributions.

### Why do we need the `by_cache` and `by_phase` variants of the primary counter?

A counter's dimensions need to partition its total: each token gets exactly one combination of dimension values, and summing across any dimension returns the same total. Putting cache or phase on `gen_ai.client.inference.tokens` would require every producer to know the cache state or phase of every token broken down by modality, but many providers report each axis on its own (e.g. OpenAI Chat reports cached token total and modality totals separately, with no joint). Forcing a single-counter encoding would either require a `cache=unknown` value that conflates "no info" with "definitely uncached," or would lose modality on the cache cells.

Splitting cache and phase into their own counters (`...input_tokens_by_cache`, `...output_tokens_by_phase`) lets each axis be reported against the same total it actually partitions, and each counter stays aggregatable. `input_tokens_by_cache` is also dimensioned by `gen_ai.token.modality` so providers that do report the joint (e.g. OpenAI Realtime, Gemini) can emit per-`(cache, modality)` cells — making cost queries that match the provider pricing structure (per-modality, per-cache-state unit prices) answerable from metrics alone. Providers that don't report the joint set `modality=unknown` on the cache cells.

(Strictly speaking, reasoning tokens are always text today across every provider, so `phase` could currently fold into the primary counter without breaking the partition. It's kept separate only for symmetry with `cache`, and to leave room for future non-text reasoning.)

### Why modality `unknown`?

So that summing `gen_ai.client.inference.tokens` across `gen_ai.token.modality` always equals the total tokens, even against providers that don't report a modality breakdown.

### Why not cache `unknown` or phase `unknown`?

`cache=uncached` and `phase=response` already act as the default bucket: if a provider doesn't report a cache breakdown, all input tokens are `cache=uncached`; if it doesn't report a reasoning breakdown, all output tokens are `phase=response`. So summing across `gen_ai.token.cache` or `gen_ai.token.phase` still equals the total without needing an `unknown` value.

### Why don't `gen_ai.token.modality` and `gen_ai.output.type` use the same enum values?

The two enums share `text` and `image`. The only real conflict is `audio` (modality) vs `speech` (output.type), and that's intentional. `gen_ai.token.modality=audio` describes the medium of a token and covers any acoustic content — speech, music, sound effects. `gen_ai.output.type=speech` describes the requested output format and is specifically spoken language (TTS-style output).

### Why remove `gen_ai.client.token.usage` instead of deprecating it?

The initial release out of this dedicated repository removes deprecated metrics and attributes as a one-time cutover under a new schema URL.

### Why no `gen_ai.usage.video.output_tokens`?

No major video-generation API exposes a token count today. OpenAI's Videos API (Sora 2 / Sora 2 Pro) returns only `{status, model, seconds, size, progress}` and bills per second of output. Google's Veo and similar models follow the same per-second billing model. With no source at a public call boundary, this attribute has nothing to populate, so it was dropped per the reference-evaluation guidelines.

### Why doesn't `output_tokens_by_phase` carry `gen_ai.token.modality`?

`input_tokens_by_cache` carries modality because cached input spans modalities — Gemini reports per-modality cached tokens, so each `(cache, modality)` combination has real data behind it. Reasoning output is text-only across every provider today.

### Why no per-modality `cache_write` attributes?

Per-modality `cache_read` exists because Gemini reports cached tokens broken down by modality (`cacheTokensDetails`). No provider reports a per-modality cache-write split — Anthropic's `cache_creation_input_tokens` and Gemini's `cached_content_token_count` are both aggregates — so there's nothing to populate per-modality cache-write attributes with today.

### Should `gen_ai.usage.*.{input,output}_tokens` be flattened to `gen_ai.usage.*_{input,output}_tokens`?

Possible renames (including existing attributes):

- `gen_ai.usage.cache_read.input_tokens` → `gen_ai.usage.cache_read_input_tokens`
- `gen_ai.usage.cache_write.input_tokens` → `gen_ai.usage.cache_write_input_tokens`
- `gen_ai.usage.reasoning.output_tokens` → `gen_ai.usage.reasoning_output_tokens`
- `gen_ai.usage.{text,image,audio,video}.input_tokens` → `gen_ai.usage.{text,image,audio,video}_input_tokens`
- `gen_ai.usage.{text,image,audio,video}.cache_read.input_tokens` → `gen_ai.usage.{text,image,audio,video}_cache_read_input_tokens`
- `gen_ai.usage.{text,image,audio,video}.output_tokens` → `gen_ai.usage.{text,image,audio,video}_output_tokens`

Currently, the new per-modality attributes follow the existing shape (`gen_ai.usage.cache_read.input_tokens` etc.). The OTel naming spec suggests dots are mainly for grouping properties under a shared object, which arguably doesn't apply here since `cache_read` and `text` read as modifiers rather than parent objects — so a flat snake_case form may be a closer fit.

## Alternatives considered

<details>
<summary>Click to expand</summary>

### 1. Complex `details` attribute with mixed-axis `variant`

Span:

- `gen_ai.usage.input_token.count`
- `gen_ai.usage.input_token.details` — complex, e.g. `{ "modality": { "text": { "count": 100, "variant": { "uncached": 70, "cache_read": 30 } } } }`
- `gen_ai.usage.output_token.count`
- `gen_ai.usage.output_token.details` — complex, e.g. `{ "modality": { "text": { "count": 180, "variant": { "response": 150, "reasoning": 30 } } } }`

Counter `gen_ai.client.inference.tokens` dimensions:

- `gen_ai.token.type` ∈ `input` / `output`
- `gen_ai.token.modality` ∈ `text` / `image` / `audio` / `video` / `document`
- `gen_ai.token.variant` ∈ `uncached` / `cache_read` / `cache_write` / `response` / `reasoning`

Discarded because providers don't report the modality × variant cross product, so the structured shape had nothing to populate; and because `variant` wasn't really one axis. `cache_read` / `cache_write` / `uncached` describe how an input token was sourced; `response` / `reasoning` describe what the model produced an output token for. They have no common semantic — the only thing tying them together is that each value happens to apply to exactly one direction. Putting them under one attribute name means:

- The attribute can't be queried on its own. `sum by (variant)` produces a meaningless mix of "tokens that were cached" and "tokens that were reasoning"; any real query has to filter by `gen_ai.token.type` first, so the consumer has to know which values pair with which type out of band.
- The enum has no stable meaning. Documenting `variant` requires saying "these values mean X when type=input and Y when type=output," which is just two enums hiding under one name.
- Any new direction-specific axis would have to be crammed into the same enum.

Two separate, direction-scoped attributes (`gen_ai.token.cache` on input, `gen_ai.token.phase` on output) avoid all of this — each is one axis with one meaning that does partition its direction's total.

### 2. Complex `details` with per-direction `cache` / `phase`

Span:

- `gen_ai.usage.input_token.count`
- `gen_ai.usage.input_token.details` — complex, e.g. `{ "modality": { "text": { "count": 100, "cache": { "uncached": 70, "read": 30 } } } }`
- `gen_ai.usage.output_token.count`
- `gen_ai.usage.output_token.details` — complex, e.g. `{ "modality": { "text": { "count": 180, "phase": { "response": 150, "reasoning": 30 } } } }`

Counter `gen_ai.client.inference.tokens` dimensions:

- `gen_ai.token.type` ∈ `input` / `output`
- `gen_ai.token.modality` ∈ `text` / `image` / `audio` / `video` / `document`
- `gen_ai.input_token.cache` ∈ `uncached` / `read` / `write` (when `type=input`)
- `gen_ai.output_token.phase` ∈ `response` / `reasoning` (when `type=output`)

Discarded for the same reason as #1: still asks producers for a cross product (modality × cache, modality × phase) that no major provider reports.

### 3. Complex `details` as an array of joint cells

Span:

- `gen_ai.usage.input_token.count`
- `gen_ai.usage.input_token.details` — complex array, each cell carrying whichever axes the producer has joint information for, e.g. `[ { "modality": "text", "count": 90 }, { "modality": "audio", "count": 10 }, { "cache": "read", "count": 40 } ]`
- `gen_ai.usage.output_token.count`
- `gen_ai.usage.output_token.details` — same shape, with `phase` in place of `cache`.

Cells are disjoint along whatever axes are present and sum to the scalar `count`; an omitted axis on a cell means "aggregated across that axis."

Sidesteps the cross-product problem in #1 / #2 — producers without joint information just emit one cell per known marginal axis with the other axes omitted.

This is an attractive option: it models current state and extends naturally once providers emit cross products. The downside is that the complex structure doesn't flatten like the options above, which makes it harder for users to query in backends.

### 4. `gen_ai.usage.input_token.*`

- `gen_ai.usage.input_token.count`
- `gen_ai.usage.input_token.modality.<modality>`
- `gen_ai.usage.input_token.caching.<caching>`
- `gen_ai.usage.output_token.count`
- `gen_ai.usage.output_token.modality.<modality>`
- `gen_ai.usage.output_token.phase.<phase>`

I kind of wanted to put `.count` at the end of each, but that doesn't work with parameterized attribute names.

### 5. `gen_ai.usage.input_token.*`, alt

- `gen_ai.usage.input_token.count`
- `gen_ai.usage.input_token.{text,image,audio,video}.count`
- `gen_ai.usage.input_token.{cache_read,cache_write}.count`
- `gen_ai.usage.output_token.count`
- `gen_ai.usage.output_token.{text,image,audio,video}.count`
- `gen_ai.usage.output_token.reasoning.count`

`.count` feels redundant under the `.usage` node.

### 6. `gen_ai.input_token.*`

- `gen_ai.input_token.count`
- `gen_ai.input_token.{text,image,audio,video}.count`
- `gen_ai.input_token.{cache_read,cache_write}.count`
- `gen_ai.output_token.count`
- `gen_ai.output_token.{text,image,audio,video}.count`
- `gen_ai.output_token.reasoning.count`

</details>


Nik-Reddy:
@trask Read through the full PR including the FAQ and alternatives.
 Three questions on the choices made, framed against the three points I was planning to surface at SIG this week before #197 landed:
 
 **Three counters vs one with all axes.** The FAQ explains the cross-product issue cleanly: providers report cache and phase marginally, not jointly with modality, so a single counter with  all four dimensions would force producers to invent cross-product cells they don't actually have. Accepting that. One follow-up: the three-counter split means `sum(inference.tokens)` and `sum(input_tokens_by_cache)` should reconcile on input total, and consumers will likely write queries that subtract or compare across them. Is it worth documenting the invariant explicitly in the metric description so downstream dashboards can rely on it (e.g. "sum over `input_tokens_by_cache.cache` equals sum over inference.tokens where token.type=input")?
 
 **Modality value set consistency.** `gen_ai.token.modality` uses  `text | image | audio | video | document | unknown`. The existing `gen_ai.output.type` uses `text | json | image | speech` and the message-part schemas use `image | video | audio | document`. Two specific gaps: `audio` here vs `speech` on `output.type`, and `document` here but not on `output.type`. If the intent is for these enums to be consumable together in queries, worth aligning - or explicitly documenting that they intentionally diverge because they describe different things (token-level vs response-level).
 
 **Histogram removal vs deprecation window.** PR replaces `gen_ai.client.token.usage` rather than deprecating it. Anyone aggregating on the old histogram today (including the in-flight provider instrumentations I'm shipping under opentelemetry-python-genai https://github.com/open-telemetry/opentelemetry-python-genai/issues/56 ) breaks at the cut. Has the SIG considered a one-release deprecation window where both emit, so downstream instrumentations can migrate without a hard cut? If the answer is "no, the histogram was broken-by-design and a hard cut is correct" - happy to take that, just want it on the record.
 
 Will hold formal review comments until this leaves draft.

trask:
> the three-counter split means sum(inference.tokens) and sum(input_tokens_by_cache) should reconcile on input total

Added this explicitly in https://github.com/open-telemetry/semantic-conventions-genai/pull/197/commits/e5f93bab100809f275de3db555ef2748bd5fffef

> Modality value set consistency.

Added to FAQ in PR description.

> Histogram removal vs deprecation window.

Added to FAQ in PR description.

reference/src/semconv_genai/semconv_model.py
```diff
@@ -215,7 +215,7 @@ def _from_yaml(
 }
 
 # No METRIC_SPECS yet: gen_ai.* metrics are deliberately omitted from the reference
-# coverage matrix because the two non-streaming metrics
-# (`gen_ai.client.operation.duration`, `gen_ai.client.token.usage`) are
+# coverage matrix because the non-streaming metrics
+# (`gen_ai.client.operation.duration`, `gen_ai.client.inference.tokens`) are
 # derivable from data already present on inference spans (span start/end and
```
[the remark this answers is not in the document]
trask: Done in bb54363 — made the comment generic since the underlying rationale (derivable from existing span attributes) applies to all the new metrics as well.

reference/scenarios/google-genai/scenario.py
```diff
@@ -90,6 +90,208 @@ def run_chat():
...
+        if response.candidates and response.candidates[0].finish_reason:
+            span.set_attribute("gen_ai.response.finish_reasons", [str(response.candidates[0].finish_reason)])
+        if hasattr(response, "usage_metadata") and response.usage_metadata:
+            if response.usage_metadata.prompt_token_count:
+                span.set_attribute("gen_ai.usage.input_tokens", response.usage_metadata.prompt_token_count)
+            if response.usage_metadata.candidates_token_count:
+                span.set_attribute("gen_ai.usage.output_tokens", response.usage_metadata.candidates_token_count)
+            thoughts_token_count = getattr(response.usage_metadata, "thoughts_token_count", None)
+            if thoughts_token_count:
+                span.set_attribute("gen_ai.usage.reasoning.output_tokens", thoughts_token_count)
```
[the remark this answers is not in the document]
trask: Done in b585683 — switched the token-count guards to `is not None` so 0 values are still captured.

reference/scenarios/google-genai/scenario.py
```diff
@@ -90,6 +90,208 @@ def run_chat():
...
+                if str(getattr(entry, "modality", "")).upper().endswith(modality):
+                    return entry.token_count
+            return None
+
+        if hasattr(response, "usage_metadata") and response.usage_metadata:
+            usage = response.usage_metadata
+            if usage.prompt_token_count:
+                span.set_attribute("gen_ai.usage.input_tokens", usage.prompt_token_count)
+            if usage.candidates_token_count:
+                span.set_attribute("gen_ai.usage.output_tokens", usage.candidates_token_count)
```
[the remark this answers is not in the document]
trask: Done in b585683 (same commit as the other thread).

model/gen-ai/spans.yaml
```diff
@@ -44,8 +44,32 @@ attribute_groups:
...
+          recommended: when the provider reports it
+      - ref: gen_ai.usage.image.output_tokens
+        requirement_level:
+          recommended: when the provider reports it
+      - ref: gen_ai.usage.audio.output_tokens
+        requirement_level:
+          recommended: when the provider reports it
+      - ref: gen_ai.usage.video.output_tokens
+        requirement_level:
+          recommended: when the provider reports it
```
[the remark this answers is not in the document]
trask: Updates:
- Added a Gemini multimodal scenario covering `gen_ai.usage.video.input_tokens`.
- Extended the OpenAI `gpt-4o-audio-preview` scenario to cover `gen_ai.usage.audio.output_tokens` (plus `gen_ai.usage.text.output_tokens`).
- Added an OpenAI GPT Image scenario covering `gen_ai.usage.image.output_tokens`.
- Dropped `gen_ai.usage.video.output_tokens` from the model since Sora (and Veo) bill per second of generated video and don't report an output-token count.

reference/scenarios/openai/scenario.py
```diff
@@ -381,6 +606,98 @@ def run_embeddings_reference(client):
...
+        if resp.usage:
+            span.set_attribute("gen_ai.usage.input_tokens", resp.usage.input_tokens)
+            span.set_attribute("gen_ai.usage.output_tokens", resp.usage.output_tokens)
+            input_details = getattr(resp.usage, "input_tokens_details", None)
+            text_input = getattr(input_details, "text_tokens", None)
+            if text_input is not None:
+                span.set_attribute("gen_ai.usage.text.input_tokens", text_input)
+            image_input = getattr(input_details, "image_tokens", None)
+            if image_input is not None and image_input > 0:
+                span.set_attribute("gen_ai.usage.image.input_tokens", image_input)
```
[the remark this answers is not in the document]
trask: Fixed in 618491e5 — dropped the `> 0` guard so `gen_ai.usage.image.input_tokens` is emitted whenever `input_tokens_details.image_tokens` is reported (including 0). Regenerated `data.json` and reports; `openai` is now listed under `gen_ai.usage.image.input_tokens` in both the event and inference-span coverage tables.

reference/scenarios/openai/scenario.py
```diff
@@ -381,6 +606,98 @@ def run_embeddings_reference(client):
...
+        if resp.usage:
+            event_attrs["gen_ai.usage.input_tokens"] = resp.usage.input_tokens
+            event_attrs["gen_ai.usage.output_tokens"] = resp.usage.output_tokens
+            input_details = getattr(resp.usage, "input_tokens_details", None)
+            text_input = getattr(input_details, "text_tokens", None)
+            if text_input is not None:
+                event_attrs["gen_ai.usage.text.input_tokens"] = text_input
+            image_input = getattr(input_details, "image_tokens", None)
+            if image_input is not None and image_input > 0:
+                event_attrs["gen_ai.usage.image.input_tokens"] = image_input
```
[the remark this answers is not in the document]
trask: Fixed in 618491e5 — dropped the `> 0` guard so `gen_ai.usage.image.input_tokens` is emitted whenever `input_tokens_details.image_tokens` is reported (including 0). Regenerated `data.json` and reports; `openai` is now listed under `gen_ai.usage.image.input_tokens` in both the event and inference-span coverage tables.

docs/gen-ai/openai.md
```diff
@@ -12,7 +12,6 @@ linkTitle: OpenAI
   - [Inference](#inference)
   - [Embeddings](#embeddings)
 - [Metrics](#metrics)
-  - [Metric: `gen_ai.client.token.usage`](#metric-gen_aiclienttokenusage)
   - [Metric: `gen_ai.client.operation.duration`](#metric-gen_aiclientoperationduration)
```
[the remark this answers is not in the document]
trask: Fixed in cf8122ab — added a `Metric: gen_ai.client.inference.tokens` section with the Weaver snippet for `openai.client.inference.tokens`, and a matching TOC entry above the existing `gen_ai.client.operation.duration` link. `make generate-all` populated the table (including `openai.response.service_tier` and `openai.response.system_fingerprint`).

docs/gen-ai/openai.md
```diff
@@ -307,126 +327,6 @@ See [common embeddings span definition](./gen-ai-spans.md#embeddings).
 OpenAI metrics follow [Generative AI metrics](gen-ai-metrics.md) with the noted additional attributes.
 Individual systems may include additional system-specific attributes. It is recommended to check system-specific documentation, if available.
```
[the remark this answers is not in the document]
trask: Fixed in cf8122ab — added a `Metric: gen_ai.client.inference.tokens` section with the Weaver snippet for `openai.client.inference.tokens`, and a matching TOC entry above the existing `gen_ai.client.operation.duration` link. `make generate-all` populated the table (including `openai.response.service_tier` and `openai.response.system_fingerprint`).

docs/gen-ai/anthropic.md
```diff
@@ -85,16 +85,23 @@ Semantic Conventions for [Anthropic](https://www.anthropic.com/) client spans ex
...
-| [`gen_ai.usage.input_tokens`](/docs/registry/attributes/gen-ai.md) | ![Development](https://img.shields.io/badge/-development-blue) | `Recommended` | int | The number of tokens used in the GenAI input (prompt). [12] | `100` |
+| [`gen_ai.usage.audio.input_tokens`](/docs/registry/attributes/gen-ai.md) | ![Development](https://img.shields.io/badge/-development-blue) | `Recommended` when the provider reports it | int | The number of input tokens that the provider reported as audio modality. [10] | `10` |
+| [`gen_ai.usage.audio.output_tokens`](/docs/registry/attributes/gen-ai.md) | ![Development](https://img.shields.io/badge/-development-blue) | `Recommended` when the provider reports it | int | The number of output tokens that the provider reported as audio modality. [11] | `10` |
+| [`gen_ai.usage.cache_creation.input_tokens`](/docs/registry/attributes/gen-ai.md) | ![Development](https://img.shields.io/badge/-development-blue) | `Recommended` | int | The number of input tokens written to a provider-managed cache this turn. [12] | `25` |
```
lmolkova: assuming model providers will start exposing more details, we'll need to break down cache write/read by modality.

E.g. OpenAI pricing shows different costs for cached text, image and realtime (speech) tokens https://openai.com/api/pricing/

lmolkova: also +1 on rename to cache_write (from open questions)

model/gen-ai/metrics.yaml
```diff
@@ -18,10 +18,15 @@ attribute_groups:
       - ref: openai.response.system_fingerprint
         requirement_level: recommended
 metrics:
-  - name: gen_ai.client.token.usage
-    instrument: histogram
+  - name: gen_ai.client.inference.tokens
```
lmolkova, line 21: the original metric was also used to count embeddings (input tokens), should we also add `gen_ai.client.embeddings.tokens`  or `gen_ai.client.embeddings.input_tokens` ?

Alternative is to expand this metric to cover all model-level tokens with something like `gen_ai.client.model.tokens`

model/gen-ai/registry.yaml
```diff
@@ -236,6 +292,74 @@ attributes:
...
+          brief: 'Tokens of unknown or unreported modality.'
+    brief: The modality of the tokens being counted.
+    note: >
+      Use `unknown` when the provider does not report a modality breakdown.
+    examples: ['text', 'audio', 'unknown']
+    stability: development
+  - key: gen_ai.token.cache
+    type:
+      members:
+        - id: none
```
lmolkova: should it rather be unknown? if i recall correctly some models don't (or didn't in old APIs) provide cache breakdown

docs/gen-ai/gen-ai-metrics.md
```diff
@@ -58,35 +64,428 @@ distinct action within a larger Generative AI workflow.
 Individual systems may include additional system-specific attributes.
 It is recommended to check system-specific documentation, if available.
 
-### Metric: `gen_ai.client.token.usage`
+### Metric: `gen_ai.client.inference.tokens`
```
lmolkova: should this be an opt-in metric? or should be postpone introducing it? it's not super-useful since  input and output tokens are almost different units

docs/gen-ai/gen-ai-metrics.md
```diff
@@ -58,35 +64,428 @@ distinct action within a larger Generative AI workflow.
...
+| `response` | Final response tokens. | ![Development](https://img.shields.io/badge/-development-blue) |
+
+<!-- prettier-ignore-end -->
+<!-- END AUTOGENERATED TEXT -->
+<!-- endweaver -->
+
+### Metric: `gen_ai.client.inference.operation.tokens`
 
-When systems report both used tokens and billable tokens, instrumentation MUST report billable tokens.
+This metric is [recommended][MetricRecommended] when the provider reports token
```
lmolkova: should this metric be opt-in? it's not super-useful given that input and output tokens are almost different units

model/gen-ai/registry.yaml
```diff
@@ -236,6 +292,74 @@ attributes:
...
+          brief: 'Image tokens.'
+        - id: audio
+          stability: development
+          value: "audio"
+          brief: 'Audio tokens.'
+        - id: video
+          stability: development
+          value: "video"
+          brief: 'Video tokens.'
+        - id: document
```
lmolkova, line 350: does any provider support this as a modality and bills separately on this?

lmolkova (commented):
Wow, thanks for this proposal! Made a first pass with some questions

model/gen-ai/registry.yaml
alexmojaki: Google reports them, so you can know how many tokens were used on documents, which is useful in theory, but they are charged the same rate as image.

docs/gen-ai/anthropic.md
alexmojaki: these details are definitely already exposed by openai and google

alexmojaki:
>  and reasoning output (priced separately from response output on o-series and Gemini 2.5+).

People do want to track reasoning tokens, but I don't think they have their own pricing on current models, and if I'm wrong I'd love to know.

docs/gen-ai/anthropic.md
trask: https://github.com/open-telemetry/semantic-conventions-genai/pull/197/commits/baf977a6c53b35118ba004cf3c0e02b92ea77d7f

model/gen-ai/registry.yaml
trask: I meant for this to be uncached tokens, probably I should rename to `uncached`? I'm not clear if we need `unknown` or not (in addition), we can definitely add it if it's needed.

trask: (went ahead and renamed `none` to `uncached` in https://github.com/open-telemetry/semantic-conventions-genai/pull/197/commits/961fff5f90ba630f898ff4f1a609364bbbb66a5e)

docs/gen-ai/anthropic.md
trask: added support for cache read/write by modality in https://github.com/open-telemetry/semantic-conventions-genai/pull/197/commits/c0e25435c78845162a95bf58dc9ac45a4167fb3c

@alexmojaki for OpenAI, I only found it available in OpenAI Realtime API (realtime voice), do you happen to know if it's available in the Responses API? I agree it makes sense to model anyways since it seems safe to assume that will be coming regardless (and already here with Gemini).

docs/gen-ai/gen-ai-metrics.md
trask: https://github.com/open-telemetry/semantic-conventions-genai/pull/197/changes/08be15973e3801bcbe0cbf87a819e532a625142b

docs/gen-ai/anthropic.md
alexmojaki: I don't see it for any non-realtime models.

model/gen-ai/metrics.yaml
trask: I like `gen_ai.client.embeddings.tokens`, added in https://github.com/open-telemetry/semantic-conventions-genai/pull/197/commits/e9e44e19970fce166c8d7ece10de28e3c845b027

model/gen-ai/metrics.yaml
```diff
@@ -30,6 +35,96 @@ metrics:
...
+    brief: 'Number of input tokens used in client embeddings operations.'
+    note: |
+      Embeddings operations consume input tokens only and produce no output
+      tokens, so this counter records input tokens exclusively.
+    stability: development
+    annotations:
+      code_generation:
+        metric_value_type: int
+    attributes:
+      - ref_group: metric_attributes.gen_ai
```
Nik-Reddy, line 127: The `gen_ai.client.embeddings.tokens` counter takes only `metric_attributes.gen_ai` and has no `gen_ai.token.modality` attribute, while `gen_ai.client.inference.tokens` requires modality. Multimodal embedding APIs are now common: Cohere Embed v3 multimodal, OpenAI text-embedding-3 with image inputs via the new image-embedding endpoints, Voyage multimodal, and CLIP-family providers all report or could report per-modality token counts on a single embedding call.

Is the omission intentional (single-modality assumption per call) or do we want to add `gen_ai.token.modality` here as well so multimodal embedding instrumentations can break down the count the same way inference does? Either answer works, just want to surface it before the schema settles since the Cohere V2 instrumentation work tracked in (https://github.com/open-telemetry/opentelemetry-python-genai/issues/56) will hit this directly.

model/gen-ai/metrics.yaml
```diff
@@ -30,6 +35,96 @@ metrics:
...
+      equal the total reported on `gen_ai.client.inference.tokens` with
+      `gen_ai.token.type=output`.
+    stability: development
+    annotations:
+      code_generation:
+        metric_value_type: int
+    attributes:
+      - ref_group: metric_attributes.gen_ai
+      - ref: gen_ai.token.phase
+        requirement_level: required
```
Nik-Reddy, line 84: Observation on attribute symmetry across the three new counters:

- `gen_ai.client.inference.tokens` requires `gen_ai.token.type` and `gen_ai.token.modality`
- `gen_ai.client.inference.input_tokens_by_cache` requires `gen_ai.token.cache` and `gen_ai.token.modality`
- `gen_ai.client.inference.output_tokens_by_phase` requires only `gen_ai.token.phase` (no modality)

Input is partitioned by both cache and modality, but output is partitioned only by phase. The base `inference.tokens` counter gives output tokens per modality, and this counter gives output tokens per phase, but the cross (reasoning vs response tokens per modality) is not addressable from these metrics alone.

Was this intentional given current provider data shape (providers do not yet report reasoning tokens broken down by modality, similar to the cache marginal-vs-joint argument already covered in the FAQ), or worth adding `gen_ai.token.modality` here for symmetry with `input_tokens_by_cache`? If intentional, a one-line FAQ entry mirroring the cache rationale would keep future contributors from re-raising it.

model/gen-ai/registry.yaml
```diff
@@ -203,13 +203,81 @@ attributes:
...
     stability: development
-  - key: gen_ai.usage.cache_creation.input_tokens
+  - key: gen_ai.usage.cache_write.input_tokens
     type: int
-    brief: The number of input tokens written to a provider-managed cache.
+    brief: The number of input tokens written to a provider-managed cache this turn.
     note: >
       The value SHOULD be included in `gen_ai.usage.input_tokens`.
     examples: [25]
     stability: development
```
Nik-Reddy, line 212: The recent commit c0e2543 added per-modality `cache_read` variants (`gen_ai.usage.{text,image,audio,video}.cache_read.input_tokens`), but `cache_write` keeps only the aggregate `gen_ai.usage.cache_write.input_tokens`.

Anthropic exposes `cache_creation_input_tokens` as a single aggregate today, even though the cached content blocks are typed. Gemini context caching reports `cached_content_token_count` as an aggregate as well. So at the API surface, providers do not give the per-modality cache-write split natively right now.

Is the asymmetry intentional for this PR (per-modality cache_write deferred until a provider reports the split natively), or should `gen_ai.usage.{text,image,audio,video}.cache_write.input_tokens` be added for parity with the cache_read set? Either way works, just flagging since the per-modality pattern applied to cache_read in c0e2543 makes the missing cache_write counterpart visible. If deferred, a one-line FAQ entry alongside the cache_read rationale would keep future readers from re-raising it.

model/gen-ai/metrics.yaml
trask: > Multimodal embedding APIs are now common: Cohere Embed v3 multimodal, OpenAI text-embedding-3 with image inputs via the new image-embedding endpoints, Voyage multimodal, and CLIP-family providers all report or could report per-modality token counts on a single embedding call.

I'm not able to find how to capture per-modality token counts for any of these embedding APIs, can you provide more details? thanks

model/gen-ai/metrics.yaml
trask: added to FAQ

> ### Why doesn't `output_tokens_by_phase` carry `gen_ai.token.modality`?
>
> `input_tokens_by_cache` carries modality because cached input spans modalities — Gemini reports per-modality cached tokens, so each `(cache, modality)` combination has real data behind it. Reasoning output is text-only across every provider today.

model/gen-ai/registry.yaml
trask: Added to FAQ

> ### Why no per-modality `cache_write` attributes?
>
> Per-modality `cache_read` exists because Gemini reports cached tokens broken down by modality (`cacheTokensDetails`). No provider reports a per-modality cache-write split — Anthropic's `cache_creation_input_tokens` and Gemini's `cached_content_token_count` are both aggregates — so there's nothing to populate per-modality cache-write attributes with today.

model/gen-ai/metrics.yaml
```diff
@@ -30,6 +35,96 @@ metrics:
...
+      - ref: gen_ai.token.modality
+        requirement_level: required
+  - name: gen_ai.client.inference.output_tokens_by_phase
+    instrument: counter
+    unit: "{token}"
+    brief: 'Number of output tokens used in client inference operations, partitioned by generation phase.'
+    note: |
+      When the provider does not separately report reasoning tokens,
+      instrumentations MUST report the full output total under
+      `gen_ai.token.phase=response`.
```
lmolkova, line 72: [OpenAI](https://developers.openai.com/api/docs/guides/reasoning) seems to report reasoning tokens inside output tokens counter 

> While reasoning tokens are not visible via the API, they still occupy space in the model’s context window and are billed as [output tokens](https://openai.com/api/pricing).

[Anthropic](https://platform.claude.com/docs/en/build-with-claude/extended-thinking?utm_source=chatgpt.com#managing-tokens-with-extended-thinking) 

> To see how many billed output tokens were spent on internal reasoning, read usage.output_tokens_details.thinking_tokens in the response. This value reflects the raw reasoning the model generated (not the summarized text returned in the body) and is always less than or equal to output_tokens. Subtract it from output_tokens to approximate the non-reasoning portion of the output.

while [Google GenAI](https://ai.google.dev/api/generate-content) seems to report candidatesTokenCount, toolUsePromptTokenCount, thoughtsTokenCount and no total output tokens.

----
So, we have
- total out (OpenAI, Anthropic) with all phases combined and reasoning part of it
OR
- individual phases (Google GenAI)

Should we have phases `unknown` (total - reasoning for OpenAI and Anthropic), `tool_use`, `visible_response`, and `reasoning`? 

[update] tool_use is not part of output -  https://github.com/open-telemetry/semantic-conventions-genai/pull/197#discussion_r3328015399, but the point on phases (none, visible, reasoning) remains

reference/scenarios/google-genai/scenario.py
```diff
@@ -90,6 +90,251 @@ def run_chat():
...
+        )
+        if response.model_version:
+            span.set_attribute("gen_ai.response.model", response.model_version)
+        if response.candidates and response.candidates[0].finish_reason:
+            span.set_attribute("gen_ai.response.finish_reasons", [str(response.candidates[0].finish_reason)])
+        if hasattr(response, "usage_metadata") and response.usage_metadata:
+            if response.usage_metadata.prompt_token_count is not None:
+                span.set_attribute("gen_ai.usage.input_tokens", response.usage_metadata.prompt_token_count)
+            if response.usage_metadata.candidates_token_count is not None:
+                span.set_attribute("gen_ai.usage.output_tokens", response.usage_metadata.candidates_token_count)
```
lmolkova, line 130: the total would be a sum of candidates, tool use and thinking tokens - https://ai.google.dev/api/generate-content

alexmojaki: I've made this mistake too, toolUsePromptTokenCount is actually counted as input tokens

chightow:
"Reasoning" tokens as far as I know aren't priced differently, they are important to track as the scratch work often ends up being a large differentiator of cost between different models ex.

Opus 4.5, 4.6 , 4.7, 4.8 for example all currently priced the same main differentiator in cost I've observed has been the extent to which the models are configured to aggressively use the "reasoning" block revealing this to end users in an easy way to show what "reasoning" is costing them and to help them make the determination if the "reasoning" provides any value to a given use case.

Thank you for this pr from downstream 

It should greatly simplify observability.

model/gen-ai/metrics.yaml
Nik-Reddy: Thanks for asking for specifics.

Cohere Embed v2: the response is `EmbedByTypeResponse`, which carries `meta: ApiMeta`. `ApiMeta.billed_units` is an `ApiMetaBilledUnits`, which exposes `input_tokens` (text) and `image_tokens` separately, alongside `images` for image count. So on a mixed text+image embedding call, the per-modality token usage is addressable directly from the response. Sources:
- https://github.com/cohere-ai/cohere-python/blob/main/src/cohere/types/embed_by_type_response.py
- https://github.com/cohere-ai/cohere-python/blob/main/src/cohere/types/api_meta.py
- https://github.com/cohere-ai/cohere-python/blob/main/src/cohere/types/api_meta_billed_units.py

Voyage multimodal: `MultimodalEmbeddingsObject` exposes `text_tokens`, `image_pixels`, `video_pixels`, and `total_tokens`. The pixel fields are not token fields directly, but Voyage documents the conversion (560 image pixels = 1 token, 1120 video pixels = 1 token), so the SDK exposes enough per-modality usage to derive modality-specific token counts. Source:
- https://github.com/voyage-ai/voyageai-python/blob/main/voyageai/object/multimodal_embeddings.py

On the other two I cited:
- OpenAI text-embedding-3 does not accept image inputs today. My mistake on grouping it in.
- CLIP-family is local model usage, not API-emitted, so it is not a good example here.

Given Cohere and Voyage report this today, I think adding `gen_ai.token.modality` to `gen_ai.client.embeddings.tokens` is the better fit, so instrumentations can break down embedding token usage the same way inference does. If you would rather defer until more providers expose this natively, an FAQ entry mirroring the cache rationale would document the current limitation.

reference/scenarios/google-genai/scenario.py
lmolkova: ah, I see, so "output only" here means something else (not output tokens)

image

and it makes sense that tool prompt is part of input. Do we need to break input tokens then? Assuming we do it later, we'd do `gen_ai.usage.tool_use.input_tokens` and potentially need per-modality breakdown `gen_ai.usage.image.tool_use.input_tokens` (and it seems there is no need to have cache + tool use variant).

Seems like something we can do later.

lmolkova:
We're going to discuss this on an ad-hoc call tomorrow at 12pm PT: [Calendar](https://calendar.google.com/calendar/u/0/r/day/2026/6/5?eid=MnI0NDloamxmdGU3b2ZnYXNsYzR1ZmZoZ2wgY18yYmY3M2UzYjZiNTMwZGE0YmFiZDQ0NGU3MmI3NmE2YWQ4OTNhNWMzZjQzY2Y0MDQ2N2FiYzdhOWE4OTdmOTc3QGc), [Zoom](https://zoom.us/j/91357539844?pwd=MU9ZRUNyVUwrcHdiclU2b2lPOVBkQT09).

Please join if you can @alexmojaki @Nik-Reddy @aabmass @trask

model/gen-ai/spans.yaml
```diff
@@ -42,10 +42,43 @@ attribute_groups:
         requirement_level: recommended
       - ref: gen_ai.usage.cache_read.input_tokens
         requirement_level: recommended
-      - ref: gen_ai.usage.cache_creation.input_tokens
+      - ref: gen_ai.usage.cache_write.input_tokens
```
alexmojaki, line 45: if we're going to make breaking change renames, consider a more hierarchical top down namespaced approach like `gen_ai.usage.tokens.input.cache_read.text`

model/gen-ai/spans.yaml
```diff
@@ -42,10 +42,43 @@ attribute_groups:
...
+      - ref: gen_ai.usage.image.input_tokens
+        requirement_level:
+          recommended: when the provider reports it
+      - ref: gen_ai.usage.audio.input_tokens
+        requirement_level:
+          recommended: when the provider reports it
+      - ref: gen_ai.usage.video.input_tokens
+        requirement_level:
+          recommended: when the provider reports it
+      - ref: gen_ai.usage.text.cache_read.input_tokens
```
alexmojaki, line 59: is it really necessary to individually spec each combination of dimensions, and only those that are currently known to exist? can there not be a 'template spec' along the lines of `gen_ai.usage.<modality>.<cache>.input_tokens` etc?

alternatively, should these live in a complex object instead?

alexmojaki:
> making cost queries that match the provider pricing structure (per-modality, per-cache-state unit prices) answerable from metrics alone

costs can't be calculated from metrics in general anyway since the price per token often depends on the number of (input) tokens in a request, unless you added another metric attribute indicating the pricing bucket.

model/gen-ai/metrics.yaml
```diff
@@ -30,6 +35,96 @@ metrics:
       - ref_group: metric_attributes.gen_ai
       - ref: gen_ai.token.type
         requirement_level: required
+      - ref: gen_ai.token.modality
+        requirement_level: required
+  - name: gen_ai.client.inference.input_tokens_by_cache
```
alexmojaki, line 40: The rules surrounding the various counters, uncached, unknown modality, etc are pretty confusing, and it seems like in the end they still don't cover all use cases, like providers that report cache breakdown and modality breakdown but not the joint. maybe metric attributes aren't a good fit here, and there should be a large number of 'simple' counters corresponding directly to the various span attributes? e.g. one counter for input tokens, one for cache read input, one for image input, one for cache read image input, etc.

alexmojaki: i think that would also allow each one to be a histogram

alexmojaki:
> Why don't gen_ai.token.modality and gen_ai.output.type use the same enum values?

was this removed?

model/gen-ai/spans.yaml
alexmojaki, line 1: Should span attributes and metrics be in the same PR?

trask:

Create `changelog.d/197.breaking.md` containing the change log entry, e.g.:

```md
Restructure GenAI token usage attributes and metrics under a unified `gen_ai.token.*` namespace. - Added per-modality span attributes `gen_ai.usage.{text,image,audio,video}.{input,output}_tokens`. - Replaced the `gen_ai.client.token.usage` histogram with counters `gen_ai.client.inference.tokens` (dimensioned by `gen_ai.token.type` and `gen_ai.token.modality`), `gen_ai.client.inference.input_tokens_by_cache` (dimensioned by `gen_ai.token.cache`), and `gen_ai.client.inference.output_tokens_by_phase` (dimensioned by `gen_ai.token.phase`). Added opt-in histograms `gen_ai.client.inference.operation.{tokens,input_tokens,output_tokens}`. - Added new enum attributes `gen_ai.token.modality` (`text`/`image`/`audio`/`video`/`document`/`unknown`), `gen_ai.token.cache` (`none`/`read`/`creation`), and `gen_ai.token.phase` (`response`/`reasoning`).
```

Thanks!

Necmttn:
From a downstream cost-accounting perspective, I would document one boundary explicitly: these counters reconcile usage, not billing truth.

They are the right shape for token totals by modality/cache/phase. Exact cost still needs a request-level price reference or ledger event because pricing can vary by:

- model/version
- cache read/write tier
- request-size bucket
- provider routing
- contract or subscription context

A short note in the metric docs would help consumers avoid turning `sum(tokens) * price` into a misleading bill. The semconv can still support cost dashboards, but the pricing source/version needs to be carried by the producer or joined from a separate ledger.

---

_Generated with [ax](https://github.com/Necmttn/ax)._

PR #198, opened 2026-05-26
Remove OTEL_SEMCONV_STABILITY_OPT_IN transition blurb

trask:
Remove the v1.36.0 transition-plan blurb from the GenAI docs since this is a new repository with a new semantic-conventions URL.

### DIFF
--- a/docs/gen-ai/README.md
+++ b/docs/gen-ai/README.md
@@ -6,27 +6,6 @@ linkTitle: Generative AI
 
 **Status**: [Development][DocumentStatus]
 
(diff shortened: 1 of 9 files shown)


lmolkova (approved):

MikeGoldsmith (approved):

DylanRussell:
Nice.. so eventually we will get the public docs updated too right (https://opentelemetry.io/docs/specs/semconv/gen-ai/gen-ai-events/)  ? Should we mark any of the events / spans as "stable" or wait to do that ?

PR #126, opened 2026-05-08
Add workflow duration metric

wrisa:
## Description

Brought from old repo: https://github.com/open-telemetry/semantic-conventions/pull/3565

## Checklist

- [ ] Motivation section filled in above
- [ ] Reference instrumentation and scenarios updated for affected libraries
- [ ] Changelog entry added under `Unreleased` in `CHANGELOG.md` for any change to the conventions that a consumer would care about. Editorial changes (typos, pure rewording, repo tooling) don't need an entry.

See [CONTRIBUTING.md]

PR #201, opened 2026-05-27
Add gen_ai.invoke_agent.duration and gen_ai.execute_tool.duration metrics

pvlsirotkin:

## Description

Adds two new GenAI semantic convention metrics for agent and tool latency,
modeled on the recently-merged `gen_ai.workflow.duration` metric
([#126](https://github.com/open-telemetry/semantic-conventions-genai/pull/126))
and following the metric-naming proposal in
[#249](https://github.com/open-telemetry/semantic-conventions-genai/issues/249)
(metric name aligned with the operation name):

- **`gen_ai.invoke_agent.duration`** (Histogram, `s`): end-to-end duration
  of a single agent invocation. Aligns with the existing
  `gen_ai.invoke_agent` client and internal spans.
- **`gen_ai.execute_tool.duration`** (Histogram, `s`): duration of a single
  tool execution. Aligns with the existing `gen_ai.execute_tool` span.

## Motivation

The conventions already define `invoke_agent` and `execute_tool` spans (with
`gen_ai.agent.name`, `gen_ai.tool.name`, `error.type`, etc.) and now
`gen_ai.workflow.duration` for the workflow boundary, but there is no
standard metric for **individual agent invocations** or **individual tool
executions**. Operators today cannot build agent- or tool-level latency /
error-rate dashboards without inventing custom metric names.

User journey: latency SLOs, error-rate dashboards, capacity planning, and
regression detection across agent / tool versions.

## Checklist


docs/gen-ai/gen-ai-metrics.md
```diff
@@ -936,6 +940,137 @@ If there is no low-cardinality workflow name available for a given framework, th
...
+It is recommended to check system-specific documentation, if available.
+
+`gen_ai.tool.execution.duration` represents the duration of a single tool
+execution performed by or on behalf of a GenAI agent. It is intended for
+instrumentations of agent frameworks (or of application code that executes
+tools on behalf of an agent) that can reliably bound a single tool call.
+
+### Metric: `gen_ai.tool.execution.duration`
+
+This metric is [recommended][MetricRecommended] for instrumentations that can
```
MikeGoldsmith, line 1011: Why is this _recommended_ while `gen_ai.agent.invocation.duration` is _required_?

I think it's because tools can be executed by uninstrumented paths but it's worth stating here so future readers don't have to derive it.

model/gen-ai/metrics.yaml
```diff
@@ -113,6 +113,56 @@ metrics:
...
+      response (or terminates with an error).
+
+      When this metric is reported alongside a `gen_ai.invoke_agent` span, the
+      metric value SHOULD be the same as the span duration.
+    instrument: histogram
+    unit: "s"
+    stability: development
+    attributes:
+      - ref_group: attributes.gen_ai.error
+      - ref: gen_ai.agent.name
```
MikeGoldsmith, line 148: Should this be `required` rather than `conditionally_required when available_`
Without `gen_ai.agent.name` the metric becomes a single global histogram and operators won't be able to break latency down by agent.

`gen_ai.tool.name` on the tool metric below is required as the primary dimension. The `invoke_agent` span already carries `gen_ai.agent.name` so I don't see why it would be missing at metric-recording time?

CHANGELOG.md
```diff
@@ -24,6 +24,11 @@
   ([#126](https://github.com/open-telemetry/semantic-conventions-genai/pull/126))
 - Add `moonshot_ai` to `gen_ai.provider.name` well-known values.
   ([#99](https://github.com/open-telemetry/semantic-conventions-genai/pull/99))
+- Add `gen_ai.agent.invocation.duration` metric to track the end-to-end duration
```
MikeGoldsmith: The PR checklist item "Reference instrumentation and scenarios updated for affected libraries" is unchecked and this PR doesn't touch `reference/scenarios/`.

The CONTRIBUTING.md requires new conventions need at least one scenario that emits them, otherwise the conformance testing has nothing to validate. Please add a scenario in this PR.

model/gen-ai/metrics.yaml
```diff
@@ -113,6 +113,56 @@ metrics:
...
+      or on behalf of a GenAI agent.
+
+      When this metric is reported alongside a `gen_ai.execute_tool` span, the
+      metric value SHOULD be the same as the span duration.
+    instrument: histogram
+    unit: "s"
+    stability: development
+    attributes:
+      - ref_group: attributes.gen_ai.error
+      - ref: gen_ai.tool.name
```
MikeGoldsmith, line 174: Should we also add `gen_ai.tool.type`? The `execute_tool` span carries it, so the framework already knows it at metric-recording time.

MikeGoldsmith (changes requested):
Looks good, thanks @pvlsirotkin.

There's a few things I think we need to fix up, but heading in the right direction.

docs/gen-ai/gen-ai-metrics.md
pvlsirotkin: Yep, exactly that. I'll add a sentence to the section explaining the tool/agent asymmetry – tools can be invoked through paths the agent framework doesn't observe.

model/gen-ai/metrics.yaml
pvlsirotkin: The existing `invoke_agent.client` / `invoke_agent.internal` spans use `conditionally_required: when available` for `gen_ai.agent.name` (see [`spans.yaml#L111-L113`](https://github.com/open-telemetry/semantic-conventions-genai/blob/main/model/gen-ai/spans.yaml#L111-L113)) – that's what I was mirroring here.

Would `recommended` work as a compromise? Stronger than the current level, but leaves room for frameworks where the agent has no stable name. Happy to go all the way to `required` if you'd rather.

Same change would apply to #202 and #203.

model/gen-ai/metrics.yaml
pvlsirotkin: Good idea, added at `recommended` to match the [`execute_tool` span level](https://github.com/open-telemetry/semantic-conventions-genai/blob/main/model/gen-ai/spans.yaml).

CHANGELOG.md
pvlsirotkin: Looked into this and the scenario tooling doesn't persist metrics to `scenarios/*/data.json` today — the writer in [`reference/src/semconv_genai/data_files.py`](https://github.com/open-telemetry/semantic-conventions-genai/blob/main/reference/src/semconv_genai/data_files.py) only emits `spans` and `events` keys (no `metrics`), and none of the existing `data.json` files include a `metrics` section. The parser collects metric signals in-memory but they're never written out. The repo README also has no Metrics report table.

The existing `google-adk` scenario pins ADK v1.34.1, which already emits all 5 of the proposed metrics, so the SDK side is ready — but the conformance pipeline needs to learn about metrics first.

Or am I missing something — is there a way to wire up metric validation that I haven't spotted? Either way, happy to do the work; just want to make sure we're aligned on the approach.

model/gen-ai/registry.yaml
```diff
@@ -266,6 +266,17 @@ attributes:
     brief: Name of the tool utilized by the agent.
     examples: ["Flights"]
     stability: development
+  - key: gen_ai.tool.version
+    type: string
+    brief: The version of the tool utilized by the agent.
+    note: |
+      The tool version is usually provided by the application that defines the
+      tool. It is typically a static value (for example, a release tag of the
+      tool's package) and is expected to have low cardinality.
```
lmolkova: are there GenAI frameworks / libs that have a notion of tool version? 

The application version is expected to be reported in resource attributes (e.g. service.version) or instrumentation scope version associated with metric / span / log.

model/gen-ai/metrics.yaml
```diff
@@ -113,6 +113,58 @@ metrics:
...
+    stability: development
+    attributes:
+      - ref_group: attributes.gen_ai.error
+      - ref: gen_ai.agent.name
+        requirement_level:
+          conditionally_required: when available
+      - ref: gen_ai.agent.version
+        requirement_level:
+          conditionally_required: when available
+  - name: gen_ai.tool.execution.duration
```
lmolkova: we've been discussing general approach to metric naming - https://github.com/open-telemetry/semantic-conventions-genai/issues/249

I think we should align operation name with metric name and the proposal for this one looks like `gen_ai.execute_tool.duration`. 

I'm planning to bring it up on tomorrow's GenAI call to confirm the naming pattern and will post an update here after

model/gen-ai/metrics.yaml
```diff
@@ -113,6 +113,58 @@ metrics:
...
+    annotations:
+      code_generation:
+        metric_value_type: double
+    brief: 'GenAI tool execution duration.'
+    note: >
+      This metric measures the duration of a single tool execution performed by
+      or on behalf of a GenAI agent.
+
+      When this metric is reported alongside a `gen_ai.execute_tool` span, the
+      metric value SHOULD be the same as the span duration.
```
lmolkova:
```suggestion
    brief: 'The duration of a single tool execution performed by
      or on behalf of a GenAI agent'
    note: >
      When this metric is reported alongside a `gen_ai.execute_tool` span, the
      metric value SHOULD be the same as the span duration.
```

model/gen-ai/metrics.yaml
```diff
@@ -113,6 +113,58 @@ metrics:
       - ref: gen_ai.workflow.name
         requirement_level:
           conditionally_required: If available.
+  - name: gen_ai.agent.invocation.duration
```
lmolkova: similar to execute_tool, following https://github.com/open-telemetry/semantic-conventions-genai/issues/249 the proposal is to align metric name with operation name and call it `gen_ai.invoke_agent.duration` on internal / local agents and `gen_ai.client.invoke_agent.duration` when invoking remote agents.

For the scope of this PR, I think it's fine to start with `gen_ai.invoke_agent.duration` only

model/gen-ai/metrics.yaml
```diff
@@ -113,6 +113,58 @@ metrics:
...
+      code_generation:
+        metric_value_type: double
+    brief: 'GenAI agent invocation duration.'
+    note: >
+      This metric measures the end-to-end duration of a single agent invocation,
+      from the moment the agent is invoked to the moment it produces its final
+      response (or terminates with an error).
+
+      When this metric is reported alongside a `gen_ai.invoke_agent` span, the
+      metric value SHOULD be the same as the span duration.
```
lmolkova: nit
```suggestion
    brief: ' The end-to-end duration of a single agent invocation,
      from the moment the agent is invoked to the moment it produces its final
      response (or terminates with an error).'
    note: >
      When this metric is reported alongside a `gen_ai.invoke_agent` span, the
      metric value SHOULD be the same as the span duration.
```

model/gen-ai/metrics.yaml
```diff
@@ -113,6 +113,58 @@ metrics:
...
+    unit: "s"
+    stability: development
+    attributes:
+      - ref_group: attributes.gen_ai.error
+      - ref: gen_ai.agent.name
+        requirement_level:
+          conditionally_required: when available
+      - ref: gen_ai.agent.version
+        requirement_level:
+          conditionally_required: when available
```
lmolkova: there seem to be other attributes on the invoke_agent spans what would make sense on metric, e.g. request model, it would make sense to align metric and span attributes (ones with low cardinality)

docs/gen-ai/gen-ai-metrics.md
```diff
@@ -936,6 +940,151 @@ If there is no low-cardinality workflow name available for a given framework, th
...
+
+Individual systems may include additional system-specific attributes.
+It is recommended to check system-specific documentation, if available.
+
+`gen_ai.agent.invocation.duration` represents the end-to-end duration of a
+single agent invocation, measured from the point where the agent is invoked
+to the point where it produces its final response (or terminates with an
+error). It is intended for instrumentations of agent frameworks (for example,
+ADK, LangChain agents, CrewAI agents) that can reliably bound a single agent
+invocation.
```
lmolkova: would it be possible to move parts of this paragraph that are not in the metric note yaml there? There seems to be a lot in common and having things in yaml is preferable over pure MD.

docs/gen-ai/gen-ai-metrics.md
```diff
@@ -936,6 +940,151 @@ If there is no low-cardinality workflow name available for a given framework, th
...
+
+## Generative AI tool metrics
+
+Individual systems may include additional system-specific attributes.
+It is recommended to check system-specific documentation, if available.
+
+`gen_ai.tool.execution.duration` represents the duration of a single tool
+execution performed by or on behalf of a GenAI agent. It is intended for
+instrumentations of agent frameworks (or of application code that executes
+tools on behalf of an agent) that can reliably bound a single tool call.
```
lmolkova: let's move it to the metric note in yaml

docs/gen-ai/gen-ai-metrics.md
```diff
@@ -936,6 +940,151 @@ If there is no low-cardinality workflow name available for a given framework, th
...
+
+Unlike `gen_ai.agent.invocation.duration` (which is required), this metric is
+only recommended because tools may be executed through paths that the agent
+framework does not observe, such as external MCP servers or
+application-managed dispatch. Instrumentations SHOULD record this metric for
+every tool execution they observe, but are not required to capture all tool
+calls across the agentic system.
+
+When this metric is reported alongside a `gen_ai.execute_tool` span, the
+metric value SHOULD be the same as the span duration.
```
lmolkova: nit: this is already captured in yaml and is generated in the block below

docs/gen-ai/gen-ai-metrics.md
```diff
@@ -936,6 +940,151 @@ If there is no low-cardinality workflow name available for a given framework, th
...
+
+This metric is [recommended][MetricRecommended] for instrumentations that can
+observe tool executions performed by or on behalf of a GenAI agent.
+
+Unlike `gen_ai.agent.invocation.duration` (which is required), this metric is
+only recommended because tools may be executed through paths that the agent
+framework does not observe, such as external MCP servers or
+application-managed dispatch. Instrumentations SHOULD record this metric for
+every tool execution they observe, but are not required to capture all tool
+calls across the agentic system.
```
lmolkova: nit:

```suggestion
```

this seems to be be captured in the sentence above.

docs/gen-ai/gen-ai-metrics.md
```diff
@@ -936,6 +940,151 @@ If there is no low-cardinality workflow name available for a given framework, th
...
+### Metric: `gen_ai.agent.invocation.duration`
+
+This metric is [required][MetricRequired] when the instrumented component
+implements agent invocation operations.
+
+When this metric is reported alongside a `gen_ai.invoke_agent` span, the
+metric value SHOULD be the same as the span duration.
+
+This metric SHOULD be specified with [ExplicitBucketBoundaries] of
+[0.01, 0.02, 0.04, 0.08, 0.16, 0.32, 0.64, 1.28, 2.56, 5.12, 10.24, 20.48, 40.96, 81.92].
```
lmolkova: should we scale up histogram boundaries? I would expect agent call duration to easily take longer than 90 sec, I think a reasonable range could be minutes, perhaps we can do something like 

[0.4, 0.8, 1.6, 3.2, 6.4, 12.8, 25.6, 51.2, 102.4, 204.8, 409.6, 819.2] ?

lmolkova (commented):
Thanks for the PR!

Left some mostly cosmetic suggestions. I'm going to bring metric naming to the GenAI SIG call tomorrow and will post the outcome on this PR.

model/gen-ai/registry.yaml
pvlsirotkin: Good point — dropped both `gen_ai.tool.version` and `gen_ai.agent.version` from the metric attribute lists. Tool/agent versions are reasonably covered by `service.version` (resource attribute) and instrumentation scope version. The ADK Python implementation also declares `gen_ai.tool.version` as a constant but doesn't actually record it, which further supports dropping. Also removed `gen_ai.tool.version` from the registry entirely (this PR was the one adding it).

docs/gen-ai/gen-ai-metrics.md
pvlsirotkin: Agreed on scaling up. Bumped to `[0.1, 0.2, 0.4, 0.8, 1.6, 3.2, 6.4, 12.8, 25.6, 51.2, 102.4, 204.8, 409.6]` — starts at 0.1s for sub-second resolution on fast agent calls, upper bound at ~6.8 min which should cover the typical synchronous range without going as wide as 14 min.

Kept the tool metric (`gen_ai.tool.execution.duration`) at `[0.01..81.92]` since single tool calls are closer to LLM-call durations. Happy to scale either further if you have data suggesting longer tails.

model/gen-ai/metrics.yaml
pvlsirotkin: Holding off on adding `gen_ai.provider.name` / `gen_ai.request.model` here for now — the original use case behind this metric is a single-model agent where those dimensions don't add much (`provider` and `model` are already captured on the child `gen_ai.inference` span, and on `gen_ai.client.operation.duration` for any LLM-call-level slicing). Adding them on the agent metric would also double-count cardinality if dashboards also slice the LLM-call metric.

Happy to revisit if there's a concrete use case where slicing agent-level duration by provider/model is useful that I'm missing.

model/gen-ai/metrics.yaml
pvlsirotkin: Applied. Will be in the next push.

model/gen-ai/metrics.yaml
pvlsirotkin: Applied. Will be in the next push.

docs/gen-ai/gen-ai-metrics.md
pvlsirotkin: Moved into the metric `note` in YAML. The MD now just has the section header + preamble + bucket boundaries; the full guidance (when to use `gen_ai.client.operation.duration` vs `gen_ai.workflow.duration`) is rendered from YAML in the autogen block below.

docs/gen-ai/gen-ai-metrics.md
pvlsirotkin: Moved into the tool metric `note` in YAML.

docs/gen-ai/gen-ai-metrics.md
pvlsirotkin: Removed the prose duplicate; the autogen block below renders the same content from YAML.

docs/gen-ai/gen-ai-metrics.md
pvlsirotkin: Removed the explanatory paragraph from the prose. The tool/agent asymmetry reasoning (Mike asked for that earlier) is now in the YAML `note` so it still shows up in the autogen block.

model/gen-ai/metrics.yaml
lmolkova: https://github.com/open-telemetry/semantic-conventions-genai/pull/270 proposes to have an entity for agents so that hosted agents would record their info there instead of recording all agent info on every measurement.

So I think whatever we come up with will change in the future once we add entities/resources into the picture and attribute on the metric itself can't be required. 

For now, I think it makes sense to keep it consistent with internal agent span that (for some unknown to me reason) has `conditionally_required: When available` level

trask:

Create `changelog.d/201.enhancement.md` containing the change log entry, e.g.:

```md
Add `gen_ai.agent.invocation.duration` metric to track the end-to-end duration of a single agent invocation, and `gen_ai.tool.execution.duration` metric to track the duration of a single tool execution.
```

Thanks!

model/gen-ai/metrics.yaml
```diff
@@ -116,6 +116,72 @@ metrics:
...
+      - ref: gen_ai.agent.name
+        requirement_level:
+          conditionally_required: when available
+  - name: gen_ai.tool.execution.duration
+    annotations:
+      code_generation:
+        metric_value_type: double
+    brief: >
+      The duration of a single tool execution performed by or on behalf of a
+      GenAI agent.
```
lmolkova: nit, tool execution can be performed by a generic application using LLM (of course you can call it an agent too)
 
```suggestion
      The duration of a single tool execution.
```

model/gen-ai/metrics.yaml
lmolkova: Let's go ahead with `gen_ai.execute_tool.duration` since it aligns with the operation name on spans.

The discussion on the call for #249 didn't raise any objections.

model/gen-ai/metrics.yaml
```diff
@@ -116,6 +116,72 @@ metrics:
...
+      code that executes tools on behalf of an agent) that can reliably
+      bound a single tool call.
+
+      Unlike `gen_ai.agent.invocation.duration` (which is required), this
+      metric is only recommended because tools may be executed through
+      paths that the agent framework does not observe — for example,
+      external MCP servers or application-managed dispatch.
+      Instrumentations SHOULD record this metric for every tool execution
+      they observe but are not required to capture all tool calls across
+      the agentic system.
```
lmolkova:
```suggestion
      Instrumentation that can reliably bound a single tool call SHOULD
      record this metric for every tool execution they can observe.
```

Suggesting a simplification. I believe we're moving away from having required and recommended levels for metrics to (on by default and off by default)  - https://github.com/open-telemetry/semantic-conventions/issues/3278

model/gen-ai/metrics.yaml
```diff
@@ -116,6 +116,72 @@ metrics:
...
+    stability: development
+    attributes:
+      - ref_group: attributes.gen_ai.error
+      - ref: gen_ai.tool.name
+        requirement_level: required
+      - ref: gen_ai.tool.type
+        requirement_level: recommended
+      - ref: gen_ai.agent.name
+        requirement_level:
+          conditionally_required: when available
```
lmolkova: nit

```suggestion
          conditionally_required: When available.
```

model/gen-ai/metrics.yaml
```diff
@@ -116,6 +116,72 @@ metrics:
...
         requirement_level:
           conditionally_required: If available.
+  - name: gen_ai.agent.invocation.duration
+    annotations:
+      code_generation:
+        metric_value_type: double
+    brief: >
+      The end-to-end duration of a single agent invocation,
+      from the moment the agent is invoked to the moment it produces its final
+      response (or terminates with an error).
```
lmolkova: nit

```suggestion
      The end-to-end duration of a single agent invocation, 
      from the moment the invocation starts until the agent emits 
      the last chunk of its final response or terminates with an error.
```

model/gen-ai/metrics.yaml
```diff
@@ -116,6 +116,72 @@ metrics:
...
+      available.
+
+      When this metric is reported alongside a `gen_ai.invoke_agent` span,
+      the metric value SHOULD be the same as the span duration.
+    instrument: histogram
+    unit: "s"
+    stability: development
+    attributes:
+      - ref_group: attributes.gen_ai.error
+      - ref: gen_ai.agent.name
```
lmolkova, line 148: invoke agent span has a few more attributes that could be useful on metrics. In particular:

- agent.id, version, and description
- request model

I think we should reference them. The agent info will also come through entity in #270

lmolkova (commented):
This looks great, just a few comments on naming and suggestions to add agent info attributes to metrics

CHANGELOG.md
pvlsirotkin: https://github.com/open-telemetry/semantic-conventions-genai/issues/281

model/gen-ai/metrics.yaml
pvlsirotkin: Sounds good — kept as `conditionally_required: When available.` (matching the internal `invoke_agent` span). Also fixed the capitalization per #245.

model/gen-ai/metrics.yaml
pvlsirotkin: Done — renamed to `gen_ai.execute_tool.duration` (and `gen_ai.invoke_agent.duration` for the agent metric).

model/gen-ai/metrics.yaml
pvlsirotkin: Done — renamed to `gen_ai.invoke_agent.duration`. Skipped the `.client.invoke_agent.duration` variant for this PR per your guidance.

model/gen-ai/metrics.yaml
pvlsirotkin: Applied.

model/gen-ai/metrics.yaml
pvlsirotkin: Applied. Good context on the required/recommended direction change — thanks for the heads-up.

model/gen-ai/metrics.yaml
pvlsirotkin: Applied — and standardized the same capitalization on every `conditionally_required` note on these metrics.

model/gen-ai/metrics.yaml
pvlsirotkin: Applied.

model/gen-ai/metrics.yaml
pvlsirotkin: Added `gen_ai.agent.id`, `gen_ai.agent.version`, and `gen_ai.request.model` (all `conditionally_required: When/If available.`). Skipped `gen_ai.agent.description` since it tends to be free-form text and risks high cardinality on a metric.

pvlsirotkin:
Done — added `changelog.d/201.enhancement.md` and removed the direct `CHANGELOG.md` edit. Thanks for the heads-up about #275.

model/gen-ai/metrics.yaml
pvlsirotkin: Quick follow-up on this: ended up dropping `gen_ai.agent.version` and `gen_ai.request.model` again on a second look. Kept `gen_ai.agent.id`.

- `gen_ai.agent.version`: the intent in this metric set is to use the `service.version` resource attribute (which is stable, applied process-wide via the resource) for agent/tool/app versioning, rather than carrying a per-signal `gen_ai.agent.version` dimension on every metric. The registry attribute itself stays (it was added by @trask in a separate PR), it just isn't a dimension on this metric.
- `gen_ai.request.model`: not always available at the time this metric is recorded. The framework knows the agent it's invoking, but the model selection can happen later inside the agent (and may differ across the multiple LLM calls within one invocation). The model breakdown is already on the child `gen_ai.inference` / `gen_ai.client.operation.duration` signals; adding it here would conflate two different boundaries.

Happy to revisit either if there's a concrete use case I'm missing.

model/gen-ai/metrics.yaml
```diff
@@ -116,6 +116,65 @@ metrics:
...
+    instrument: histogram
+    unit: "s"
+    stability: development
+    attributes:
+      - ref_group: attributes.gen_ai.error
+      - ref: gen_ai.tool.name
+        requirement_level: required
+      - ref: gen_ai.tool.type
+        requirement_level: recommended
+      - ref: gen_ai.agent.name
```
lmolkova, line 178: if #270 is merged before this PR, we'll need to add entity associations and adjust requirement level accordingly, just leaving a comment so we don't forget

changelog.d/201.enhancement.md
```diff
@@ -0,0 +1 @@
+Add `gen_ai.invoke_agent.duration` metric to track the end-to-end duration of a single agent invocation, and `gen_ai.execute_tool.duration` metric to track the duration of a single tool execution.
```
lmolkova, line 1: could you please update PR description and title to reflect new naming? thanks!

lmolkova (approved):
LGTM, thanks!

model/gen-ai/metrics.yaml
lmolkova: I believe agent version != service version for remote agents and even when you're hosting an agent the version of agent is NOT the version user sees - you can update your service without incrementing agent version because nothing has changed in user-related agent config or behavior.

For the  request model, we can document that it's the model provided when agent was created. Given that pretty much all frameworks take it as constructor parameter, it seems to be common and important use case.

They can both be `recommended: When available and meaningful / provide information not captured service.version / etc`

changelog.d/201.enhancement.md
aabmass: IMO let's merge this one first, if it's ready, and I'll update my PR.

model/gen-ai/metrics.yaml
pvlsirotkin: Update: added `gen_ai.agent.version` back. Still skeptical about `gen_ai.request.model` though — the multi-model agent case (planner picks executor, "thinking" model delegates to "doing" model, etc.) isn't well covered: the framework would have to pick one value (configured primary? first call? last call?) and different frameworks would pick differently. Happy to add it if there's a clear single-model convention you'd like to land on.

docs/gen-ai/gen-ai-metrics.md
MikeGoldsmith: Still missing this reference even though it was resolved.

MikeGoldsmith (approved):
Looks good, I'd like to see a scenario added when we can to cover the gaps and add the missing sentence for `gen_ai.agent.invocation.duration` (see below).

pvlsirotkin:

**State of the PR:**
- All review feedback so far has been addressed in commits.
- 2 review threads remain open, both waiting on @lmolkova:
  1. [Add more invoke_agent attrs (agent.id, version, description, request.model)](https://github.com/open-telemetry/semantic-conventions-genai/pull/201#discussion_r3391935709) — last I added `agent.id` and `agent.version` back; still pushing back on `request.model` due to the multi-model-agent ambiguity. Awaiting @lmolkova's take.
  2. [Earlier thread on `provider.name` / `request.model`](https://github.com/open-telemetry/semantic-conventions-genai/pull/201#discussion_r3376499335) — same topic, my pushback there mirrors the above.

- Naming follows #249 (lmolkova confirmed on the SIG call).
- Versioning intent across this metric set is to use `service.version` on the resource for the common single-agent-per-process case; the entity work in #270 will reshape per-agent versioning later.
- We were not sure that `request.model` info would be available at agent-metric-recording time — model can be selected later, can differ across LLM calls within one invocation. Open to adding if a clear single-model convention lands.

Thanks @MikeGoldsmith @lmolkova @trask @aabmass for all the feedback so far.

model/gen-ai/metrics.yaml
lmolkova: We can clarify that this is the model provided when creating the agent and should be provided if and only if specific framework has this construction-time parameter. It should not be populated when not applicable or when multiple models can be configured on the same agent. The motivation to include it is consistency between spans and metric. Invoke agent internal spans has request model, it's low cardinality and common parameter on existing agent frmaeowrks, so metric should have it too. 

This would be recommended attribute (with note clarifying when to record it) and ADK does not have to populate it. I don't think this PR is blocked on it, but if we merge it as is, I'll send a follow up to clarify and add request model name with these caveats .

PR #202, opened 2026-05-27
Add gen_ai.agent.input.content.size and gen_ai.agent.output.content.size metrics

pvlsirotkin:

## Description

Adds two new GenAI semantic convention metrics for the byte size of content
a GenAI agent processes at the agent boundary, modeled on the recently-merged
`gen_ai.workflow.duration` metric
([#126](https://github.com/open-telemetry/semantic-conventions-genai/pull/126)):

- **`gen_ai.agent.input.content.size`** (Histogram, `By`): byte size of the
  content the agent receives at its entrypoint for a single invocation.
- **`gen_ai.agent.output.content.size`** (Histogram, `By`): byte size of the
  final response content the agent produces for a single invocation.

## Motivation

The existing GenAI conventions cover token usage on the **model** side
(`gen_ai.client.token.usage`) but provide no equivalent for the
**agent**-boundary content size. For non-text-only agents (image inputs,
audio outputs, multi-part tool results), token counts are also a poor proxy
for the actual bytes the agent processes.

User journey: capacity planning, identifying unusually large requests or
responses, detecting payload-size regressions across agent versions, and
correlating content size with latency or error rate.

## Checklist


docs/gen-ai/gen-ai-metrics.md
```diff
@@ -936,6 +939,113 @@ If there is no low-cardinality workflow name available for a given framework, th
...
+## Generative AI agent payload size metrics
+
+Individual systems may include additional system-specific attributes.
+It is recommended to check system-specific documentation, if available.
+
+`gen_ai.agent.request.size` and `gen_ai.agent.response.size` measure the
+serialized byte size of the input and output payloads of a single GenAI agent
+invocation. They are intended for instrumentations of agent frameworks (for
+example, ADK, LangChain agents, CrewAI agents) that can reliably observe the
+agent's input and final output.
```
trask: could we define this instead in terms of sizes of `gen_ai.input.messages` and `gen_ai.output.messages`, in order to make it more precise and comparable across frameworks?

schema-snapshot/registry.yaml
```diff
@@ -13834,6 +13927,97 @@ registry:
...
+      - 2025-05-01
+      key: gen_ai.agent.version
+      provenance:
+        path: ./model/gen-ai/registry.yaml
+      requirement_level:
+        conditionally_required: when available
+      stability: development
+      type: string
+    brief: GenAI agent request size.
+    instrument: histogram
```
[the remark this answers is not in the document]
trask: resolving, see #224

schema-snapshot/registry.yaml
```diff
@@ -13834,6 +13927,97 @@ registry:
...
+      - 2025-05-01
+      key: gen_ai.agent.version
+      provenance:
+        path: ./model/gen-ai/registry.yaml
+      requirement_level:
+        conditionally_required: when available
+      stability: development
+      type: string
+    brief: GenAI agent response size.
+    instrument: histogram
```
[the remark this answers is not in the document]
trask: resolving, see #224

model/gen-ai/metrics.yaml
```diff
@@ -113,6 +113,61 @@ metrics:
...
+    instrument: histogram
+    unit: "By"
+    stability: development
+    attributes:
+      - ref: gen_ai.agent.name
+        requirement_level:
+          conditionally_required: when available
+      - ref: gen_ai.agent.version
+        requirement_level:
+          conditionally_required: when available
```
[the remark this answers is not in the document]
pvlsirotkin: `when available` is the more common phrasing repo-wide (11 occurrences in `model/gen-ai/*.yaml` vs 2 for `If available.`), so I'll leave it as-is. Happy to standardize either way if maintainers have a preference — flagging for @trask / @lmolkova in case there's an existing style guideline I should follow.

model/gen-ai/metrics.yaml
```diff
@@ -113,6 +113,61 @@ metrics:
...
```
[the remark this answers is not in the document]
pvlsirotkin: Same as the [other thread](https://github.com/open-telemetry/semantic-conventions-genai/pull/202#discussion_r3311913316) — `when available` matches the more common repo-wide phrasing. Leaving as-is unless maintainers prefer otherwise.

docs/gen-ai/gen-ai-metrics.md
pvlsirotkin: Thought about this. The tricky bit is that `gen_ai.input.messages` at the agent boundary is the *cumulative* chat history (could be 1 turn or 50), whereas these are meant to measure the *per-invocation* increment — just the new user input on each turn and the agent's final response. Tying them to `input.messages` size would make the metric grow with conversation length and lose the per-turn anomaly signal.

Also a cost angle: `input.messages` is `opt_in`, so instrumentations would need to build and serialize the full JSON just to measure its length, even when `input.messages` itself isn't being captured.

Pushed a doc update to make the per-invocation scope explicit. Does that work, or would you rather pursue cross-framework comparability via a separate cumulative-size metric?

docs/gen-ai/gen-ai-metrics.md
```diff
@@ -936,6 +939,122 @@ If there is no low-cardinality workflow name available for a given framework, th
...
+  cumulative chat history that might be assembled into a model call.
+* `gen_ai.agent.response.size` is the byte size of the final response the
+  agent produces for this invocation, excluding intermediate tool calls or
+  reasoning steps.
+
+They are intended for instrumentations of agent frameworks (for example,
+ADK, LangChain agents, CrewAI agents) that can reliably observe the agent's
+input and final output.
+
+These metrics complement `gen_ai.agent.invocation.duration` by giving
```
MikeGoldsmith: `gen_ai.agent.invocation.duration` is not a known metric, did you mean `gen_ai.workflow.duration` or note the invoke_agent span's duration as the related signal

model/gen-ai/metrics.yaml
```diff
@@ -113,6 +113,61 @@ metrics:
...
+      content (for example, text plus inline binary data), the size SHOULD be
+      the sum of the byte lengths of each part.
+
+      This metric is intended for instrumentations of agent frameworks that
+      can reliably observe an agent's input payload (for example, ADK,
+      LangChain agents, CrewAI agents).
+    instrument: histogram
+    unit: "By"
+    stability: development
+    attributes:
```
MikeGoldsmith, line 146: Should we add `metric_attributes.gen_ai` and `attributes.gen_ai.error` refs here? I think other metrics use them.

model/gen-ai/metrics.yaml
```diff
@@ -113,6 +113,61 @@ metrics:
...
+      - ref: gen_ai.agent.name
+        requirement_level:
+          conditionally_required: when available
```
MikeGoldsmith: I don't think we need a condition, agent.name should always be available right?

docs/gen-ai/gen-ai-metrics.md
```diff
@@ -936,6 +939,122 @@ If there is no low-cardinality workflow name available for a given framework, th
...
+operators visibility into payload volume — useful for capacity planning,
+spotting unusually large per-turn requests or responses, and correlating
+size with latency or error rate.
+
+### Metric: `gen_ai.agent.request.size`
+
+This metric is [recommended][MetricRecommended] for instrumentations that
+can observe the input payload provided to an agent at invocation time.
+
+Instrumentations SHOULD record the size as the byte length of the serialized
```
MikeGoldsmith: I think we should be clear on how to calculate the byte length, eg UTF-8 byte length, JSON serialised, etc.

MikeGoldsmith (changes requested):

docs/gen-ai/gen-ai-metrics.md
```diff
@@ -936,6 +939,122 @@ If there is no low-cardinality workflow name available for a given framework, th
...
+request content as the agent receives it. For multi-part content (for example,
+text plus inline binary data), the size SHOULD be the sum of the byte lengths
+of each part.
```
lmolkova: If we go down this route, we should probably define more strictly what's counted (e.g. BlobPart.content length +  TextPart.content.length + ...)

docs/gen-ai/gen-ai-metrics.md
```diff
@@ -936,6 +939,122 @@ If there is no low-cardinality workflow name available for a given framework, th
...
+* `gen_ai.agent.response.size` is the byte size of the final response the
+  agent produces for this invocation, excluding intermediate tool calls or
+  reasoning steps.
+
+They are intended for instrumentations of agent frameworks (for example,
+ADK, LangChain agents, CrewAI agents) that can reliably observe the agent's
+input and final output.
+
+These metrics complement `gen_ai.agent.invocation.duration` by giving
+operators visibility into payload volume — useful for capacity planning,
```
lmolkova: the payload volume does not make much sense for in-process agents. Is it meant to be for remote calls / a2a ? If so, we can consider it for a2a.* namespace

docs/gen-ai/gen-ai-metrics.md
```diff
@@ -936,6 +939,122 @@ If there is no low-cardinality workflow name available for a given framework, th
...
+Individual systems may include additional system-specific attributes.
+It is recommended to check system-specific documentation, if available.
+
+`gen_ai.agent.request.size` and `gen_ai.agent.response.size` measure the
+serialized byte size of the **per-invocation** input and output of a single
+GenAI agent invocation:
+
+* `gen_ai.agent.request.size` is the byte size of the new content provided to
+  the agent for this invocation (typically the latest user message), not the
+  cumulative chat history that might be assembled into a model call.
```
lmolkova: how would agent know what is new? If caller passes 10 messages, what should instrumentation do with them?

Also the metric name does not describe this semantics:
- request assumes something physical like HTTP / gRPC call and in this case, agent gets a regular object (list of messages) in one of the parameters.
- request.size does not assume 'latest' / part of
- instrumentation would map this list into its own version of messages which will have a different serialized size  


A better name for this metric could be `gen_ai.agent.input.content.size`.

lmolkova (commented):
I left some questions and suggestions, but I'm curious what user scenarios or specific failure modes these metrics are targeting?

I'm trying to think what I can understand from them / alert on and the only thing that comes to mind is that drastic changes are not expected. But if they happen, it would be visible on underlying model token usage too, especially if we find means to stamp agent name on nested metrics.

model/gen-ai/metrics.yaml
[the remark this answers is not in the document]
trask: @pvlsirotkin thanks for flagging this! I've normalized these now in https://github.com/open-telemetry/semantic-conventions-genai/pull/245

docs/gen-ai/gen-ai-metrics.md
pvlsirotkin: Fair points on both. Plan:

- Rename to `gen_ai.agent.input.content.size` / `gen_ai.agent.output.content.size` — `input/output.content` is cleaner than `request/response` and doesn't imply HTTP/gRPC.
- Drop the "per-invocation new content" framing. The metric semantics become: byte size of the content the agent receives / produces *at the agent boundary*, whatever the framework sees natively. ADK measures the new `user_content` parameter; other frameworks would measure whatever their entrypoint receives. No requirement to distinguish "new" vs cumulative — frameworks just record what they observe.
- On the in-process case (your other comment): the primary use case is remote / a2a agents where content volume is obviously meaningful. But even for in-process agents content volume is useful for capacity planning, anomaly detection (e.g. a user pasting 10MB into a prompt), and sizing downstream services (token budgets, vector DB inputs, storage). Happy to keep this in `gen_ai.agent.*` for cross-framework consistency rather than splitting in-process vs a2a into separate namespaces.

Will push the rename + simplified semantics next.

docs/gen-ai/gen-ai-metrics.md
pvlsirotkin: Replied in the [thread above](https://github.com/open-telemetry/semantic-conventions-genai/pull/202#discussion_r3389133476) — in short: even for in-process agents the content volume is useful (capacity planning, anomaly detection, sizing downstream services), so proposing to keep this in `gen_ai.agent.*` rather than fragmenting into separate in-process vs a2a metrics.

docs/gen-ai/gen-ai-metrics.md
pvlsirotkin: Good catch — that was a forward reference to #201 which is also still in review. Removed the cross-reference

model/gen-ai/metrics.yaml
pvlsirotkin: Added `attributes.gen_ai.error` (error.type is genuinely useful here — did the agent fail to produce content?).

Held off on `metric_attributes.gen_ai` for now: it adds `provider.name`, `request.model`, and `server.address`/`server.port`, which feels mismatched for a content-size metric on a (often in-process) agent — the provider/model breakdowns are already captured on the child `gen_ai.inference` / `gen_ai.client.operation.duration` signals, and the address attributes don't apply to in-process agents. (Same reasoning we landed on for #201.) Happy to add later if a concrete use case comes up.

model/gen-ai/metrics.yaml
pvlsirotkin: Bumped to `recommended`. Stronger than the current `conditionally_required when available` (it means "emit by default, dashboards work out of the box"), but doesn't make frameworks with unnamed/inline agents noncompliant. Same compromise we landed on in [#201](https://github.com/open-telemetry/semantic-conventions-genai/pull/201#discussion_r3330218403). Happy to go all the way to `required` if you'd prefer.

docs/gen-ai/gen-ai-metrics.md
pvlsirotkin: The note now just says: "byte size of the content the agent receives/produces, as observed at the framework's entrypoint. Exact encoding is framework-defined; instrumentations SHOULD document what they count." This is a per-framework signal that's useful within a framework's data, not a cross-framework byte-comparable value.

cc @lmolkova who had the same request below.

docs/gen-ai/gen-ai-metrics.md
pvlsirotkin: Updated approach — see [Mike's thread](https://github.com/open-telemetry/semantic-conventions-genai/pull/202#discussion_r3389225861). Short version: dropped the "exclude framing bytes" rule because it's not portably implementable (different frameworks expose content very differently) and overpromises cross-framework comparability. The metric is now framework-defined: instrumentations record what they observe at the entrypoint and document what they count. Useful within a framework's data, not strictly comparable across frameworks.

model/gen-ai/metrics.yaml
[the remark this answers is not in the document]
pvlsirotkin: Moot after the latest revision — both `agent.name` references are now `recommended` (no condition string at all), so the phrasing inconsistency goes away. Also picking up the new sentence-case style from #245 anywhere I have `conditionally_required` notes elsewhere.

docs/gen-ai/gen-ai-metrics.md
pvlsirotkin: The metrics are now `gen_ai.agent.input.content.size` / `gen_ai.agent.output.content.size`, defined as the byte size of content the agent receives/produces at its entrypoint (no claim about new vs cumulative — frameworks just record what they observe). Byte-counting algorithm is spelled out concretely in the metric `note`: UTF-8 for text, raw bytes for binary

trask:

Create `changelog.d/202.enhancement.md` containing the change log entry, e.g.:

```md
Add `gen_ai.agent.input.content.size` and `gen_ai.agent.output.content.size` metrics to track the byte size of content the GenAI agent receives and produces at the agent boundary.
```

Thanks!

model/gen-ai/metrics.yaml
```diff
@@ -116,6 +116,70 @@ metrics:
...
+
+      Instrumentations SHOULD record the byte size of the content the
+      agent receives, as observed at the framework's entrypoint. The exact
+      encoding is framework-defined (for example, a framework that exposes
+      content as typed parts MAY sum the UTF-8 byte length of text parts
+      and the raw byte length of binary parts; a framework that handles
+      content as a serialized payload MAY use the byte length of that
+      serialization). Instrumentations SHOULD document what they count so
+      operators can interpret the values correctly within a given
+      framework.
```
lmolkova, line 142: we can formally define it based on message parts documented for input messages attribute in https://github.com/open-telemetry/semantic-conventions-genai/blob/main/docs/gen-ai/non-normative/models.py

This will be the same for different frameworks and will align nicely with https://github.com/open-telemetry/semantic-conventions-genai/pull/143 

(and the same comment for output messages)

model/gen-ai/metrics.yaml
```diff
@@ -116,6 +116,70 @@ metrics:
       - ref: gen_ai.workflow.name
         requirement_level:
           conditionally_required: If available.
+  - name: gen_ai.agent.input.content.size
+    annotations:
+      code_generation:
+        metric_value_type: int
+    brief: >
+      The byte size of the content the GenAI agent receives at the agent
+      boundary for a single invocation.
```
lmolkova, line 125: This is very ambiguous given that we mostly talk about in-process agents - they don't receive content, they get language-specific arguments that include content, config, flags, etc)

maybe

```suggestion
      The byte size of the content passed in input messages to the agent.
```

(and the same comment for output messages)

model/gen-ai/metrics.yaml
```diff
@@ -116,6 +116,70 @@ metrics:
...
+    note: |
+      Intended for instrumentations of agent frameworks (for example, ADK,
+      LangChain agents, CrewAI agents) that can observe the content passed
+      to the agent at its entrypoint. Useful for capacity planning,
+      anomaly detection (for example, a user pasting a very large prompt),
+      and sizing downstream services (token budgets, vector DB inputs,
+      storage).
```
lmolkova, line 132:
```suggestion
```

this is great in the PR description, but we don't usually put justification into the spec

docs/gen-ai/gen-ai-metrics.md
lmolkova: we can make cross-framework number relatively cheap to calculate now if we tie it to input and output sages. We can explicitly document the algorithm like

- iterate over parts (can optimize and not create otel version of parts if content is not being recorded)
- for each part:
  - if it's text do `len((TextPart.content).encode('utf-8'))` and add to the accumulating content size
  - if it's blob do `len(BlobPart.content)`
  - if it's reasoning part (who knows why someone gave a reasoning part to input - maybe its their bug, but we'll count it) do `len((ReasoningPart.content).encode('utf-8'))` 
  - if it's provider-specific part with variable content, it should also be summed up
  - ignore other parts

We'll need to update this algorithm when we add new parts with content.


Same with output - whatever is in the final output, we'll count it according to a similar algo, even if it somehow contains reasoning parts.

pvlsirotkin:

**State of the PR:**
- The metrics were renamed to `gen_ai.agent.input.content.size` / `gen_ai.agent.output.content.size` per @lmolkova's earlier feedback (`request/response` was misleading for in-process agents).
- 3 review threads remain open, all with the latest comment from @lmolkova:
  1. [Formal definition based on message parts in `non-normative/models.py`](https://github.com/open-telemetry/semantic-conventions-genai/pull/202#discussion_r3360367230) — @lmolkova suggested anchoring the byte-counting to the documented message-part schema; needs follow-up.
  2. [Ambiguity for in-process agents](https://github.com/open-telemetry/semantic-conventions-genai/pull/202#discussion_r3360377881) — @lmolkova noted in-process agents receive language-specific arguments (content + config + flags) rather than a content payload; she proposed an alternative wording.
  3. [Spec language vs PR description content](https://github.com/open-telemetry/semantic-conventions-genai/pull/202#discussion_r3360367803) — @lmolkova flagged that some of the justification in the prose belongs in the PR description, not the spec.

- These metrics are the most contested of the three (#201 / #202 / #203). The fundamental question @lmolkova raised is whether agent payload-size belongs in `gen_ai.*` at all vs. an `a2a.*` namespace for remote agents — defended in-process value (capacity planning, anomaly detection), but the SIG may push for a narrower scope.
- Byte-counting algorithm is currently "framework-defined; instrumentation SHOULD document what they count" — anchoring to the canonical message-part schema (her suggestion) would make it more comparable but adds implementation cost.

Thanks @MikeGoldsmith @lmolkova @trask for all the feedback so far.

pvlsirotkin:
Bytes is an out-of-place unit in the GenAI world.
Agent inputs/outputs size is not a useful measure agent cost. For example, an agent can be told to "Read a large file" and only this prompt would be recorded as "agent input". The following tool call, which actually loads the large file into the context wouldn't.
A more useful existing metric, gen_ai.client.token.usage already exists, and a more robust solution to measure agent cost, would be to add an "agent name" attribute to this metric. It would handle the "Read a large file" example from above.

PR #203, opened 2026-05-27
Add gen_ai.agent.steps metric

pvlsirotkin:

## Description

Adds a new GenAI semantic convention metric for the number of steps a GenAI
agent takes during a single invocation, modeled on the recently-merged
`gen_ai.workflow.duration` metric
([#126](https://github.com/open-telemetry/semantic-conventions-genai/pull/126)):

- **`gen_ai.agent.steps`** (Histogram, `{step}`): number of discrete steps a
  GenAI agent takes during a single invocation.

The metric counts only steps the agent itself authored (e.g. in ADK, events
whose `event.author` equals the agent name). Tool results are not counted
(authored by the tool, not the agent). Sub-agent steps are not counted (each
step is recorded exactly once across the full call tree). Failed and partial
steps still count — they consumed the agent's execution.

Step counts MUST NOT be compared across different frameworks; instrumentations
MUST document what counts as a step for their framework.

## Motivation

The existing GenAI conventions describe agents and tools but offer no
standardized way to answer "how many steps did this agent take to produce
its response?" Operators today derive that figure ad-hoc from span counts,
which is fragile and depends on sampling configuration.

User journey: capacity planning, regression detection across agent versions,
and diagnosing pathological tool loops (an agent stuck calling tools without
making progress).

## Checklist


model/gen-ai/metrics.yaml
```diff
@@ -113,6 +113,40 @@ metrics:
...
+    annotations:
+      code_generation:
+        metric_value_type: int
+    brief: 'Number of steps in a single GenAI workflow.'
+    note: |
+      This metric measures the number of discrete steps observed within a
+      single GenAI workflow. A "step" is intentionally framework-defined: it
+      typically corresponds to a single observable event emitted within the
+      workflow (for example, a tool call, a tool result, or a model response
+      in ADK; a node execution in LangGraph).
```
trask: It would be helpful if we can define a "step" more concretely, even if it's a medium sized list of things that it covers.

If it has to be framework specific, we may want to think about adding `gen_ai.framework.name` and framework specific conventions to define it concretely for each of the common frameworks.

It feels a bit vague at this point which will lead to lots of questions when people try to implement it.

pvlsirotkin: Good point. Pushed an update that lists what counts as a step in the four most common frameworks (ADK events, LangGraph nodes, LangChain `AgentAction`/`AgentFinish`, CrewAI agent turns) and adds a `MUST document` requirement for instrumentations.

model/gen-ai/metrics.yaml
```diff
@@ -113,6 +113,40 @@ metrics:
...
+    attributes:
+      - ref: gen_ai.workflow.name
+        requirement_level:
+          conditionally_required: If available.
+      - ref: gen_ai.agent.name
+        requirement_level:
+          conditionally_required: when available
+      - ref: gen_ai.agent.version
+        requirement_level:
+          conditionally_required: when available
```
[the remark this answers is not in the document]
pvlsirotkin: Fair point — standardized all three entries in this metric to `If available.` to match the adjacent `gen_ai.workflow.duration` entry.

docs/gen-ai/gen-ai-metrics.md
```diff
@@ -936,6 +943,74 @@ If there is no low-cardinality workflow name available for a given framework, th
...
+visibility into how "chatty" or compute-intensive a workflow is, and
+is useful for capacity planning and regression detection across
+versions.
+
+**Attributes:**
+
+| Key | Stability | [Requirement Level](https://opentelemetry.io/docs/specs/semconv/general/attribute-requirement-level/) | Value Type | Description | Example Values |
+| --- | --- | --- | --- | --- | --- |
+| [`gen_ai.agent.name`](/docs/registry/attributes/gen-ai.md) | ![Development](https://img.shields.io/badge/-development-blue) | `Conditionally Required` If available. | string | Human-readable name of the GenAI agent provided by the application. | `Math Tutor`; `Fiction Writer` |
+| [`gen_ai.agent.version`](/docs/registry/attributes/gen-ai.md) | ![Development](https://img.shields.io/badge/-development-blue) | `Conditionally Required` If available. | string | The version of the GenAI agent. | `1.0.0`; `2025-05-01` |
```
MikeGoldsmith: As a freeform string, that has potential to have an unbounded cardinality. Do we need it on a metric? Could it added to spans where cardinality is generally more acceptable?

docs/gen-ai/gen-ai-metrics.md
```diff
@@ -936,6 +943,74 @@ If there is no low-cardinality workflow name available for a given framework, th
 <!-- END AUTOGENERATED TEXT -->
 <!-- endweaver -->
 
+### Metric: `gen_ai.workflow.steps`
+
+This metric is [recommended][MetricRecommended] for instrumentations that
+can reliably attribute steps (events) to a specific workflow or agent.
+
+A "step" is intentionally framework-defined; it typically corresponds to a
```
MikeGoldsmith: As the step is per-framework, I think we need stronger wording to say step counts MUST NOT be compared between frameworks.

model/gen-ai/metrics.yaml
```diff
@@ -113,6 +113,48 @@ metrics:
...
+      When the `gen_ai.agent.name` dimension is set, instrumentations SHOULD
+      only count steps attributed to that agent (for example, events
+      authored by that agent); events attributed to sub-agents or to a
+      parent workflow SHOULD NOT be counted.
+
+      This metric complements `gen_ai.workflow.duration` by giving operators
+      visibility into how "chatty" or compute-intensive a workflow is, and
+      is useful for capacity planning and regression detection across
+      versions.
+    instrument: histogram
```
MikeGoldsmith, line 155: Can we add the histogram buckets you have in the metric description above?

model/gen-ai/metrics.yaml
```diff
@@ -113,6 +113,48 @@ metrics:
...
+    unit: "{step}"
+    stability: development
+    attributes:
```
MikeGoldsmith, line 158: Probably should have the `attributes.gen_ai.error` ref here too.

docs/gen-ai/gen-ai-metrics.md
```diff
@@ -936,6 +943,74 @@ If there is no low-cardinality workflow name available for a given framework, th
...
+single observable event emitted within the workflow (for example, a tool
+call, a tool result, or a model response in ADK; a node execution in
+LangGraph). When the `gen_ai.agent.name` dimension is set, instrumentations
+SHOULD only count steps attributed to that agent; events attributed to
```
MikeGoldsmith: What does "attributed to" mean? Authored by, caused by, etc?

What happens with tool calls?

model/gen-ai/metrics.yaml
```diff
@@ -113,6 +113,48 @@ metrics:
       - ref: gen_ai.workflow.name
         requirement_level:
           conditionally_required: If available.
+  - name: gen_ai.workflow.steps
+    annotations:
+      code_generation:
+        metric_value_type: int
+    brief: 'Number of steps in a single GenAI workflow.'
+    note: |
```
MikeGoldsmith, line 125: Do we only count completed steps, or also include partial / failed steps too?

MikeGoldsmith (changes requested):

docs/gen-ai/gen-ai-metrics.md
```diff
@@ -936,6 +943,74 @@ If there is no low-cardinality workflow name available for a given framework, th
...
+
+When the `gen_ai.agent.name` dimension is set, instrumentations SHOULD
+only count steps attributed to that agent (for example, events
+authored by that agent); events attributed to sub-agents or to a
+parent workflow SHOULD NOT be counted.
+
+This metric complements `gen_ai.workflow.duration` by giving operators
+visibility into how "chatty" or compute-intensive a workflow is, and
+is useful for capacity planning and regression detection across
+versions.
```
lmolkova: assuming we were able to stamp workflow name on all nested spans and metrics emitted in the context of this workflow, we'd had

- count of agent invocations per workflow name
- count of tool calls per workflow name
- .. (any other metrics, e.g. if we instrumented transfers, we'd had a count of transfers)

Would this metric be still necessary?

In the current shape this metric seems to be an aggregation of multiple independent things which maybe also queried and sumed up individually, but then would provide better observability.

docs/gen-ai/gen-ai-metrics.md
```diff
@@ -936,6 +943,74 @@ If there is no low-cardinality workflow name available for a given framework, th
...
+
+**Attributes:**
+
+| Key | Stability | [Requirement Level](https://opentelemetry.io/docs/specs/semconv/general/attribute-requirement-level/) | Value Type | Description | Example Values |
+| --- | --- | --- | --- | --- | --- |
+| [`gen_ai.agent.name`](/docs/registry/attributes/gen-ai.md) | ![Development](https://img.shields.io/badge/-development-blue) | `Conditionally Required` If available. | string | Human-readable name of the GenAI agent provided by the application. | `Math Tutor`; `Fiction Writer` |
```
lmolkova: Does it mean that it should be recorded per each invoke_agent? Wokflow duration is currently defined as 'not invoke agent' (maybe we should revisit it?) and does not have agent name / version / etc attributes

docs/gen-ai/gen-ai-metrics.md
```diff
@@ -936,6 +943,74 @@ If there is no low-cardinality workflow name available for a given framework, th
...
+| -------- | --------------- | ----------- | -------------- | --------- | ------ |
+| `gen_ai.workflow.steps` | Histogram | `{step}` | Number of steps in a single GenAI workflow. [1] | ![Development](https://img.shields.io/badge/-development-blue) | |
+
+**[1]:** This metric measures the number of discrete steps observed within a
+single GenAI workflow. A "step" is a framework-defined unit of
+execution that corresponds to a single observable event within the
+workflow. Common framework-specific definitions:
+
+- **ADK**: each event in the agent's invocation context (model
+  response, tool call, tool result, agent transfer).
```
lmolkova: does it make sense to call tool result or agent transfers as steps? presumably count of tool results if the same as count of tool calls.

model/gen-ai/metrics.yaml
```diff
@@ -113,6 +113,48 @@ metrics:
...
+      This metric measures the number of discrete steps observed within a
+      single GenAI workflow. A "step" is a framework-defined unit of
+      execution that corresponds to a single observable event within the
+      workflow. Common framework-specific definitions:
+
+      - **ADK**: each event in the agent's invocation context (model
```
aabmass: Would it be useful to link https://adk.dev/runtime/event-loop/#event?

model/gen-ai/metrics.yaml
```diff
@@ -113,6 +113,48 @@ metrics:
...
+      visibility into how "chatty" or compute-intensive a workflow is, and
+      is useful for capacity planning and regression detection across
+      versions.
+    instrument: histogram
+    unit: "{step}"
+    stability: development
+    attributes:
+      - ref: gen_ai.workflow.name
+        requirement_level:
+          conditionally_required: If available.
```
aabmass: Assuming this one should be required since the metric is called "workflow"

model/gen-ai/metrics.yaml
```diff
@@ -113,6 +113,48 @@ metrics:
...
+      - **LangChain agents**: each `AgentAction` or `AgentFinish` step.
+      - **CrewAI**: each agent turn within a task.
+
+      Instrumentations MUST document what counts as a step for their
+      framework.
+
+      When the `gen_ai.agent.name` dimension is set, instrumentations SHOULD
+      only count steps attributed to that agent (for example, events
+      authored by that agent); events attributed to sub-agents or to a
+      parent workflow SHOULD NOT be counted.
```
aabmass: I think the same applies to workflow too, we shouldn't count nested workflows/

A simpler invariant might "each step should only be recorded once"

docs/gen-ai/gen-ai-metrics.md
aabmass: I think this would be low cardinality though, similar to service.* resource attributes

aabmass (approved):

pvlsirotkin:
Big update on this PR after re-reading the original RFC and the production ADK implementation: the metric was always meant to count steps a specific *agent* takes during a single invocation, not steps in some larger workflow. An earlier revision of this PR pushed it into the `gen_ai.workflow.*` namespace, which turned out to be the source of most of the confusion in this review round (per-invocation vs cumulative, whether to count sub-agents, whether `workflow.name` should be required, whether this is just an aggregation of child counts, etc.).

Just pushed a commit that:

- Renames the metric to `gen_ai.agent.steps` (out of the workflow namespace).
- Drops `gen_ai.workflow.name` and `gen_ai.agent.version` from the dimensions. The only dimension is `gen_ai.agent.name` (at `recommended` level, matching #201/#202).
- Adds `attributes.gen_ai.error` (per @MikeGoldsmith).
- Rewrites the metric `note` to spell out the per-agent semantics:
  - Counts only events the agent itself authored.
  - Tool results are not counted (authored by the tool, not the agent).
  - Sub-agent steps are not counted; each step recorded exactly once across the call tree (@aabmass' invariant).
  - Failed and partial steps still count (they consumed agent execution time).
  - Step counts MUST NOT be compared across frameworks (@MikeGoldsmith's ask for stronger wording).
  - Concrete framework definitions for ADK, LangGraph, LangChain, CrewAI, with a link to https://adk.dev/runtime/event-loop/#event (@aabmass' suggestion).
  - Histogram bucket recommendation included in the note (@MikeGoldsmith's ask).

The MD page got a new `## Generative AI agent metrics` section so the metric no longer lives under "Generative AI workflow metrics". MD body is kept thin; the full semantics live in the YAML `note` and are rendered into the autogen block.

Replying inline on each open thread.

model/gen-ai/metrics.yaml
pvlsirotkin: Revisited this with the refocus on per-agent scope (see [summary](https://github.com/open-telemetry/semantic-conventions-genai/pull/203#issuecomment-4671740422)). The 'step' definition is now anchored to ADK's `event.author` model with concrete examples for the other common frameworks. Skipping `gen_ai.framework.name` for now — keeping the metric framework-defined but with the explicit MUST-NOT-compare-across-frameworks wording @MikeGoldsmith asked for.

model/gen-ai/metrics.yaml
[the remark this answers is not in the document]
pvlsirotkin: Moot after the refocus — `gen_ai.agent.name` is now `recommended` (no condition string), and `workflow.name` / `agent.version` were dropped. No `conditionally_required` notes remain on this metric.

docs/gen-ai/gen-ai-metrics.md
pvlsirotkin: Side-stepped by the refocus: `gen_ai.workflow.name` is no longer on this metric. Only `gen_ai.agent.name` remains, which is low cardinality (similar to what @aabmass noted).

docs/gen-ai/gen-ai-metrics.md
pvlsirotkin: Added explicit wording to the metric note: 'Step counts MUST NOT be compared across different frameworks. They are only meaningful within the context of a single framework's instrumentation.'

model/gen-ai/metrics.yaml
pvlsirotkin: Moved the bucket recommendation into the YAML `note` so it shows up in the autogen block, not just the MD prose.

model/gen-ai/metrics.yaml
pvlsirotkin: Added.

docs/gen-ai/gen-ai-metrics.md
pvlsirotkin: Spelled out as 'authored by' with the ADK `event.author` model as the concrete anchor. Tool calls authored by the agent ARE counted; tool results (authored by the tool, not the agent) are NOT counted. Both are explicit in the note now.

model/gen-ai/metrics.yaml
pvlsirotkin: New answer: tbc
Old answer: Both. The note now says: 'Steps are counted regardless of outcome: a step that errored or partially completed still consumed the agent's execution and SHOULD be counted.'

docs/gen-ai/gen-ai-metrics.md
pvlsirotkin: With the refocus to per-agent scope (see [summary](https://github.com/open-telemetry/semantic-conventions-genai/pull/203#issuecomment-4671740422)) this is no longer a workflow-level aggregation. It's a per-invocation count of an agent's own steps — distinct from summing child signals because (a) not every event is independently traced, and (b) operators get a single histogram they can directly compute percentiles on rather than needing to sum multiple counters and join by attribute.

docs/gen-ai/gen-ai-metrics.md
pvlsirotkin: Good question. With the per-agent refocus the answer is concrete: only events the agent itself authored count. Tool calls (authored by agent) — yes. Tool results (authored by tool) — no, so no double-counting. Agent transfers (initiated by the agent) — yes. Both rules now explicit in the metric note.

model/gen-ai/metrics.yaml
pvlsirotkin: Added the link in the ADK example in the metric note. Thanks for the pointer.

model/gen-ai/metrics.yaml
pvlsirotkin: Side-stepped by the rename — the metric is now `gen_ai.agent.steps` and `gen_ai.workflow.name` was dropped.

model/gen-ai/metrics.yaml
pvlsirotkin: Adopted your invariant verbatim in the note: 'each step SHOULD be recorded exactly once across the full call tree' — sub-agent and surrounding-workflow steps are excluded.

docs/gen-ai/gen-ai-metrics.md
pvlsirotkin: changed the metrics from workflow to agent, so not relevant anymore I believe

trask:

Create `changelog.d/203.enhancement.md` containing the change log entry, e.g.:

```md
Add `gen_ai.agent.steps` metric to track the number of steps a GenAI agent takes during a single invocation.
```

Thanks!

docs/gen-ai/gen-ai-metrics.md
```diff
@@ -931,6 +933,81 @@ If there is no low-cardinality workflow name available for a given framework, th
...
+- **ADK**: each [event](https://adk.dev/runtime/event-loop/#event) in
+  the invocation context whose `event.author` equals the agent name
+  (typically a model response, a tool call the agent issued, or an
+  agent transfer the agent initiated). Tool results, which are
+  authored by the tool rather than the agent, are not counted.
+- **LangGraph**: each node execution the agent itself performed in
+  the workflow graph.
+- **LangChain agents**: each `AgentAction` or `AgentFinish` step the
+  agent itself emitted.
+- **CrewAI**: each agent turn the agent itself took within a task.
```
lmolkova, line 973: I did a bit of a research and I think this generalization would be confusing, it does not make sense to merge these things into one metric unless they are comparable:

1. For ADK it would count LLM and tool calls, handoffs 
2. Langgraph has a notion of super-step which is ["considered a single iteration over the graph nodes"](https://docs.langchain.com/oss/python/langgraph/graph-api#graphs) (active ones). So by definition it's more than one node and `step` used in the metric name would be misleading. Langgraph also has configurable `recursion_limit` for number of super-steps and  `remaining_steps` in the step that provides number of remaining ones until limit is reached. This is a signal that these things could be meaningful to LangGraph users. 
3. I can't find a term 'turn' in crewai, my AI found `iterations` in agent executor - https://github.com/crewAIInc/crewAI/blob/b6e5d632c111e098d883274f0a26f02836a2172e/lib/crewai/src/crewai/agents/crew_agent_executor.py#L218 and there is a corresponding config for max_interations, but iteration is a full ReAct loop. It has something called `step_callback` but does not clearly document when it's called.


I think this metric tries to capture something that's a proxy to number of loops / iterations within one agent, but tries to express it through internal frameowrk implementation details.

Some alternative approaches:

0. Stamp agent name on the underlying llm and tool calls and derive average number of calls per agent from `gen_ai.inference | execute_tool.duration`  and invoke agent duration. This won't give outliers or percentiles, but arguably  many loops / events also leads to long invoke_agent duration.
1. Measure `gen_ai.agent.inference_calls_per_invocation`, `gen_ai.agent.tool_calls_per_invocations`, `gen_ai.agent.transfers_per_invocation` - these are understandable and generically applicable. The list can probably grow with retrievals and memory calls
2. Have specialized metrics for frameworks. `[gcp.]adk.event_loop.event.count` with attributes for author and type or `[gcp].adk.event_loop.events_per_invocation` would be specific and understandable. Langgraph and crewai would have specialized metrics too 

Approaches 1 and 2 can exist together: `gen_ai.agent.inference_calls_per_invocation` could be a common 'give me a sense of how long this agent invocation was' that shows up on common dashboards and  `[gcp].adk.event_loop.events_per_invocation` can be a detailed breakdown. This would align with [t-shaped signals](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/how-to-write-conventions/t-shaped-signals.md) approach.

pvlsirotkin:

**State of the PR:**
- The metric was renamed and refocused: `gen_ai.workflow.steps` → `gen_ai.agent.steps`, scoped to counting steps a specific agent itself authored during one invocation (per the underlying RFC and the production ADK implementation). The earlier "workflow" framing caused most of the confusion in prior review rounds.
- 3 review threads remain open:
  1. [Generalization concern](https://github.com/open-telemetry/semantic-conventions-genai/pull/203#discussion_r3365690525) — @lmolkova thinks the metric as currently defined isn't comparable across frameworks because what counts as a "step" varies. Needs follow-up.
  2. [Workflow-level aggregation alternative](https://github.com/open-telemetry/semantic-conventions-genai/pull/203#discussion_r3360175878) — @lmolkova suggested an alternative: if `workflow.name` could be stamped on all nested signals, this metric might be derivable by query rather than emitted directly. My response defended the per-agent metric as still useful — awaiting her take.
  3. [Completed vs partial/failed steps](https://github.com/open-telemetry/semantic-conventions-genai/pull/203#discussion_r3356657040) — my reply has a `New answer: tbc / Old answer:...` artifact from an edit conflict; @RKest should clean this up and confirm the answer.

- The metric counts events the agent itself authored (e.g. ADK events where `event.author == agent_name`). Tool results are NOT counted (authored by tool). Sub-agent steps are NOT counted (each step once across the call tree).
- Step counts MUST NOT be compared across frameworks; instrumentations MUST document what they count.
- This was the source of a lot of cross-talk between @trask, @MikeGoldsmith, @lmolkova, @aabmass — recommend reading the [top-level summary comment](https://github.com/open-telemetry/semantic-conventions-genai/pull/203#issuecomment-4671740422) first for context.

Thanks everyone for the feedback so far.

Necmttn:
Step counts will be useful if the spec separates the unit from comparability.

I would make each instrumentation publish `gen_ai.agent.step.definition` or a documented enum such as `authored_event|llm_turn|planner_iteration|framework_step`, with the metric still named `gen_ai.agent.steps`. Then dashboards can compare versions within the same framework safely, while cross-framework views know when step counts are not comparable. Tests should include tool result and subagent cases that are intentionally not counted.

---

_Generated with [ax](https://github.com/Necmttn/ax)._

RKest:
> Some alternative approaches:
> ...
> 1. Measure gen_ai.agent.inference_calls_per_invocation, gen_ai.agent.tool_calls_per_invocations, gen_ai.agent.transfers_per_invocation - these are understandable and generically applicable. The list can probably grow with retrievals and memory calls
> ...

I went ahead and implemented this alternative in https://github.com/open-telemetry/semantic-conventions-genai/pull/336

lmolkova:
I'm going to close this PR since it's continued in #336, feel free to reopen if necessary

PR #211, opened 2026-05-28
Clarify that billed token counts should be reported for Cohere usage

### LINKED ISSUE #13: GenAI: document that usage tokens for cohere are captured differently
Cohere reports billable tokens separately from usage prompt/completion tokens.

We need to document it.

trask:
Cohere reports both `usage.tokens` (model-consumed) and `usage.billed_units` (customer-billed). Add a note to `gen_ai.usage.input_tokens` and `gen_ai.usage.output_tokens` that instrumentations SHOULD report the billed count when a provider reports both, and update the Cohere reference scenario to use `billed_units`.

Closes #13


lmolkova (approved):

reference/scenarios/cohere/scenario.py
```diff
@@ -44,11 +44,11 @@ def run_chat(client):
             span.set_attribute("gen_ai.response.id", resp.id)
         if hasattr(resp, "finish_reason") and resp.finish_reason:
             span.set_attribute("gen_ai.response.finish_reasons", [resp.finish_reason])
-        if hasattr(resp, "usage") and resp.usage and hasattr(resp.usage, "tokens") and resp.usage.tokens:
-            if hasattr(resp.usage.tokens, "input_tokens"):
-                span.set_attribute("gen_ai.usage.input_tokens", int(resp.usage.tokens.input_tokens))
-            if hasattr(resp.usage.tokens, "output_tokens"):
-                span.set_attribute("gen_ai.usage.output_tokens", int(resp.usage.tokens.output_tokens))
+        if hasattr(resp, "usage") and resp.usage and hasattr(resp.usage, "billed_units") and resp.usage.billed_units:
```
MikeGoldsmith, line 47: Should the scenario fall back to `usage.tokens` when `usage.billed_units` is absent? Real responses may report only one. If the spec intent is "billed if present, else consumed", worth showing that here (and saying so in the registry note).

MikeGoldsmith (approved):

reference/scenarios/cohere/scenario.py
trask: > Real responses may report only one

Is this true for Cohere?

MikeGoldsmith: I'm not sure, the PR description says Cohere sends both. I'm unsure whether it sends one or the other, or potentially both at the same time. If it's one or the other, a fallback wouldn't be necessary.

trask: As far as I know, Cohere always sends both and so no need for a fallback.

MikeGoldsmith:
I think this is ready for merge 👍🏻

Issue #178, opened 2026-05-19
Clarify gen_ai.provider.name semantics for invoke_agent spans in framework instrumentations


wrisa:
**Context**

  The `gen_ai.provider.name` attribute is currently defined as identifying the
  GenAI *inference* provider (openai, aws.bedrock, anthropic, etc.). The registry
  enum lists only LLM/embedding service providers.

  The `invoke_agent` span definition in gen-ai-agent-spans.md marks
  `gen_ai.provider.name` as **Required**, but does not clarify what the value
  should be when the agent span is emitted by a *framework* instrumentation
  (LangChain, LangGraph, CrewAI, etc.) rather than by a provider SDK.

  **The ambiguity**

  In a LangGraph/LangChain instrumentation, an `invoke_agent` span is created
  when a LangGraph graph or ReAct agent is invoked. At that point the
  "provider" could reasonably mean:

  1. **The framework** — e.g. `langgraph`, since that is what is actually
     being invoked. But `langgraph` is not a defined enum value in the registry,
     and frameworks are not inference providers.

  2. **The underlying LLM provider** — e.g. `openai`, available via LangChain's
     `ls_provider` metadata. But this leaks implementation details of the inner
     LLM call onto the outer agent span, and if there are multiple models (e.g.
     a planner using `anthropic` and a tool-caller using `openai`) the value is
     ambiguous.

  3. **Omit it entirely** — treat `gen_ai.provider.name` as not applicable for
     pure-framework agent spans and relax the Required level for `invoke_agent`.
     The framework identity could instead be expressed via `telemetry.sdk.name`
     or a new attribute.

  **Concrete impact**

  The `opentelemetry-instrumentation-langchain` package currently sets
  `gen_ai.provider.name` on `invoke_agent` spans to whatever `ls_provider` the
  LangChain metadata carries (e.g. `openai`). This means the agent span and all
  `chat` child spans carry the same provider value, providing no differentiation,
  and is arguably incorrect — the agent span is not an OpenAI call.

  **Questions for the spec**

  - Should `gen_ai.provider.name` on `invoke_agent` represent the *framework*
    or the *backing LLM provider*?
  - If it should be the framework, should `langgraph`, `langchain`, `crewai`,
    etc. be added as defined enum values?
  - If the semantics differ from inference spans, should the attribute
    description for agent spans be updated to clarify?
  - Should the requirement level for `gen_ai.provider.name` on `invoke_agent`
[...]

PR #212, opened 2026-05-28
Generalize the `gen_ai.provider.name` description

trask:
Generalize the `gen_ai.provider.name` description: remove inference-centric paragraphs, and add guidance for per-operation semantic conventions to clarify which kinds of providers apply when it is not clear from context.

Related to #178


model/gen-ai/spans.yaml
```diff
@@ -218,6 +218,27 @@ spans:
...
+          attributes may help identify the actual system in use.
+
+          The `gen_ai.provider.name` attribute acts as a discriminator that
+          identifies the GenAI telemetry format flavor specific to that provider
+          within GenAI semantic conventions.
+          It SHOULD be set consistently with provider-specific attributes and signals.
+          For example, GenAI spans, metrics, and events related to AWS Bedrock
+          should have the `gen_ai.provider.name` set to `aws.bedrock` and include
+          applicable `aws.bedrock.*` attributes and are not expected to include
+          `openai.*` attributes.
```
trask: @lmolkova maybe we need a weaver feature to prepend to the inherited note 😅

trask: went a different route

lmolkova (approved):

MikeGoldsmith (approved):
Would it be worth a follow-up that re-adds the "look at `gen_ai.request.model`, `gen_ai.response.model`, `server.address` to identify the actual system" pointer in per-operation specs so it isn't lost?

trask:
> Would it be worth a follow-up that re-adds the "look at `gen_ai.request.model`, `gen_ai.response.model`, `server.address` to identify the actual system" pointer in per-operation specs so it isn't lost?

I didn't really understand that pointer, can you explain a bit more detail how it would be used?

MikeGoldsmith:
> > Would it be worth a follow-up that re-adds the "look at `gen_ai.request.model`, `gen_ai.response.model`, `server.address` to identify the actual system" pointer in per-operation specs so it isn't lost?
> 
> I didn't really understand that pointer, can you explain a bit more detail how it would be used?

The following hint was used a few times and was removed. I wasn't very clear, I meant to ask: do you think it's worth keeping that hint somewhere?

```
The `gen_ai.request.model`, `gen_ai.response.model`, and `server.address`
attributes may help identify the actual system in use.
```

trask:
It didn't really understand the sentence, so seemed ok to me to remove it, but if you have a better understanding of how it is useful, I can try to find a better place / phrasing for it.

MikeGoldsmith:
I think it's okay to remove 👍🏻

Issue #93, opened 2026-04-16
GenAI Client Duration Metric has required provider attribute


keith-decker:
### Area(s)

area:gen-ai

### What's the problem?

The duration metric for GenAI client duration metric at https://github.com/open-telemetry/semantic-conventions/blob/main/docs/gen-ai/gen-ai-metrics.md#metric-gen_aiclientoperationduration lists provider as required. When implementing the metrics for Execute Tools in GenAI Utils for python, it's become apparant that provider is not always available, since tool calls may not make calls to LLM providers.

### Describe the solution you'd like

Perhaps setting that attribute to Conditionally Required may be the most prudent fix?

### Tip

<sub>[React](https://github.blog/news-insights/product-news/add-reactions-to-pull-requests-issues-and-comments/) with 👍 to help prioritize this issue. Please use comments to provide useful context, avoiding `+1` or `me too`, to help us triage it. Learn more [here](https://opentelemetry.io/community/end-user/issue-participation/).</sub>

Nik-Reddy:
@keith-decker I'd like to take this on. I noticed this while working on GenAI metric attributes in this repo #96, `gen_ai.provider.name` is required via the shared `metric_attributes.gen_ai` group, but operations like tool execution don't always involve a GenAI provider.

Since that shared group is referenced by all 8 GenAI metrics, relaxing it globally would also weaken metrics where provider is always known (token usage, TTFT, server-side metrics). 

My inclination is to scope the fix to `gen_ai.client.operation.duration` only, overriding `gen_ai.provider.name` to conditionally_required, ie required when the operation involves a GenAI provider call, which is the lowest-blast-radius approach.

Does that scoping sound right, or should provider be relaxed more broadly across client metrics? Happy to open a PR if this direction makes sense.

trask:
Related to https://github.com/open-telemetry/semantic-conventions-genai/issues/178

trask:
> When implementing the metrics for Execute Tools

I believe tool executions are internal operations (as opposed to client operations), and so shouldn't be included in `gen_ai.client.operation.duration`. I sent https://github.com/open-telemetry/semantic-conventions-genai/pull/215 to hopefully clarify.

trask:
Resolved by #214

(heads up there may be further changes to this metric in #215)

PR #215, opened 2026-05-29
Clarify scope of `gen_ai.client.operation.duration` metric

trask:
Related to #93

Clarify that `gen_ai.client.operation.duration` covers GenAI client operations such as inference, embeddings, retrieval, memory, agent creation, and remote agent invocation, and should not be recorded for internal operations such as tool execution, workflow invocation, planning, or in-process agent invocation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,7 +32,8 @@
 
 ### 📚 Clarifications 📚
 
-- Clarify that `gen_ai.client.operation.duration` covers GenAI client operations such as
-  inference, embeddings, retrieval, memory, agent creation, and remote agent invocation,
-  and should not be recorded for internal operations such as tool execution, workflow
-  invocation, planning, or in-process agent invocation.
+- Clarify that `gen_ai.client.operation.duration` corresponds to the GenAI inference,
+  embeddings, retrieval, memory, create agent, and invoke agent client spans, and that
+  when the metric is reported alongside one of these spans its value should be the same
+  as the span duration.
+  ([#215](https://github.com/open-telemetry/semantic-conventions-genai/pull/215))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,3 +31,8 @@
   ([#136](https://github.com/open-telemetry/semantic-conventions-genai/pull/136))
 
 ### 📚 Clarifications 📚
+
+- Clarify that `gen_ai.client.operation.duration` covers GenAI client operations such as
+  inference, embeddings, retrieval, memory, agent creation, and remote agent invocation,
+  and should not be recorded for internal operations such as tool execution, workflow
+  invocation, planning, or in-process agent invocation.
--- a/model/gen-ai/metrics.yaml
+++ b/model/gen-ai/metrics.yaml
@@ -33,7 +33,12 @@ metrics:
   - name: gen_ai.client.operation.duration
     instrument: histogram
     unit: "s"
-    brief: 'GenAI operation duration.'
+    brief: 'GenAI client operation duration.'
+    note: >
+      This metric covers GenAI client operations such as inference, embeddings, retrieval,
+      memory, agent creation, and remote agent invocation. It SHOULD NOT be recorded for
+      internal operations such as tool execution, workflow invocation, planning, or
+      in-process agent invocation.
     stability: development
     annotations:
       code_generation:
--- a/model/gen-ai/metrics.yaml
+++ b/model/gen-ai/metrics.yaml
@@ -34,11 +34,17 @@ metrics:
     instrument: histogram
     unit: "s"
     brief: 'GenAI client operation duration.'
-    note: >
-      This metric covers GenAI client operations such as inference, embeddings, retrieval,
-      memory, agent creation, and remote agent invocation. It SHOULD NOT be recorded for
-      internal operations such as tool execution, workflow invocation, planning, or
-      in-process agent invocation.
+    note: |
+      This metric corresponds to the GenAI client spans:
+      [Inference](/docs/gen-ai/gen-ai-spans.md#inference),
+      [Embeddings](/docs/gen-ai/gen-ai-spans.md#embeddings),
+      [Retrieval](/docs/gen-ai/gen-ai-spans.md#retrievals),
+      [Memory](/docs/gen-ai/gen-ai-spans.md#memory),
+      [Create Agent](/docs/gen-ai/gen-ai-agent-spans.md#create-agent-span),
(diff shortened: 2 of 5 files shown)


model/gen-ai/metrics.yaml
```diff
@@ -33,7 +33,12 @@ metrics:
   - name: gen_ai.client.operation.duration
     instrument: histogram
     unit: "s"
-    brief: 'GenAI operation duration.'
+    brief: 'GenAI client operation duration.'
+    note: >
+      This metric covers GenAI client operations such as inference, embeddings, retrieval,
+      memory, agent creation, and remote agent invocation. It SHOULD NOT be recorded for
+      internal operations such as tool execution, workflow invocation, planning, or
+      in-process agent invocation.
```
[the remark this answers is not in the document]
trask: Fixed in https://github.com/open-telemetry/semantic-conventions-genai/commit/981640a.

Reworded the note to tie the metric to the corresponding GenAI client spans (Inference, Embeddings, Retrieval, Memory, Create Agent, Invoke Agent Client) with links to each span definition, and added the standard "metric value SHOULD be the same as the span duration" wording used by other OTel duration metrics (e.g. RPC).

CHANGELOG.md
```diff
@@ -31,3 +31,8 @@
   ([#136](https://github.com/open-telemetry/semantic-conventions-genai/pull/136))
 
 ### 📚 Clarifications 📚
+
+- Clarify that `gen_ai.client.operation.duration` covers GenAI client operations such as
+  inference, embeddings, retrieval, memory, agent creation, and remote agent invocation,
+  and should not be recorded for internal operations such as tool execution, workflow
+  invocation, planning, or in-process agent invocation.
```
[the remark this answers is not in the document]
trask: Fixed in https://github.com/open-telemetry/semantic-conventions-genai/commit/981640a.

Added the `([#215](...))` PR link to the changelog entry.

model/gen-ai/metrics.yaml
```diff
@@ -33,7 +33,18 @@ metrics:
...
-    brief: 'GenAI operation duration.'
+    brief: 'GenAI client operation duration.'
+    note: |
+      This metric corresponds to the GenAI client spans:
+      [Inference](/docs/gen-ai/gen-ai-spans.md#inference),
+      [Embeddings](/docs/gen-ai/gen-ai-spans.md#embeddings),
+      [Retrieval](/docs/gen-ai/gen-ai-spans.md#retrievals),
+      [Memory](/docs/gen-ai/gen-ai-spans.md#memory),
+      [Create Agent](/docs/gen-ai/gen-ai-agent-spans.md#create-agent-span),
+      and [Invoke Agent Client](/docs/gen-ai/gen-ai-agent-spans.md#invoke-agent-client-span).
```
lmolkova, line 44: we'll need to add metric for internal agent and tool calls to cover the gap. What would we call them? Also would it be weird to have remote invoke_agent and local invoke_agent reported with different names?

Thinking through options:
1. Per kind: `gen_ai.client.operation.duration` for client spans, `gen_ai.internal.operation.duration` for internal ?
2. Per operation group(?): `gen_ai.client.inference | embedding | memory | retrieval  | agent_crud .duration`, `gen_ai.execute_tool | workflow | invoke_agent.duration` (both client and internal because internal / client is minor impl detail and does not really matter?)
3. Some other form of grouping with dedicated metrics for important things: inference, invoke_agent, invoke_workflow, execute_tool, retrieval, memory and a bag with random stuff  (memory store CRUD, agent CRUD, datastore CRUD like `gen_ai.client.control_plane.operation.duration`)

Options 2 and 3 allow us to be stricter when defining metrics, have different histogram boundaries (important for exponential too - some operations are super-short and agents can last for days which could lead to losing precision).

I think we should separate both kinds of agent invocations from shorter operations. 

PS: I learned my stance on 'inference' not covering 'embeddings' is partially wrong (seems app developers assume they are different while academics consider embeddings to be a part of inference), so I'm open to merge them into one term and update our terminology, but if won't help much with quite a few operation types.

Nik-Reddy: Implementer-side angle on the three options:

Concretely, the spans that need a duration metric but are not covered by this clarification are Execute tool, Invoke agent internal, Invoke workflow, and Plan. They have very different shape from the six client spans:

- Execute tool, Plan: typically sub-second, similar shape to Inference, so the existing `[0.01, 0.02, 0.04, ...]` boundaries fit.
- Invoke agent internal, Invoke workflow: can run minutes to hours; the current boundaries lose precision well before the long tail of real agent runs.

Option 1 (client vs internal) is the smallest blast radius today, but it groups all four internal kinds into one metric where the bucket-shape problem cannot be solved cleanly for both ends.

Option 3 reads cleanest from the instrumenter side: name the agent-run and workflow metrics distinctly so they can carry longer ExplicitBucketBoundaries, keep a shorter-tail metric for execute_tool and plan, and let the client set keep its current boundaries. It also lets each metric scope its own required attributes, which incidentally addresses the framing that started #93 ( `gen_ai.provider.name` being required via the shared `metric_attributes.gen_ai` group even where the operation has no GenAI provider).

On local vs remote invoke_agent: separate metric names per kind reads less confusing than one name sliced by `gen_ai.operation.name`, because operators querying p95 of agent invocations almost always want to slice by "did this go over the wire" anyway and would otherwise have to invert that on every query.

If option 3 (or a variant) is the direction, the internal-operation metric will need an analogous note pointing at its corresponding spans, so it may be worth at least agreeing on the shape before #215 merges so the two scopes land paired.

lmolkova: Created this issue based on the offline discussion with @trask and @aabmass  https://github.com/open-telemetry/semantic-conventions-genai/issues/249

Nik-Reddy: Thanks, this clarification looks directionally aligned with the broader modeling discussion in #249.
 
 I think the important part this PR captures is that **`gen_ai.client.operation.duration`** should not be treated as a catch-all metric for internal framework operations such as tool execution, workflow invocation, planning, or in-process agent invocation. That distinction matters for agent telemetry because those operations have different span boundaries, attributes, and duration distributions from client-side inference/retrieval/agent calls.
 
 Longer term, #249 may split this into more precise operation-class duration metrics, but this PR seems useful as an immediate clarification for the current metric.

trask:

Create `changelog.d/215.clarification.md` containing the change log entry, e.g.:

```md
Clarify that `gen_ai.client.operation.duration` corresponds to the GenAI inference, embeddings, retrieval, memory, create agent, and invoke agent client spans, and that when the metric is reported alongside one of these spans its value should be the same as the span duration.
```

Thanks!

PR #218, opened 2026-05-29
Change gen_ai.request.top_k from double to int

trask:
Changes `gen_ai.request.top_k` from `double` to `int`.

`top_k` is a count of candidate tokens and is modeled as an integer by all known providers (Anthropic `top_k`, Cohere `k`, Google `topK`). The original `double` type appears to have been inherited by pattern from the neighboring `top_p` / `temperature` sampling attributes when `top_k` was added in open-telemetry/semantic-conventions#1059.

Surfaced during review of #217.


trask:
Closing in favor of #218

PR #220, opened 2026-05-29
Clarify MCP context propagation

### LINKED ISSUE #85: MCP: update context propagation
it's now in the protocol  https://modelcontextprotocol.io/community/seps/414-request-meta

trask:
Clarify MCP context propagation now that MCP SEP-414 defines the unprefixed `traceparent`, `tracestate`, and `baggage` keys in request `params._meta`. This removes the obsolete provisional warning and points the convention text at the final SEP.

Resolves #85


lmolkova (approved):

docs/gen-ai/mcp.md
```diff
@@ -42,24 +42,23 @@ HTTP request in the corresponding request and response streams.
...
-> context propagation discussions in MCP repository:
-> [modelcontextprotocol#246](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/246)
-> and
-> [modelcontextprotocol#414](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/414).
->
-> If the MCP or JSON-RPC specifications provide official guidance, instrumentations
-> SHOULD prioritize that over the recommendation provided in this section.
+> MCP defines an exception to the `params._meta` DNS-prefixing convention for
+> the W3C Trace Context and Baggage keys. Instrumentations SHOULD use the
+> unprefixed `traceparent`, `tracestate`, and `baggage` keys when propagating
```
alexmojaki: instrumentation doesn't typically think about specific keys like traceparent, it just calls `inject` as in e.g. https://opentelemetry.io/docs/languages/js/propagation/#generic-example

should other keys be prefixed?

trask: https://github.com/open-telemetry/semantic-conventions-genai/pull/220/commits/82370e20086b918bed526f85e3df45e201dae1cd

alexmojaki: i feel like that implies that more needs to be said, like an example of this DNS prefixing from an instrumentor's perspective. if all other keys are prefixed with e.g. `opentelemetry.io.` then i can see how that would work. if the prefix is supposed to vary per key...

trask: I'm not sure that other keys should be prefixed with `io.opentelemetry`, since I think that implies ownership of the definition of those keys by OpenTelemetry, which isn't really true as we're just passing them along.

The only other (non-deprecated) propagator in the OpenTelemetry Specification is B3 (Zipkin). Ideally any prefixing for those keys would be defined by the Zipkin project (https://github.com/openzipkin/b3-propagation).

Since MCP is baking W3C context propagation into their protocol, maybe we should say something like "MCP instrumentation SHOULD use W3C propagator and ignore any propagators configured by the user". I think something like this was done previously for AWS Lambda instrumentation since AWS propagated context to Lambdas using their own special "aws xray" headers.

trask: > Since MCP is baking W3C context propagation into their protocol, maybe we should say something like "MCP instrumentation SHOULD use W3C propagator and ignore any propagators configured by the user".

@alexmojaki @lmolkova let me know what you think: ce8e852592921ea40ff716d5e66c97ba0ef37149

lmolkova: I like the spirit of using W3C propagator to force more apps to use W3C.

At the same time I think the goal is to have client and the server it calls use the same correlation protocol. 
If MCP server uses something custom (e.g. composite thing that propagates w3c along with some custom x-correlation-id), client should be able to configure the same protocol. I don't see any harm in using current propagator configured in the OTel SDK - if different propagators are configured by client and server it would break, but it takes a deliberate effort to configure each of them.

I.e. we enforce w3c (good) but ask instrumentation to do something unusual (not great) and make things less flexible (not great). I can live with this, but don't think it's necessary. I'd prefer user-configured propagator

alexmojaki: I think the MCP people just failed to realise that there are other possible keys. I think not using the configured propagator as-is is weird.

trask: makes sense. I've updated it to be generic:

> Instrumentations SHOULD propagate context using the configured [OpenTelemetry propagators](https://opentelemetry.io/docs/specs/otel/context/api-propagators/) by injecting it into the MCP request [`params._meta`](https://modelcontextprotocol.io/specification/2025-11-25/basic#_meta) property bag when creating a request or notification. The receiver extracts the context from `params._meta` and uses it as the remote parent.
>
> Although the MCP convention expects keys in `params._meta` to be DNS-prefixed, the context propagation keys SHOULD be written unprefixed. [SEP-414](https://modelcontextprotocol.io/community/seps/414-request-meta) defines this explicit handling for the W3C Trace Context and Baggage keys (`traceparent`, `tracestate`, and `baggage`).


https://github.com/open-telemetry/semantic-conventions-genai/pull/220/commits/0393513c908bf895e7b3ec480a2c1eb073c584af

alexmojaki (approved):

Issue #171, opened 2026-05-19
[semconv] propose `gen_ai.agent.finish_reason` attribute for agent loop termination


Nik-Reddy:
## Summary

Add a `gen_ai.agent.finish_reason` attribute on agent invocation spans (`gen_ai.operation.name=invoke_agent`, `create_agent`) so consumers can tell from telemetry *why* an agent loop terminated. This is the agent-loop equivalent of `gen_ai.response.finish_reasons` on LLM completion spans.

## Current behavior

LLM completion spans already record termination cause via `gen_ai.response.finish_reasons[]` with values like `stop`, `length`, `tool_calls`, `content_filter`. Agent invocation spans have no equivalent. From the trace alone, an observer cannot distinguish between:

- the agent decided the task was done
- the agent hit an implementation-defined iteration cap
- the agent exhausted a token or wall-clock budget
- the agent was cancelled by the caller
- the agent hit a safety / policy guardrail
- the agent transferred control to another agent
- the agent stopped because a critical tool call failed
- the agent terminated due to an unrecoverable error in its own loop

All of these collapse into one `Ok`/`Error` status today, which is too coarse for SLO, cost, reliability, and evaluation use cases.

## Proposed attribute

`gen_ai.agent.finish_reason` (string, conditionally required when the agent loop has terminated)

Suggested enum (Development stability):

- `completed`: the agent decided the task is done and produced a final output
- `max_iterations`: implementation-defined iteration limit reached
- `max_tokens`: implementation-defined token budget exhausted
- `time_budget`: implementation-defined wall-clock budget exhausted
- `cancelled`: cancelled by the caller (signal, context cancellation, user UI)
- `error`: terminated due to an unrecoverable error in the agent loop itself
- `guardrail`: terminated by a safety or policy guardrail
- `handoff`: control transferred to another agent (pairs with A2A spans, see #70)
- `tool_failure`: terminated because a critical tool call failed unrecoverably

Open to renaming individual values; this list is meant to start the SIG conversation rather than fix the final enum.

## Use cases

1. SLO dashboards split agent duration and token-cost rollups by  `finish_reason` so completed runs are separated from budget-truncated and error runs.
2. Reliability alerts on `max_iterations` and `time_budget` rates  that today require joining spans with framework-specific logs.
3. Cost attribution: separate completed runs from runs that hit a budget cap, which have different cost semantics.
[...]

PR #238, opened 2026-06-03
Add `gen_ai.agent.finish_reason` attribute for agent loop termination

Nik-Reddy:
## Summary

Introduces `gen_ai.agent.invocation.status` on the `invoke_agent.client` and `invoke_agent.internal` spans, with a single initial enum value: `completed`. It records the terminal success signal for an agent invocation, which is the one gap not already covered by existing conventions.

Refs: #171

## Design history

This PR went through several rounds of review-driven narrowing:

1. **Original proposal**: `gen_ai.agent.finish_reason` with three values (`completed`, `max_iterations`, `guardrail`).
2. **@lmolkova** pointed out these conflate three distinct concepts: error info (`error.type`), invocation status (idle/in-progress/completed/failed/cancelled), and final-inference finish reason (`gen_ai.response.finish_reasons`). Most agent providers also don't use "finish reason" terminology at all.
3. Reworked to `gen_ai.agent.invocation.status` with `completed`/`failed`/`cancelled`, moving failure causes to `error.type` and span error status.
4. **@lmolkova** followed up: `failed`/`cancelled` are still just duplicating `error.type` and span error status, and aren't GenAI/agent-specific since any operation can be cancelled.
5. **Final scope**: dropped `failed` and `cancelled` entirely. `completed` is the only value, because it's the one positive signal existing conventions don't already provide. Failure is fully covered by `error.type` and span error status.

## Why `completed` alone isn't redundant

`error.type` and span error status tell you an invocation failed. Nothing today tells you it succeeded. This attribute fills that one gap without re-encoding information conventions already carry elsewhere.

## Reference scenarios

Three scenarios exercise the attribute on the success path:

| Scenario | Span variant | How status is derived |
| --- | --- | --- |
| `openai-agents` | `invoke_agent.internal` | Set after `Runner.run()` returns without raising. The SDK raises (`MaxTurnsExceeded`, guardrail tripwires, etc.) on failure, so a normal return is the completion signal. |
| `pydantic-ai` | `invoke_agent.internal` | Set after `agent.run_sync()` returns without raising, same reasoning. |
| `azure-ai-foundry` | `invoke_agent.client` | Read from the actual Responses API `response.status` field (`== "completed"`), not assumed from a successful HTTP call. The Responses API can return `incomplete`/`failed`/`cancelled` without raising. |

`data.json` and the coverage reports under `reference/reports/` were regenerated via `uv run run-scenario` and `uv run update-reports`.

## Files changed

- `model/gen-ai/registry.yaml`, `model/gen-ai/spans.yaml`: attribute definition and span references.
- `schema-snapshot/registry.yaml`: regenerated via `weaver registry generate`.
- `docs/registry/attributes/gen-ai.md`, `docs/gen-ai/*.md`: regenerated via `make generate-all` (includes `update-upstream-links`, so cross-repo attribute links resolve to the pinned `v1.44.0` upstream tag instead of dangling relative paths).
- `reference/scenarios/{openai-agents,pydantic-ai,azure-ai-foundry}/`: scenario and `data.json` updates.
- `reference/reports/invoke-agent-{client,internal}-span.md`: regenerated.
- `changelog.d/238.enhancement.md`: Towncrier fragment (migrated from a direct `CHANGELOG.md` edit per repo policy).

## Validation

- `uv run run-scenario openai-agents / pydantic-ai / azure-ai-foundry`: clean, no violations tied to this attribute.
- `uv run update-reports`: reports regenerated.
- `make generate-all`: docs, model, and schema snapshot regenerated and committed in sync.
- All 40 CI checks pass, including `required-status-check`.


docs/gen-ai/gen-ai-agent-spans.md
```diff
@@ -254,7 +255,26 @@ applicable `aws.bedrock.*` attributes and are not expected to include
 the canonical name of exception that occurred, or another low-cardinality error identifier.
 Instrumentations SHOULD document the list of errors they report.
 
-**[4] `gen_ai.conversation.id`:** Instrumentations SHOULD populate conversation id when they have it readily available
+**[4] `gen_ai.agent.finish_reason`:** when the agent loop has terminated and the termination reason is known
```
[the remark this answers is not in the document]
Nik-Reddy: Leaving the wording as-is. This line is generated by weaver from `model/gen-ai/spans.yaml`'s `conditionally_required` text, and the convention in that field is a lowercase fragment, not a full sentence. The same file renders many peers the same way, e.g. **`[8] gen_ai.output.type:`** when applicable and if the request includes an output format. a few lines below. Rewriting only this one would make it the odd one out in the generated doc.

docs/registry/attributes/gen-ai.md
```diff
@@ -240,6 +258,16 @@ If there is no low-cardinality workflow name available for a given framework, th
 
 ---
 
+`gen_ai.agent.finish_reason` has the following list of well-known values. If one of them applies, then the respective value MUST be used; otherwise, a custom value MAY be used.
+
+| Value | Description | Stability |
+| --- | --- | --- |
+| `completed` | The agent loop terminated normally and produced a final output. | ![Development](https://img.shields.io/badge/-development-blue) |
+| `guardrail` | The agent loop terminated because a guardrail or safety policy stopped the invocation. | ![Development](https://img.shields.io/badge/-development-blue) |
+| `max_iterations` | The agent loop terminated because an implementation-defined iteration (turn) limit was reached. | ![Development](https://img.shields.io/badge/-development-blue) |
```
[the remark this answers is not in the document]
Nik-Reddy: Fixed. Reordered the YAML members, note bullets, and examples in `model/gen-ai/registry.yaml` to alphabetical (`completed`, `guardrail`, `max_iterations`) so the summary table and the well-known values table render in the same order. CHANGELOG and regenerated docs updated in the same commit.

model/gen-ai/registry.yaml
```diff
@@ -261,6 +261,41 @@ attributes:
     brief: The version of the GenAI agent.
     examples: ["1.0.0", "2025-05-01"]
     stability: development
+  - key: gen_ai.agent.finish_reason
+    type:
+      members:
+        - id: completed
+          stability: development
+          value: "completed"
+          brief: 'The agent loop terminated normally and produced a final output.'
```
MikeGoldsmith: The "and produced a final output" clause is stricter than what the scenarios verify. In openai-agents/scenario.py, we are setting `completed` after Runner.run returns regardless of final_output truthiness.

I'd suggest dropping the "and produced a final output" clause and leave just "The agent loop terminated normally.".

reference/scenarios/openai-agents/scenario.py
```diff
@@ -158,6 +158,7 @@ async def _capture_create(*args, **kwargs):
                         ]
                     ),
                 )
+            agent_span.set_attribute("gen_ai.agent.finish_reason", "completed")
```
MikeGoldsmith: It would be good to have a scenario that doesn't result in the completed finish reason, eg max_iterations or guardrail being hit.

MikeGoldsmith (changes requested):
Looks good - left two small suggestions.

model/gen-ai/registry.yaml
Nik-Reddy: Dropped it.

reference/scenarios/openai-agents/scenario.py
Nik-Reddy: Added **run_agent_max_iterations** in the same file. It runs the same agent with **max_turns=1**, catches `MaxTurnsExceeded`, and sets `gen_ai.agent.finish_reason`="max_iterations" on the agent span. main now runs both scenarios so the snapshot exercises the completed and max_iterations termination paths.

Nik-Reddy:
@MikeGoldsmith I addressed both items.
 1. Dropped the "and produced a final output" wording from the completed note.
 2. Added a non-completed openai-agents scenario, run_agent_max_iterations, which catches MaxTurnsExceeded and records gen_ai.agent.finish_reason="max_iterations".

model/gen-ai/spans.yaml
```diff
@@ -121,6 +121,9 @@ attribute_groups:
       - ref: gen_ai.agent.version
         requirement_level:
           conditionally_required: When available.
+      - ref: gen_ai.agent.finish_reason
+        requirement_level:
+          conditionally_required: when the agent loop has terminated and the termination reason is known
```
MikeGoldsmith: We use capitalised sentences and end with a period now 👍🏻 

```suggestion
          conditionally_required: When the agent loop has terminated and the termination reason is known.
```

MikeGoldsmith (approved):
Thanks for you effort on this @Nik-Reddy 👍🏻

Nik-Reddy:
@MikeGoldsmith thanks for the review and approval. I applied the wording suggestion and updated the generated schema snapshot.

reference/scenarios/openai-agents/scenario.py
```diff
@@ -158,15 +159,130 @@ async def _capture_create(*args, **kwargs):
...
+                        }
+                        for t in tools
+                        if isinstance(t, FunctionTool)
+                    ]
+                ),
+            )
+            try:
+                await Runner.run(agent, input_text, max_turns=1)
+            except MaxTurnsExceeded:
+                agent_span.set_attribute("gen_ai.agent.finish_reason", "max_iterations")
```
aabmass: 👍 very clear thanks

aabmass (approved):

Nik-Reddy:
@trask PR rebased, conflicts resolved, Could you please take a look for maintainer merge when you get a chance? Thanks!

trask:
Ooops! Sorry about that, reopening.

Nik-Reddy:
@trask  Separately on #[249](https://github.com/open-telemetry/semantic-conventions-genai/issues/249), I added a comment supporting the operation-class split.

 My main question is whether the SIG agrees that the first stable set should stay focused on concrete SDK/reference-scenario boundaries like `invoke_agent`, `execute_tool`, `invoke_workflow`, `plan`, and `manage_agent`, instead of trying to model broader framework concepts like crews, teams, tasks, branches, checkpoints, or sessions right away.

 That would also keep #[238](https://github.com/open-telemetry/semantic-conventions-genai/pull/267)/#[267](https://github.com/open-telemetry/semantic-conventions-genai/pull/267) scoped correctly: they only define finish reasons for the `invoke_agent` boundary, not the full span/metric taxonomy.

AgentGymLeader:
+1 on keeping the first stable set tied to concrete operation boundaries.

For `gen_ai.agent.finish_reason`, the framing that helps implementers is that it records why the agent loop / `invoke_agent` operation stopped, not the whole task or session lifecycle. `guardrail` reads well at that level as an observed termination reason — the loop stopped because a guardrail fired — without the convention having to standardize the policy or its format.

So I'd favor starting from the SDK/reference-scenario boundaries you listed (`invoke_agent`, `execute_tool`, `invoke_workflow`, `plan`, `manage_agent`) and leaving crews/teams/tasks/checkpoints/sessions for a later layer once the operation model is less ambiguous.

model/gen-ai/registry.yaml
```diff
@@ -278,6 +278,40 @@ attributes:
...
+          value: "completed"
+          brief: 'The agent loop terminated normally.'
+        - id: guardrail
+          stability: development
+          value: "guardrail"
+          brief: 'The agent loop terminated because a guardrail or safety policy stopped the invocation.'
+        - id: max_iterations
+          stability: development
+          value: "max_iterations"
+          brief: 'The agent loop terminated because an implementation-defined iteration (turn) limit was reached.'
```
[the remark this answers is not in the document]
Nik-Reddy: Addressed. Added an OpenAI Agents input-guardrail reference scenario that emits **`gen_ai.agent.finish_reason = "guardrail"`** on the `invoke_agent` span, and further rebased.

trask:

Create `changelog.d/238.enhancement.md` containing the change log entry, e.g.:

```md
Add `gen_ai.agent.finish_reason` attribute on agent invocation spans (`invoke_agent.client`, `invoke_agent.internal`) with initial enum values `completed`, `guardrail`, and `max_iterations`. Records why the agent loop terminated; complements per-LLM-call `gen_ai.response.finish_reasons`. ([#171](https://github.com/open-telemetry/semantic-conventions-genai/issues/171))
```

Thanks!

model/gen-ai/registry.yaml
```diff
@@ -285,6 +285,40 @@ attributes:
...
```
trask: do any of these reasons result in marking the span as an error and stamping `error.type`?

reference/scenarios/azure-ai-foundry/scenario.py
```diff
@@ -164,6 +164,7 @@ def run_invoke_agent(client):
                 span.set_attribute("gen_ai.usage.output_tokens", response.usage.output_tokens)
 
             print(f"    -> {response_text or response.id}")
+            span.set_attribute("gen_ai.agent.finish_reason", "completed")
```
lmolkova: is should come from something on the actual response like it would in a case of a real instrumentation

reference/scenarios/openai-agents/scenario.py
```diff
@@ -158,15 +160,190 @@ async def _capture_create(*args, **kwargs):
...
```
lmolkova: this looks more like `error.type` than finish reason - all otel conventions spans should set this attribute when operation fails.

reference/scenarios/openai-agents/scenario.py
```diff
@@ -158,15 +160,190 @@ async def _capture_create(*args, **kwargs):
...
+        agent_span.set_attribute(
+            "gen_ai.system_instructions", json.dumps([{"parts": [{"type": "text", "content": agent.instructions}]}])
+        )
+        agent_span.set_attribute(
+            "gen_ai.input.messages", json.dumps([{"role": "user", "parts": [{"type": "text", "content": input_text}]}])
+        )
+        try:
+            await Runner.run(agent, input_text)
+        except InputGuardrailTripwireTriggered:
+            agent_span.set_attribute("gen_ai.agent.finish_reason", "guardrail")
```
lmolkova: this also looks like an error.type on invoke_agent span

model/gen-ai/spans.yaml
```diff
@@ -121,6 +121,9 @@ attribute_groups:
         brief: The version of the invoked GenAI agent.
         requirement_level:
           conditionally_required: When available.
+      - ref: gen_ai.agent.finish_reason
```
lmolkova: I don't think agent frameworks/APIs use this terminology in general. 

| agent provider | status / finish reason terminology / links | possible values  | 
| --- | --- | --- |
| Claud managed agents | [Session status](https://platform.claude.com/docs/en/managed-agents/session-operations#session-statuses) | idle, running, rescheduling, terminated. session.status_idle can include stop_reason, e.g. requires_action; self-hosted worker docs mention end_turn stop reason. |
| AWS Bedrock | Does not have status / finish reason / etc. When operation fails, [returns exceptions](https://docs.aws.amazon.com/bedrock/latest/APIReference/API_agent-runtime_InvokeAgent.html) | n/a |
| A2A | [TaskStatus](https://agent2agent.info/docs/concepts/task/#task-related-interfaces) has state and message with details |  submitted, working, input-required, completed, canceled, failed, unknown |
| OpenAI agents | no finish reason / status, throws exceptions when not successfull | Exceptions include MaxTurnsExceeded, ModelBehaviorError, ModelRefusalError, guardrail tripwire errors. |
| ADK | [Event type and fields like usage_metadata](https://adk.dev/streaming/dev-guide/part3/#the-event-class) | turn_complete, interrupted, finish_reason, error_code, error_message, Errors surface through [error_code/error_message](https://adk.dev/streaming/dev-guide/part3/#error-events). |
| OpenAI assistants / responses (used by Azure AI Foundry) | [status](https://developers.openai.com/api/reference/resources/responses/methods/create#(resource)%20responses%20%3E%20(model)%20response%20%3E%20(schema)%20%3E%20(property)%20status), [incomplete_reason](https://developers.openai.com/api/reference/resources/responses/methods/create#(resource)%20responses%20%3E%20(model)%20response%20%3E%20(schema)%20%3E%20(property)%20incomplete_details) | status: completed, failed, in_progress, cancelled, queued, or incomplete; reason: "max_output_tokens" or "content_filter" |

lmolkova (changes requested):
Based on what different agent providers and frameworks support, we have 3 distinct concepts here:

1. error info. When operation fails (raises exception), we need to report `error.type` attribute. This would most likely cover max_iterations or other similar things (Langgraph recursion limit)

2.  invocation status: idle | in-progress | completed | failed | cancelled. 

3. why model stopped generating content (but operation was successful) which is literally the finish reason of the final inference call done my the agent. Not always available. The same enum as we use today. I think we should keep the same attribute (or introduce a singular version of it for the last reason. E.g. `gen_ai.final_response.finish_reason` or just `gen_ai.response.finish_reason`)

Nik-Reddy:
Thanks @lmolkova and @trask, reworked this based on the feedback.
 
 Main change: Removed `gen_ai.agent.finish_reason` and replaced it with `gen_ai.agent.invocation.status`, scoped only to terminal invocation outcomes: `completed`, `failed`, and `cancelled`.
 
 I also split the concepts as suggested:
 - invocation terminal outcome: `gen_ai.agent.invocation.status`
 - failure cause: `error.type`
 - final model stop reason: existing `gen_ai.response.finish_reasons`
 
 For the OpenAI Agents reference scenarios, `MaxTurnsExceeded` and input guardrail tripwire now mark the invoke_agent span as error and set `error.type`, instead of using them as agent finish reasons.
 
Alongside, also added registry text clarifying that idle/in-progress lifecycle states are out of scope for this span attribute, and that failure causes such as max iterations, guardrail tripwire, tool failure, or framework exceptions should not be encoded in the status value.

model/gen-ai/registry.yaml
Nik-Reddy: Addressed in the latest push. Removed `gen_ai.agent.finish_reason` and replaced it with terminal `gen_ai.agent.invocation.status`. Failure causes are now explicitly out of scope for the status value and should be recorded on `error.type`; failed invocations should also mark the span as error

reference/scenarios/azure-ai-foundry/scenario.py
Nik-Reddy: @lmolkova addressed it. The Azure AI Foundry reference scenario now sets `gen_ai.agent.invocation.status = "completed"` only when the actual response status is completed, instead of hardcoding a synthetic finish reason.

model/gen-ai/spans.yaml
Nik-Reddy: Addressed by removing **`gen_ai.agent.finish_reason`**. The replacement attribute is **`gen_ai.agent.invocation.status`**, scoped only to terminal invocation outcomes. The registry text also clarifies that idle/in-progress lifecycle states are out of scope for this span attribute, and final model stop reasons remain on **gen_ai.response.finish_reasons**.

reference/scenarios/openai-agents/scenario.py
Nik-Reddy: Addressed it. The input guardrail tripwire scenario now marks the invoke_agent span as error and sets `error.type` to the framework exception, with **`gen_ai.agent.invocation.status = "failed"`**.

reference/scenarios/openai-agents/scenario.py
Nik-Reddy: Addressed it. The **MaxTurnsExceeded** scenario now marks the invoke_agent span as error and sets error.type to the framework exception, with **gen_ai.agent.invocation.status = "failed"**.

docs/gen-ai/gen-ai-agent-spans.md
```diff
@@ -194,42 +194,43 @@ Examples: OpenAI Assistants API, AWS Bedrock Agents.
 
 **Span kind** SHOULD be `CLIENT`.
 
-**Span status** SHOULD follow the [Recording Errors](https://github.com/open-telemetry/semantic-conventions/blob/v1.42.0/docs/general/recording-errors.md) document.
+**Span status** SHOULD follow the [Recording Errors](https://github.com/open-telemetry/semantic-conventions/blob/v1.41.1/docs/general/recording-errors.md) document.
```
lmolkova: please rebase and regenerate docs

reference/scenarios/azure-ai-foundry/scenario.py
```diff
@@ -164,7 +164,11 @@ def run_invoke_agent(client):
                 span.set_attribute("gen_ai.usage.output_tokens", response.usage.output_tokens)
 
             print(f"    -> {response_text or response.id}")
+            if getattr(response, "status", None) == "completed":
+                span.set_attribute("gen_ai.agent.invocation.status", "completed")
         except Exception as e:
+            span.set_attribute("gen_ai.agent.invocation.status", "failed")
```
lmolkova: can we do better than 'failed'? should it be `response.status` instead?

reference/scenarios/openai-agents/scenario.py
```diff
@@ -158,15 +160,185 @@ async def _capture_create(*args, **kwargs):
...
+                        for t in tools
+                        if isinstance(t, FunctionTool)
+                    ]
+                ),
+            )
+            try:
+                await Runner.run(agent, input_text, max_turns=1)
+            except MaxTurnsExceeded as exc:
+                agent_span.set_attribute("gen_ai.agent.invocation.status", "failed")
+                agent_span.set_attribute("error.type", exc.__class__.__name__)
```
lmolkova: it seems `error.type` provides better precision here than `gen_ai.agent.invocation.status` - why do we need status?

reference/scenarios/openai-agents/scenario.py
```diff
@@ -158,15 +160,185 @@ async def _capture_create(*args, **kwargs):
...
+        agent_span.set_attribute(
+            "gen_ai.system_instructions", json.dumps([{"parts": [{"type": "text", "content": agent.instructions}]}])
+        )
+        agent_span.set_attribute(
+            "gen_ai.input.messages", json.dumps([{"role": "user", "parts": [{"type": "text", "content": input_text}]}])
+        )
+        try:
+            await Runner.run(agent, input_text)
+        except InputGuardrailTripwireTriggered as exc:
+            agent_span.set_attribute("gen_ai.agent.invocation.status", "failed")
```
lmolkova: same here - the status seems duplicative of error.type and span status

model/gen-ai/registry.yaml
```diff
@@ -285,6 +285,53 @@ attributes:
...
+          stability: development
+          value: "completed"
+          brief: 'The agent invocation completed successfully.'
+        - id: failed
+          stability: development
+          value: "failed"
+          brief: 'The agent invocation failed.'
+        - id: cancelled
+          stability: development
+          value: "cancelled"
```
lmolkova: it seems that's the only status not represented well today with `error.type` but also it's not specific to genai or agents; inference or tool calls can be cancelled. We should not invent an attribute for each of them.

lmolkova (changes requested):
It seems most of the current scenarios being solved are satisfied with `error.type` attribute. Would it make sense to start with it and only add additional status(es) when we know they are needed?

Nik-Reddy:
Reworked based on feedback from @lmolkova and @trask.
 - Removed gen_ai.agent.finish_reason entirely
 - Replaced with gen_ai.agent.invocation.status containing only completed
 - Dropped failed and cancelled values — failures are fully covered by error.type + span error status
 - Rebased on main, regenerated docs and schema snapshot via weaver
  
The only gap in existing OTel conventions is a positive success signal for agent loop termination. Error cases (max_iterations, guardrail, tool_failure) are properly represented by error.type and don't need a separate attribute. This keeps the surface area minimal and non-duplicative.
  
Still TODO: Regenerate data.json for reference scenarios (will do in a follow-up push).

reference/scenarios/azure-ai-foundry/scenario.py
Nik-Reddy: Removed failed entirely. The attribute now only carries completed. Failure information goes through `error.type` and span error status.

reference/scenarios/openai-agents/scenario.py
Nik-Reddy: It makes sense that error.type covers failures better. Removed all failure values. 

The remaining completed value fills the one gap error.type can't: signaling that the agent loop terminated successfully.

reference/scenarios/openai-agents/scenario.py
Nik-Reddy: removed failed and cancelled. Only completed remains, which is not duplicative since there's no existing attribute that signals successful agent loop termination.

model/gen-ai/registry.yaml
Nik-Reddy: Makes sense. Removed cancelled from this attribute. If cancellation needs a cross-cutting convention in the future, it should be handled at the OTel-wide level, not per-operation. This attribute now only carries completed.

Nik-Reddy:
Follow-up push finished.
- Regenerated `data.json` for the `openai-agents`, `pydantic-ai`, and `azure-ai-foundry` reference scenarios to reflect `gen_ai.agent.invocation.status`.
- Regenerated docs via a full `make generate-all` (the prior push had skipped `update-upstream-links`, leaving broken relative links to upstream-owned attribute pages like `server.md`/`error.md`); this fixed the `generated-docs` and `links` checks.
- Fixed the `azure-ai-foundry` scenario to derive `gen_ai.agent.invocation.status` from the actual `response.status` returned by the Responses API instead of hardcoding `"completed"`.

Nik-Reddy:
updated body with latest state of the changes incorporated, can you review and merge @trask @lmolkova

Nik-Reddy:
@trask the changelog fragment is already in place at changelog.d/238.enhancement.md, and there is no direct edit to CHANGELOG.md left in this branch. This should be resolved.

lmolkova (changes requested):
please take a look at the `gen_ai.response.status` - https://github.com/open-telemetry/semantic-conventions-genai/blob/main/docs/registry/attributes/gen-ai.md#gen-ai-response-status

it can be expanded to cover agents and referenced on agent spans.

In the current shape, gen_ai.agent.invocation.status attribute does not convey any information at all. A failed agent invocation would have `error.type` and span status set to error, any queries can assume that invocation has completed successfully when error.type is not present.

Nik-Reddy:
@lmolkova fair point on redundancy. Before reworking, want to confirm the shape:

- Same `gen_ai.response.status` key on `invoke_agent` spans, or a different attribute? `gen_ai.response.*` today means the model's response, not sure that reads right on a span wrapping multiple model/tool calls.
- `in_progress` is in the existing enum for polled/background generations - applicable to agents too (set mid-span), or agents only ever hit `completed`?
- If reused as-is, agent spans would still only ever show `completed` (failures stay on `error.type`) - same gap you just called out, just on a shared attribute. Is that fine, or do you want `failed`/`cancelled` added to `response.status` for agents?

Can rework once I know which of these you're picturing.

PR #242, opened 2026-06-04
Limit gen_ai.agent.id to stable / static identifiers

lmolkova:
Clarify `gen_ai.agent.id` usage: 

- remove it from internal spans
- update brief and note to provide more context and require stable id
- update brief of refs to explain which agent should be recorded.

This PR also lays ground for the next problem: recording current agent along with to-be-invoked by it agent on the same telemetry item - this is not in scope of this PR, see https://github.com/open-telemetry/semantic-conventions-genai/issues/243

Context:

1. Among hosted agents (AWS Bedrock agent, GCP Agent Platform, Azure AI Foundry, Claude managed agents, OpenAI Assistants) agent id identifies agent for external callers. It's created when agent is created and stays stable until agent is deleted. It effectively identifies a set of settings and configurations and is unrelated to infra / internal deployment details.

2. Among frameworks only one actually has a notion of agent id - crewAI which by default generates a guid when `Agent` class is instantiated. This is high-cardinality id that seems to be something completely different from what hosted agents consider to be an identifier. 

Giving limited usage, usefulness, and high cardinality of the latter, I propose to drop it - stable agent id is something that belongs on metrics (either through resources on hosting side or through explicit attribute on client side) 

## Checklist

See [CONTRIBUTING.md]


reference/scenarios/autogen/scenario.py
```diff
@@ -135,10 +135,6 @@ def _capture_tool_call_id(*_args, tool_call_id=None, **_kwargs):
                 tools=[get_weather],
                 max_tool_iterations=2,
             )
-            agent_id = getattr(agent, "id", None)
```
lmolkova, line 138: I could not find `id` property on the autogen or ag2 agent classes

reference/scenarios/crewai/data.json
```diff
@@ -1,23 +1,12 @@
 {
   "spans": {
-    "create_agent": [
```
lmolkova, line 3: crewai agents are internal, this is just a constructor call, we should not instrument it

reference/scenarios/crewai/scenario.py
```diff
@@ -210,6 +193,148 @@ def _capture_completion(*args, **kwargs):
             print(f"    -> {str(result)[:60]}")
 
 
+def run_agent():
```
lmolkova, line 196: we didn't have an example of invoke_agent for crewai

model/gen-ai/registry.yaml
```diff
@@ -243,8 +243,15 @@ attributes:
     stability: development
   - key: gen_ai.agent.id
     type: string
-    brief: The unique identifier of the GenAI agent.
-    examples: ['asst_5j66UpCpwteGg4YSxUnt7lPY']
+    brief: The unique and stable identifier of the GenAI hosted agent resource.
+    note: >
+      For hosted agents, this SHOULD be the provider-assigned stable identifier of the
```
aabmass, line 265: I think it's pretty clear since you said "resource", but it would be to explicitly call out it's not an instance level identifier.

model/gen-ai/spans.yaml
```diff
@@ -378,16 +378,20 @@ spans:
       - ref: gen_ai.request.model
         sampling_relevant: true
       - ref: gen_ai.agent.id
+        brief: The unique identifier of the GenAI agent created during this operation.
```
aabmass:
```suggestion
        brief: The stable unique identifier of the GenAI agent created during this operation.
```

model/gen-ai/spans.yaml
```diff
@@ -428,6 +432,10 @@ spans:
         sampling_relevant: true
       - ref: server.port
         sampling_relevant: true
+      - ref: gen_ai.agent.id
+        brief: The unique identifier of the invoked GenAI agent.
```
aabmass:
```suggestion
        brief: The stable unique identifier of the invoked GenAI agent.
```

aabmass (approved):

model/gen-ai/registry.yaml
lmolkova: There is 

>       It's NOT RECOMMENDED to record in-memory agent instance ids on this attribute due to their transient nature.

a couple of lines below. Do you think we need more?

docs/gen-ai/gen-ai-agent-spans.md
```diff
@@ -56,13 +56,13 @@ Semantic conventions for individual GenAI systems and frameworks MAY specify dif
...
-| [`server.address`](https://github.com/open-telemetry/semantic-conventions/blob/v1.41.1/docs/registry/attributes/server.md) | ![Stable](https://img.shields.io/badge/-stable-lightgreen) | `Recommended` | string | GenAI server address. [6] | `example.com`; `10.1.2.80`; `/tmp/my.sock` |
+| [`gen_ai.agent.description`](/docs/registry/attributes/gen-ai.md) | ![Development](https://img.shields.io/badge/-development-blue) | `Conditionally Required` If provided by the application. | string | The free-form description of the GenAI agent created during this operation. | `Helps with math problems`; `Generates fiction stories` |
+| [`gen_ai.agent.id`](/docs/registry/attributes/gen-ai.md) | ![Development](https://img.shields.io/badge/-development-blue) | `Conditionally Required` If applicable. | string | The stable unique identifier of the GenAI agent created during this operation. [4] | `asst_5j66UpCpwteGg4YSxUnt7lPY`; `arn:aws:bedrock:us-east-1:123:agent/42`; `urn:agent:projects-123:projects:123:locations:us-east1:aiplatform:reasoningEngines:456` |
```
MikeGoldsmith, line 60: nit: I don't think the GCP identifer looks right, it duplicates projects-123 and projects:123

MikeGoldsmith (approved):

docs/gen-ai/gen-ai-agent-spans.md
lmolkova: I thought this too - https://docs.cloud.google.com/agent-registry/concepts#agent-identifier

> urn:agent:projects-PROJECT_NUMBER:projects:PROJECT_NUMBER:locations:REGION:reasoningEngines:AGENT_ID

PR #226, opened 2026-05-31
Replace Jupiter notebook with models with python file and add CI check that json schemas are up-to-date

lmolkova:
Fixes #64 

- moves python code with models from jupiter notebook to python file (better for review and can be run easily)
- add make target to generate schemas
- add CI job to validate that schemas and python code are in sync

The following significant inconsistencies found:
- `gen-ai-input-messages.json`, `gen-ai-output-messages.json`
    - `$defs` reordered alphabetically (was declaration order); no content change.
    - `ServerToolCallPart.server_tool_call` and `ServerToolCallResponsePart.server_tool_call_response` lost the `oneOf: [{$ref}]` wrapper (and the auto-generated `"title": "Server Tool Call"` / `"Server Tool Call Response"`); now inline `{$ref, description}`. The fields are typed as `Union[GenericServerToolCall]` / `Union[GenericServerToolCallResponse]`, but pydantic collapses a single-member union to a plain reference. As soon as a second concrete type is added, `oneOf` reappears naturally.

- `gen-ai-system-instructions.json`. Same + added missing parts. This schema was not updated. I don't think it's right to have all the possible parts in system instructions - will follow up in a different PR.

PR #257, opened 2026-06-05
Limit supported  part types  on `gen_ai.system_instructions` to text only

lmolkova:
Follow up to https://github.com/open-telemetry/semantic-conventions-genai/pull/226

None of the model providers support non-text parts and not all the input/output parts make sense on the system instructions (e.g. tool calls should never appear there), so limiting supported part to text and we can always extend to support other part types in the future.

## Checklist

See [CONTRIBUTING.md]


MikeGoldsmith (approved):
I think we'll need to update the scenarios too as they still used the old `gen_ai.system_instructions`. Also, I agree with copilot that we can probably remove GenericPart in `SystemInstructionPart`, `TextPart` is enough.

trask:

Create `changelog.d/257.deprecation.md` containing the change log entry, e.g.:

```md
Limit supported `gen_ai.system_instructions` part types to text.
```

Thanks!

lmolkova:
> Also, I agree with copilot that we can probably remove GenericPart in `SystemInstructionPart`, `TextPart` is enough.

@MikeGoldsmith GenericPart is there to tell consumers that they should be ready to get things other than Text and give individual providers place to add custom parts. The same pattern is used in input / output messages

trask (approved):

PR #258, opened 2026-06-06
Add `gen_ai.request.reasoning.level` attribute

### LINKED ISSUE #189: Consider standardizing `gen_ai.request.reasoning_effort`
Hi, and thanks for the continued work on the GenAI semantic conventions.

I may be missing some prior discussion here, but I noticed a small gap while looking at the current GenAI attribute registry and recent provider documentation.

Today, the GenAI registry defines several request-side attributes for traditional generation controls, such as:

- `gen_ai.request.temperature`
- `gen_ai.request.top_p`
- `gen_ai.request.top_k`
- `gen_ai.request.max_tokens`

It also already has usage-side reasoning information, such as:

- `gen_ai.usage.reasoning.output_tokens`

However, I could not find a request-side attribute for the reasoning/thinking effort requested by the client.

OpenTelemetry GenAI attribute registry:

- https://opentelemetry.io/docs/specs/semconv/registry/attributes/gen-ai/

## Background

I noticed this while reviewing the Gemini 3.5 Flash developer guidance after Google I/O last week. The part that stood out to me was the Gemini 3.x parameter update guidance: Google says that `temperature`, `top_p`, and `top_k` are no longer recommended because Gemini 3's reasoning capabilities are optimized for the default settings, and recommends not changing those values. The same guide also says to use the `thinking_level` string enum instead of the raw numeric `thinking_budget` parameter.

This seemed related to a broader pattern across providers: newer reasoning-capable models increasingly expose a request-side control for how much reasoning or thinking the model should perform.

Relevant provider examples:

- **OpenAI — `reasoning.effort`**
  - OpenAI documents `reasoning.effort` as the parameter that guides the model on how much to think when performing a task.
  - Supported values are model-dependent and can include `none`, `minimal`, `low`, `medium`, `high`, and `xhigh`.
  - https://developers.openai.com/api/docs/guides/reasoning

- **Google Gemini — `thinking_level`**
  - Google's Gemini 3.5 Flash guidance documents `thinking_level` with values such as `minimal`, `low`, `medium`, and `high`.
  - The same guidance says to use `thinking_level` instead of `thinking_budget`.
  - https://ai.google.dev/gemini-api/docs/interactions/whats-new-gemini-3.5

- **Anthropic Claude — `effort`**
  - Anthropic documents `effort` as a parameter for controlling how many tokens Claude uses when responding, trading off response thoroughness and token efficiency.
  - For Claude Opus 4.6 and Claude Sonnet 4.6, the docs say that `effort` replaces `budget_tokens` as the recommended way to control thinking depth.
  - https://platform.claude.com/docs/en/build-with-claude/effort

### A note on scope differences across providers

While the three provider parameters are conceptually related, they differ slightly in documented scope:

- OpenAI's `reasoning.effort` and Google's `thinking_level` are documented as controls over the model's reasoning/thinking depth.
- Anthropic's `effort`, in contrast, is documented as affecting all tokens in the response — text, tool calls, and extended thinking — and is applicable whether or not extended thinking is enabled.

In other words, Anthropic's `effort` is documented with a somewhat broader scope than the OpenAI and Google equivalents.

## Problem

The current GenAI semantic conventions can record traditional sampling controls and the resulting reasoning token usage, but there does not seem to be a standard request-side attribute for the requested reasoning effort itself.

That leaves a small observability gap.

For example, when investigating latency, token usage, cost, or quality differences across GenAI requests, it would be useful to know whether the request asked the model to use low, medium, high, or maximum reasoning effort.

Today, instrumentations may need to either omit this information or use unofficial/provider-specific attributes.

## Proposal

Consider adding a common request-side attribute:

```text
gen_ai.request.reasoning_effort
```

Possible description:

```text
The requested qualitative reasoning or thinking effort for a GenAI model, when the provider exposes such a control.
```

Possible example values:

```text
none
minimal
low
medium
high
xhigh
max
```

The exact supported values would remain provider/model-dependent.

## Possible provider mapping

| Provider | Provider parameter | Possible SemConv attribute |
|---|---|---|
| OpenAI | `reasoning.effort` | `gen_ai.request.reasoning_effort` |
| Google Gemini | `thinking_level` | `gen_ai.request.reasoning_effort` |
| Anthropic Claude | `effort` | `gen_ai.request.reasoning_effort` |

## Why a common attribute may be useful

The provider parameter names differ, but the user-facing concept seems similar enough to be useful as a common attribute: a qualitative control over how much reasoning/thinking effort the model should spend.

A common attribute would make it easier to analyze:

- latency by requested reasoning effort
- token usage and cost by requested reasoning effort
- quality or evaluation score by requested reasoning effort
- behavior changes when moving from older sampling-heavy configurations to newer reasoning-model configurations

This also complements the existing usage-side attribute:

```text
gen_ai.usage.reasoning.output_tokens
```

In other words:

- `gen_ai.request.reasoning_effort` would describe what the client requested.
- `gen_ai.usage.reasoning.output_tokens` describes what was actually used.

## Non-goals

This proposal is not suggesting that reasoning effort replaces `temperature`, `top_p`, or `top_k`.

It is also not trying to normalize every provider-specific reasoning mechanism into one exact semantic model. The intent is narrower: to provide a standard place to record qualitative request-side reasoning/thinking effort when such a provider control exists.

This proposal also does not require numeric thinking budgets to be mapped into this attribute. If needed, a separate attribute such as `gen_ai.request.reasoning_budget_tokens` might be more appropriate for numeric budget-style controls.

## Open questions

A few things may need discussion:

1. Is `gen_ai.request.reasoning_effort` the right name, or should this be provider-specific?
2. Should the values be documented as a stable enum, or only as known/example values because providers may diverge?
3. Should numeric thinking budgets be handled separately?
4. Should this attribute be `Recommended`, or start as `Opt-In`?

If this direction seems reasonable, I would be happy to try drafting a small spec PR.

katsuhisa91:
## Description

Add `gen_ai.request.reasoning.level` attribute to record the qualitative reasoning/thinking effort level requested for a GenAI model.

Resolves #189

## Motivation

**User journey — latency/cost/quality investigation.**
When investigating latency, token usage, cost, or quality differences across GenAI requests, operators need to know whether the request asked the model to use low, medium, high, or maximum reasoning effort. Today instrumentations must either omit this information or use provider-specific attributes.

This attribute complements the existing usage-side `gen_ai.usage.reasoning.output_tokens` (what was actually used) by recording what the client requested.

**Prior art:**

| Provider | Parameter | Example values |
|----------|-----------|----------------|
| OpenAI | `reasoning.effort` | `low`, `medium`, `high` |
| Google Gemini | `thinking_level` | `minimal`, `low`, `medium`, `high` |
| Anthropic | `output_config.effort` | `low`, `medium`, `high` |

All three major providers now expose a qualitative reasoning effort control. The parameter names differ, but the concept is consistent enough to standardize.

**Design decisions** (per [discussion on #189](https://github.com/open-telemetry/semantic-conventions-genai/issues/189#issuecomment-4550084910) by @trask and [review feedback](https://github.com/open-telemetry/semantic-conventions-genai/pull/258#pullrequestreview-3035289083) by @lmolkova):

- **General, not provider-specific** — a single common attribute.
- **Free-form string** — records whatever value the provider uses, rather than forcing into a fixed enum.
- **Dot-separated namespace** — renamed to `gen_ai.request.reasoning.level` (suggested by @lmolkova) for future extensibility (e.g., `gen_ai.request.reasoning.budget`).
- **Provider-specific parameter mapping** — documented in provider span refinements rather than in the attribute note itself (per @lmolkova's review). Refinement notes now include links to each provider's official documentation.
- **Numeric budgets out of scope** — kept separate per #189 open questions.
- **Recommended** — when applicable.

## Prototype

Added `gen_ai.request.reasoning.level` to OpenAI and Anthropic reference scenarios. The attribute value is sourced from a `request_reasoning_level` variable that is also passed to each provider's SDK call, matching the pattern used by other request attributes in the scenarios (per [review feedback](https://github.com/open-telemetry/semantic-conventions-genai/pull/258#pullrequestreview-3060922549) by @JWinermaSplunk). Provider-specific refinement notes document the parameter mapping with links to official documentation ([OpenAI](https://platform.openai.com/docs/guides/reasoning), [Anthropic](https://docs.anthropic.com/en/docs/build-with-claude/effort)).

## Checklist


schema-snapshot/registry.yaml
```diff
@@ -401,6 +401,21 @@ refinements:
...
+      - high
+      key: gen_ai.request.reasoning_effort
+      note: |
+        The value SHOULD be the exact string value sent to the provider (e.g., `low`, `medium`, `high`). When a provider uses a different parameter name for the same concept (such as OpenAI `reasoning.effort`, Google Gemini `thinking_level`, or Anthropic `effort`), the value SHOULD still be recorded under this common attribute.
+      provenance:
+        path: ./model/gen-ai/registry.yaml
+      requirement_level:
+        recommended: When applicable.
+      stability: development
+      type: string
```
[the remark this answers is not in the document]
katsuhisa91: Not applied — `schema-snapshot/registry.yaml` is fully generated by `make generate-all` (Weaver). The attribute appears in both the `registry:` section (definition) and the `refinements:` sections (per-span references), which is the expected structure. No manual edits are made to this file.

model/gen-ai/registry.yaml
```diff
@@ -168,6 +168,19 @@ attributes:
...
+      The reasoning or thinking effort level requested for a GenAI model,
+      reflecting the provider-specific value sent in the request.
+    note: >
+      The value SHOULD be the exact string value sent to the provider
+      (e.g., `low`, `medium`, `high`). When a provider uses a different
+      parameter name for the same concept (such as OpenAI `reasoning.effort`,
+      Google Gemini `thinking_level`, or Anthropic `effort`), the value
+      SHOULD still be recorded under this common attribute.
+    examples: ['low', 'medium', 'high']
+    stability: development
```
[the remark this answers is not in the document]
katsuhisa91: Not applied — flow-style is the existing convention in this file (e.g., `examples: [100]`, `examples: [0.0]`, `examples: [40]` for other request attributes). Block-style is only used for nested arrays like `stop_sequences`.

model/gen-ai/registry.yaml
```diff
@@ -168,6 +168,19 @@ attributes:
     type: boolean
     brief: Indicates whether the GenAI request was made in streaming mode.
     stability: development
+  - key: gen_ai.request.reasoning_effort
+    type: string
+    brief: >
+      The reasoning or thinking effort level requested for a GenAI model,
+      reflecting the provider-specific value sent in the request.
```
lmolkova: nit (just to keep brief minimalistic):

```suggestion
      The reasoning or thinking effort level requested for a GenAI model.
```

model/gen-ai/registry.yaml
```diff
@@ -168,6 +168,19 @@ attributes:
...
+    note: >
+      The value SHOULD be the exact string value sent to the provider
+      (e.g., `low`, `medium`, `high`). When a provider uses a different
+      parameter name for the same concept (such as OpenAI `reasoning.effort`,
+      Google Gemini `thinking_level`, or Anthropic `effort`), the value
+      SHOULD still be recorded under this common attribute.
```
```suggestion
      The value SHOULD be the exact string value sent to the provider.
      
      Semantic conventions for individual providers SHOULD document
      which input parameter maps to this attribute.
```

Please update corresponding spans for OpenAI and anthopic (you can override note on them) and document the property. We don't have google-specific doc yet but the property should be documented there one there is a dedicated doc.

model/gen-ai/registry.yaml
```diff
@@ -168,6 +168,19 @@ attributes:
     type: boolean
     brief: Indicates whether the GenAI request was made in streaming mode.
     stability: development
+  - key: gen_ai.request.reasoning_effort
```
lmolkova: it seems there are other reasoning properties 

OpenAI has summary in the same [`Reasoning` type](https://developers.openai.com/api/reference/resources/responses/methods/create#(resource)%20responses%20%3E%20(method)%20create%20%3E%20(params)%200.non_streaming%20%3E%20(param)%20reasoning%20%3E%20(schema)), Google has [budget and whether to include thoughts](https://docs.cloud.google.com/python/docs/reference/aiplatform/latest/google.cloud.aiplatform_v1.types.GenerationConfig.ThinkingConfig)

I'm not proposing to add them yet, but we should consider how we'll add them in the future.

It sounds like `gen_ai.request.reasoning.level` `gen_ai.request.reasoning.budget`, etc would be a better choice than `_` (since these are multiple properties describing reasoning config)

reference/reports/inference-span.md
```diff
@@ -29,6 +29,7 @@
 | gen_ai.request.frequency_penalty | [autogen], [crewai], [google-adk], [openai], [pydantic-ai] |
 | gen_ai.request.max_tokens | [anthropic], [autogen], [crewai], [google-adk], [openai], [pydantic-ai] |
 | gen_ai.request.presence_penalty | [autogen], [crewai], [google-adk], [openai], [pydantic-ai] |
+| gen_ai.request.reasoning_effort | (none) |
```
lmolkova: we should add attribute to applicable instrumentations. Some of them should already have reasoning scenarios

model/gen-ai/registry.yaml
katsuhisa91: Thank you for the suggestion — applied. Keeping the brief minimal makes sense.

model/gen-ai/registry.yaml
katsuhisa91: Thank you, this is a much cleaner approach. I'll update the note as suggested and add the attribute to the OpenAI and Anthropic span definitions with provider-specific notes documenting the parameter mapping.

model/gen-ai/registry.yaml
katsuhisa91: Great point — I agree that dot-separated namespacing (`gen_ai.request.reasoning.level`, `gen_ai.request.reasoning.budget`, etc.) is better suited for future extensibility. I'll rename the attribute to `gen_ai.request.reasoning.level` in the next push.

reference/reports/inference-span.md
katsuhisa91: Good call. I'll check which instrumentations already have reasoning scenarios and add the attribute there.

katsuhisa91:
Addressed all feedback.
 
- Renamed to `gen_ai.request.reasoning.level` for future extensibility (`.budget`, etc.)
- Simplified `brief` and `note` per suggestions
- Added provider-specific notes to OpenAI (`reasoning.effort`) and Anthropic (`thinking.effort`) span refinements
- Added attribute to OpenAI and Anthropic reference scenarios
- Regenerated docs, schema snapshot, and reference reports
 
Thanks for the thorough review!

reference/scenarios/anthropic/scenario.py
```diff
@@ -42,6 +42,7 @@ def run_chat():
         span_attributes["server.port"] = port
     with _reference_tracer.start_as_current_span("chat claude-sonnet-4-20250514", attributes=span_attributes) as span:
         span.set_attribute("gen_ai.request.max_tokens", request_max_tokens)
+        span.set_attribute("gen_ai.request.reasoning.level", "medium")
```
JWinermaSplunk: Where does `medium` come from in the actual SDK request? Any chance we can source this attribute value from there?

reference/scenarios/openai/scenario.py
```diff
@@ -59,6 +59,7 @@ def run_chat_reference(client):
         span.set_attribute("gen_ai.request.frequency_penalty", request_frequency_penalty)
         span.set_attribute("gen_ai.request.presence_penalty", request_presence_penalty)
         span.set_attribute("gen_ai.request.top_p", request_top_p)
+        span.set_attribute("gen_ai.request.reasoning.level", "medium")
```
JWinermaSplunk: Same thought as above.

reference/scenarios/anthropic/scenario.py
katsuhisa91: Good catch — updated to source the value from a `request_reasoning_level` variable that is also passed to the SDK call via `output_config={"effort": ...}` ([docs](https://docs.anthropic.com/en/docs/build-with-claude/effort)). Same pattern as the other request attributes.
 
Also corrected the Anthropic span refinement note from `thinking.effort` to `output_config.effort` — `thinking` is for extended thinking configuration, while `output_config.effort` controls the reasoning effort level.

reference/scenarios/openai/scenario.py
katsuhisa91: Done — same approach. The value now comes from `request_reasoning_level` which is passed as `reasoning_effort` to the SDK call ([docs](https://platform.openai.com/docs/guides/reasoning)).

JWinermaSplunk (approved):

trask:

Create `changelog.d/258.enhancement.md` containing the change log entry, e.g.:

```md
Add `gen_ai.request.reasoning.level` attribute to record the requested reasoning/thinking effort level for inference and agent invocation spans. ([#189](https://github.com/open-telemetry/semantic-conventions-genai/issues/189))
```

Thanks!

katsuhisa91:
Thanks @trask! Created `changelog.d/258.enhancement.md` and removed the direct edit to `CHANGELOG.md`.
 
~~The `policies` job failed due to a Docker Hub timeout (`context deadline exceeded`) — unrelated to the code changes. Could you re-run the failed jobs when you get a chance?~~
 
~~**Update:** The `policies` job now passes after re-run. The remaining `links` CI failure is also unrelated to this PR — it's caused by a Microsoft URL in [`docs/gen-ai/azure-ai-inference.md` (line 33)](https://github.com/open-telemetry/semantic-conventions-genai/blob/main/docs/gen-ai/azure-ai-inference.md?plain=1#L33) that returns HTTP 404 after redirect (the page renders in a browser but the status code is 404). This file is not modified by this PR, so this affects all PRs running the `links` check.~~
 
**Update 2:** All CI checks now pass. Ready for review. ✨

model/gen-ai/spans.yaml
```diff
@@ -81,6 +81,9 @@ attribute_groups:
         requirement_level: recommended
       - ref: gen_ai.request.presence_penalty
         requirement_level: recommended
+      - ref: gen_ai.request.reasoning.level
+        requirement_level:
+          recommended: When applicable.
```
trask: what's the motivation for including it on agent spans? is there any reference scenarios that could be added for that? thanks

katsuhisa91: Good point, thank you. I don't have a concrete agent scenario for this — the attribute was added to `invoke_agent.common` alongside the other request parameters, but reasoning effort is really an inference-level concern. I'll remove it from agent spans and keep it on inference spans only.

trask (approved):

lmolkova (approved):

katsuhisa91:
Thank you all for the thorough review and guidance. 
This was my first contribution to this repo, and the review process was a great learning experience. Looking forward to contributing more!

Issue #249, opened 2026-06-04
Span + metric definitions principles for agents, inference, control plane


lmolkova:
Capturing discussions and proposals inspired by #215 #203 #201 

## Current span/metric/operation name definitions

| Span Type                       | Duration metric                    | `gen_ai.operation.name` values                                                                                                    |
| ------------------------------- | ---------------------------------- | --------------------------------------------------------------------------------------------------------------------------------- |
| gen_ai.inference.client         | `gen_ai.client.operation.duration` | `chat`, `generate_content`, `text_completion`                                                                                     |
| gen_ai.embeddings.client        | `gen_ai.client.operation.duration` | `embeddings`                                                                                                                      |
| gen_ai.retrieval.client         | `gen_ai.client.operation.duration` | `retrieval`                                                                                                                       |
| gen_ai.memory.client            | `gen_ai.client.operation.duration` | `create_memory_store`, `search_memory`, `create_memory`, `update_memory`, `upsert_memory`, `delete_memory`, `delete_memory_store` |
| gen_ai.create_agent.client      | `gen_ai.client.operation.duration` | `create_agent`                                                                                                                    |
| gen_ai.invoke_agent.client      | `gen_ai.client.operation.duration` | `invoke_agent`                                                                                                                    |
| gen_ai.invoke_agent.internal    | `gen_ai.client.operation.duration` | `invoke_agent`                                                                                                                    |
| gen_ai.execute_tool.internal    | `gen_ai.client.operation.duration` | `execute_tool`                                                                                                                    |
| gen_ai.invoke_workflow.internal | `gen_ai.workflow.duration`         | `invoke_workflow`                                                                                                                 |
| gen_ai.plan.internal            | `gen_ai.client.operation.duration` | `plan`                                                                                                                            |

This table is highly controversial and the modeling is problematic:
- does `gen_ai.client.operation.duration` capture tool calling or internal agents? is it documented?
- one metric definition for very different operations (not much in common between tool calling and memory updates)

## Proposal

**Naming**

Span and metric names are built from the same segments so they stay aligned:

- `span   = gen_ai.<operation-class>.<kind>` # partially enforced by semconv policies, this is not yet visible on OTLP, can be changed
- `metric = gen_ai.[client.]<operation-class>.<measured-thing>`

Rules:

1. First segment is always `gen_ai`.
[...]

Issue #243, opened 2026-06-04
How to represent multiple agents on the same telemetry


lmolkova:
A few examples:

1. Agent A is a hosted agent, it calls other agents that can be remote or in-process sub-agents.
   - here it could be possible to record hosted agent info as an entity and record 'target' agent on signal attributes. This would result in using the same attributes on resources and signals requiring backends or users to dedup / qualify which one of them they want. This is arguably confusing. 
2. Agent B is one of the agents that exist in the process and it calls other agents (remote or local)
   - here we could say that we only record the 'current' agent info on spans and metrics. It's not possible to record 'root' agent info. We can't solve this problem fully - agent hierarchy can be deep.
3. Any agent C calls another agent anywhere and we want to record the fact it happened and info of both agents on the same span / event / metric. 
   - ?

Scenario 3 is relatively popular and it makes sense to record details of both agents: 

- OpenAI agents [handoff](https://openai.github.io/openai-agents-python/handoffs/) - already trying to represent from / to agent in https://github.com/open-telemetry/semantic-conventions-genai/pull/98
- Autogen [`agent.send(message, recipient: Agent)`](https://docs.ag2.ai/latest/docs/api-reference/autogen/Agent/#autogen.Agent)
- [ADK delegation and agent-as-a-tool](https://adk.dev/agents/custom-agents/#delegation) 

We  essentially need a way to record causation for agent calls (that's not always expressed with span hierarchy because everything is async)

lmolkova:
# Option 1: introduce `gen_ai.agent.peer.*` (or similar) namespace so we can record source and target on the same telemetry.

- `gen_ai.agent.id | name | version | description` are used when recording telemetry  about operations performed by this agent.
- `gen_ai.agent.peer.id | name | version | description` are used when recording calls to this agent but not by it. E.g. `invoke_agent` (client) would always use these attributes.

This is an analogy with `service.name` and `service.peer.name`.

This can cover handoff / delegation scenarios (p3 above) and with p1 when all agents are hosted. 
It could help with in-proc hierarchies:  if agent knows who called it, it can be recorded as peer too.

Variations of this option:
- `gen_ai.agent.source | caller.id` and `gen_ai.agent.target | callee.id` (a bit more flexible and descriptive, but might be over-complicated)
- `gen_ai.sub_agent.id` - sub-agents imply hierarchy, but agents can have other relationships

# Option 2:  Separate entity and signal attributes

[...]

Issue #239, opened 2026-06-03
Opaque governance references for GenAI agent decision points


AgentGymLeader:
## Summary

Agent traces show what happened during a model call, tool call, or agent step. For enterprise agent runtimes, operators also need to correlate a decision point with producer-owned records that govern the action — policy, review, approval, workflow, or execution-context records — without carrying those payloads in telemetry.

This issue proposes a small, payload-free set of producer-defined attributes for those decision join points. The referenced records stay opaque and out of scope; the convention only defines a safe surface for downstream joins.

## Proposed attributes

| attribute | intent | notes |
| --- | --- | --- |
| `gen_ai.agent.decision.id` | correlate one decision point | opaque, producer-defined, not interpreted by the convention |
| `gen_ai.agent.decision.outcome` | record the observed result of the decision | small enum (see below) |
| `gen_ai.agent.governance.ref` | point to producer-owned governance context | opaque string; referenced context stays out of scope |

Values should be opaque handles (`ctx_7f3a9c`, `decision_01J...`), never embedded JSON, prompts, tool arguments, policy bodies, evidence hashes, or user data. These are span/event material, not metric labels (high cardinality).

## Outcome vocabulary

A first cut: `allow`, `block`, `review`, `escalate`, `defer`.

One distinction from the discussion is worth building in. Two cases look identical downstream but a reviewer reads them differently:

- **Stopped at the boundary** — the agent attempted something and a control denied it. There is a decision, a terminal `block`, and the tell is the missing child execute span.
- **Outside the reachable set** — the capability was never reachable to begin with. Nothing was attempted because there was nothing to attempt; it wasn't in the surface the agent could act over.

If the enum collapses these into `block`, the trace can't tell whether a control fired or the thing was simply never on the table. Both read as "didn't happen," but one is "stopped at the boundary" and the other is "wasn't reachable by construction," and an auditor cares which. This suggests an outcome value that separates "denied" from "outside the reachable set."

## A useful invariant (from the discussion)

A denied decision should still leave a span or event behind. The deliberate absence of a child execute span next to a terminal `block`, together with the opaque governance ref, is what makes a denial auditable after the fact. If a blocked call produces no telemetry, "denied" and "never attempted" become indistinguishable downstream.

## Scope

[...]

Issue #132, opened 2026-05-09
Proposal: gen_ai.threat.detection.* attributes for runtime threat detection on agent spans


eeee2345:
This issue proposes adding `agent.threat.detection.*` attributes to agent spans so that runtime threat-detection events on AI agents can be expressed in a vendor-neutral way.

GenAI agents emit spans for invoke_agent, execute_tool, and similar operations (already in this repo's registry.yaml under gen_ai.operation). When a runtime threat-detection layer (a guardrail, a tool-call inspector, an MCP-aware proxy, or an embedded scanner) evaluates that operation against a named detection ruleset, there is currently no convention for recording the outcome on the span. Vendors today emit ad-hoc attributes such as guardrail.rule, security.flag, or vendor-prefixed names, which fragments dashboards, alerting, and policy. Open issues #94 (Generic Grouping Attributes for Agentic Workflow Spans), #95 (MCP Tool approval), and #89 (agentic AI failure repair) all touch the same underlying gap: agent-runtime decisions need a stable schema.

Original draft proposed `gen_ai.threat.detection.*` with a four-attribute sketch. The thread discussion (see comments) resolved the two open questions:

---

**Updated shape (based on thread discussion with @luckyPipewrench and @AgentGymLeader)**

Namespace moves to `agent.*` — control points that sit outside provider spans (forward proxy, MCP proxy, CI wrapper, host-level network filter) should not be forced to misrepresent where enforcement happened. `agent.*` composes with `gen_ai.*` spans without that constraint.

Five attributes plus one event:

- `agent.threat.detection.rule_id` (string) — stable rule identifier, scoped by `agent.threat.detection.ruleset`; example: `ATR-2026-00081`, `guardrail-pii-7`
- `agent.threat.detection.ruleset` (string) — namespace + version; example: `atr@v2.2.1`, `pipelock-rules@2026.05`, `vendor-x-bundle@2025.10`. Explicitly namespace-neutral — open rulesets, proprietary bundles, and local policy packs are all peers.
- `agent.threat.detection.severity` (enum: `low` | `medium` | `high` | `critical`) — property of the rule itself, independent of enforcement policy.
- `agent.threat.detection.action` (enum: `allow` | `block` | `warn` | `ask`) — enforcement decision at detection time, not necessarily the final outcome. `ask` covers the MCP tool-approval pending state from #95 where the absence of execution is intentional and successful.
[...]

PR #262, opened 2026-06-08
gen-ai: add run guardrail span and security finding

nagkumar91:
## Overview

This proposal defines semantic conventions for observing GenAI security guardrails: policy, safety, DLP, model armor, prompt-injection, and other controls that inspect GenAI content or actions before, during, or after a GenAI operation.

The goal is to let instrumentation consistently answer:

- which guardrail ran,
- what content or action it evaluated,
- what verdict the guardrail returned,
- what action the caller enforced,
- which policy or rule produced the finding,
- and how the finding can be correlated with external security systems.

## What this adds

- `gen_ai.run_guardrail.client` for out-of-process guardrail services.
- `gen_ai.run_guardrail.internal` for in-process guardrail checks.
- `gen_ai.security.finding` for specific findings produced by a guardrail run.
- `gen_ai.security.guardrail.*` attributes for the evaluating service/component.
- `gen_ai.security.verdict.*` attributes for the result returned by the guardrail.
- `gen_ai.security.action.type` for the caller/framework enforcement action.
- `gen_ai.security.policy.*` and `gen_ai.security.policy.rule.id` for policy and rule attribution.
- `gen_ai.security.risk.finding` and `gen_ai.security.risk.score` for finding classification and scoring.
- `gen_ai.security.external_finding_id` for SIEM/security-system correlation.
- OpenAI reference coverage for both client/internal guardrail spans and the security finding.

## Current modeling choices

- Uses `guardrail` terminology throughout.
- Uses `run_guardrail` as the operation name.
- Does not include `gen_ai.agent.id` on guardrail spans; agent identity can be inferred from upstream context when available.
- Requires `gen_ai.security.policy.id` on findings.
- Models findings separately from generic evaluation results; non-security categorical evaluation outcomes should continue to use `gen_ai.evaluation.result`.

## Validation

- `make generate-all`
- `make check-policies`
- `cd reference && uv run run-scenario openai && uv run update-reports`
- `cd reference && uv tool run --from ruff==0.15.15 ruff check src/semconv_genai/classify.py src/semconv_genai/semconv_model.py src/semconv_genai/data_files.py src/semconv_genai/report.py scenarios/openai/scenario.py`


AgentGymLeader:
@nagkumar91 this is a solid shape. The verdict-vs-action split is the right call — keeping "what the guardrail returned" separate from "what the caller actually enforced" matches how these break in practice (a framework can return block and the caller still lets it through, and you want both on the trace).

Two things I'd pin down while it's still in development:

1. The blocked case should be readable from absence, not only from the finding event. If a guardrail returns block and the caller enforces it before the operation runs, the guarded op should have no success span. A terminal block plus no downstream execute span is what lets a reader tell "blocked" apart from "never ran" or "telemetry got dropped." Worth stating normatively how the guardrail span and finding link back to the operation they gated (a span relationship, not just a shared trace), so the block and the missing span are tied together. This also lines up with the span-relationship discussion in #243.

2. Keep external_finding_id an opaque, producer-defined correlation handle, and keep the finding record itself out of scope. The convention should define a reference you can join on (SIEM, incident, etc.), not reach into the schema, format, or lifecycle of the finding body. Same line that came up on #249 for evidence records: define the join point, leave the record to the producer. Current Conditionally Required wording is fine, I'd just make the "reference, not a record" intent explicit so it doesn't drift into standardizing payloads later.

Minor: action.type (allow/block/modify/escalate) overlaps the decision-outcome vocabulary in #239. Might be worth keeping them aligned so an MCP approval, a generic agent decision, and a guardrail block all read the same way in one query, instead of each backend special-casing security.

If it'd help, I'm happy to turn the opaque-reference and span-relationship wording into a small follow-up PR. Don't want to split the review though, so only if that's useful to you.

trask:

Create `changelog.d/262.component.md` containing the change log entry, e.g.:

```md
Add GenAI security guardrail span conventions and a security finding event.
```

Thanks!

nagkumar91:
Addressed the Towncrier changelog migration: moved the guardrail changelog entry to `changelog.d/262.component.md` and removed the direct `CHANGELOG.md` entry in 01b596a.

aabmass:
Hey @nagkumar91, does this address the feedback we went through in the call and in [this doc](https://docs.google.com/document/d/14CD0_ApP0W_ek9VNJN1N_2Bz5ifiJ978LTv5AD4mf_o/edit?tab=t.0)? Just lmk and I'll review and get someone from Model Armor to review.

nagkumar91:
> Hey @nagkumar91, does this address the feedback we went through in the call and in [this doc](https://docs.google.com/document/d/14CD0_ApP0W_ek9VNJN1N_2Bz5ifiJ978LTv5AD4mf_o/edit?tab=t.0)? Just lmk and I'll review and get someone from Model Armor to review.

@aabmass some yes, some still needs a discussion/votes? P1 items like file/token/modalities/protocol and agent tags are not included yet. Feel free to ask the team to review/post review comments and we can add it on.

Rul1an:
Strong direction, and the privacy posture is already careful: `content.input.value`/`.output.value` are opt_in with sensitive-content warnings, `finding.evidence` is a non-content channel with a MUST NOT on sensitive data, and `verdict.type` (what the guardrail returned) is kept separate from `action.type` (what the caller enforced). Those boundaries are right, and the `run_guardrail.client` vs `.internal` split captures out-of-process vs in-process cleanly.

`gen_ai.security.content.input.hash` is the field everyone gets pointed to as the privacy-preserving path, and as written it doesn't pin what it hashes. Its purpose is forensic correlation, but that only works if two implementations agree on the exact bytes and the algorithm, and "hash of the input content" leaves open the encoding, any normalization, and what the input even is for a structured `tool_call_input` (a JSON object, not a string). Two services inspecting the same content can produce different hashes, so the field can't do its one job.

Small fix: name a canonical byte form plus a hash. For structured targets, RFC 8785 (JCS) over the arguments object then SHA-256; for free-text, state UTF-8 and any normalization. The same pinning lets `external_finding_id` correlate deterministically rather than per-vendor. Happy to sketch one structured and one free-text example so it is testable. You may already have some of this in the design discussion, so take it as one observation from the public PR.

nagkumar91:
Thanks everyone for the feedback here. I pushed one more update in 35b2921 to fold in the parts that seemed clear enough to handle in this PR.

What changed:

- Split the target model into `gen_ai.security.target.type` and `gen_ai.security.target.subtype`.
  - `target.type` is now the broad direction: `input` or `output`.
  - `target.subtype` captures the more specific surface, for example `llm`, `tool_call`, `tool_definition`, `retrieval`, `memory`, `message`, or `action`.
- Added `target.subtype` to the guardrail spans and the `gen_ai.security.finding` event.
- Updated the reference scenario from `llm_input` to `target.type=input` + `target.subtype=llm`.
- Strengthened the `gen_ai.security.content.input.hash` guidance based on @Rul1an's point: it now calls out algorithm tagging, canonical bytes, UTF-8 / normalization for text, deterministic serialization for structured inputs such as tool calls, and HMAC/secret-keyed hashing for sensitive content.
- Clarified `gen_ai.security.content.modified` to mean the content actually changed, not just that a `modify` verdict/action was selected.
- Added wording that a finding is the semantic concept represented by this event/log signal.

I did not add file size/type, token size, modalities, protocol, `gen_ai.agent.tag`, or severity in this pass. My preference is to keep those as follow-ups unless folks feel strongly they need to block this PR. They seem useful, but I think we need clearer cross-provider semantics and concrete use cases before standardizing them here.

Same for `gen_ai.agent.tag`: I would love to understand the specific use case it enables that cannot be represented through upstream agent context (`gen_ai.agent.id`, parent spans, links, or resource/scope context). If there is a concrete gap, we can add it in a focused follow-up.

Rul1an:
This is the right shape now: tagging the algorithm, naming the canonical byte form, and splitting text (UTF-8 plus a stated normalization) from structured (deterministic serialization for a tool_call) is what lets the hash do its forensic-correlation job. The piece that keeps it working across two implementations is a named default per target subtype, not only a tag: if each service tags whatever it used, two inspectors of the same tool_call still diverge, so one default canonical form plus algorithm per subtype (JCS then SHA-256 for tool_call, UTF-8 with a named normalization then SHA-256 for text) makes an unkeyed hash reproducible without coordination. The HMAC path for sensitive content is the right call; it correlates within a shared-key domain rather than across orgs, which is the honest bound to state alongside it.

If useful I can add two worked examples as a small follow-up, one tool_call and one free-text, with the exact bytes and resulting digest so a second implementation reproduces them from the example alone. Glad to keep it out of this review if you'd rather land the convention first.

Rul1an:
Here are the two worked examples, reproducible from the bytes alone.

Structured (`tool_call`), RFC 8785 (JCS) over the input object then SHA-256:

- input: `{"name":"query_table","arguments":{"table":"employees","limit":10}}`
- canonical bytes: `{"arguments":{"limit":10,"table":"employees"},"name":"query_table"}`
- `sha256:40cfbce5aa3b2794c73996f4db4e6c2b45a5492cdc2a4bc71773e88eeebc3d95`

Free-text, Unicode NFC then UTF-8 then SHA-256, on `café au lait`:

- normalized bytes (hex): `636166c3a9206175206c616974`
- `sha256:7c413039fbb2248e2b18b98e7a8d4d85bdcac7cd79b9477a0923f97e3a1f2b50`
- the same text as NFC and as NFD (`e` + U+0301) both reach that digest once NFC-normalized; the raw NFD bytes hash to `sha256:2f97bdcfe9d8165b6d25df3f56d7b83e16637d199505d01eff4a3e94eb1d9e15` instead, which is why the normalization has to be named rather than implementation-defined.

The JCS recipe is exact for objects of strings and integers; floating-point values would also need RFC 8785 number formatting. Happy to drop these into the reference scenario or a doc example, whichever you prefer.

habibam (approved):

sjain700 (approved):

AgentGymLeader:
On correlation, I'd keep the shape from #132: alongside `external_finding_id` (SIEM-facing), keep a producer-side correlation that stays opaque and producer-defined, so the span stays lean and the heavier context lives in the producer's own store, joined by id at audit time. I'd leave the external record format unspecified rather than mandate a canonical form.

On `escalate`: it works well as a real "doesn't proceed yet" decision. Worth saying the later resolution (allow or block by a human or external workflow) is a separate event rather than an overwrite, so both the decision-time state and the outcome survive.

hemanshubelani (approved):

model/gen-ai/spans.yaml
```diff
@@ -130,6 +130,62 @@ attribute_groups:
     attributes:
       - ref_group: attributes.gen_ai.invoke_agent.common
       - ref_group: attributes.gen_ai.address_and_port
+
+  - id: attributes.gen_ai.run_guardrail.common
+    visibility: internal
+    attributes:
+      - ref_group: attributes.gen_ai.error
+      - ref: gen_ai.security.guardrail.id
```
lmolkova, line 158: do we need to nest guardrail under security? 

Semconv recommends shorter names if it does not reduce clarity, I think this is the case here

> Use shorter names when it does not compromise clarity. Drop namespace components or words in multi-word components when they are not necessary. For example, vcs.change.id describes pull request id as precisely as vcs.repository.change.id does.


https://github.com/open-telemetry/semantic-conventions/blob/main/docs/general/naming.md

model/gen-ai/spans.yaml
```diff
@@ -130,6 +130,62 @@ attribute_groups:
...
+        requirement_level:
+          conditionally_required: If available.
+      - ref: gen_ai.security.guardrail.version
+        requirement_level: recommended
+      - ref: gen_ai.security.verdict.reason
```
lmolkova, line 163: can we use guardrail as a namespace consistently? What are the scenarios for general security where these attributes could be reused? If we don't have them yet, let's start with just the guardrail namespace

model/gen-ai/spans.yaml
```diff
@@ -517,6 +573,120 @@ spans:
...
+
+      This span SHOULD be used when the guardrail evaluation is performed in the
+      same process as the instrumented GenAI operation.
+
+      Guardrail spans SHOULD be children of the GenAI span they protect, such as
+      [inference](/docs/gen-ai/gen-ai-spans.md#inference) or
+      [execute tool](/docs/gen-ai/gen-ai-spans.md#execute-tool-span) spans.
+
+      Multiple guardrail spans MAY exist under a single operation span if
+      multiple guardrails are chained.
```
lmolkova: These spans are produced by different instrumentations, and the span hierarchy usually depends on the application and which instrumentations are enabled. We can’t put requirements on that hierarchy. The core principle is that a span uses the ambient context as its parent. It usually can’t choose, inspect, or reason about that parent span. Any hierarchy guidance would create false expectations for consumers about something instrumentations don’t control.


```suggestion
```

model/gen-ai/registry.yaml
```diff
@@ -393,6 +393,354 @@ attributes:
           }
         ]
 
+    stability: development
+  - key: gen_ai.security.guardrail.id
+    type: string
+    brief: The unique identifier of the security guardrail service or component.
+    note: |
+      This identifies the service, component, or library performing the security
```
lmolkova, line 424: this would usually come through instrumentation scope name or resource attributes. Library name and version usually become instrumentation scope name and version.

model/gen-ai/registry.yaml
```diff
@@ -393,6 +393,354 @@ attributes:
...
+    stability: development
+  - key: gen_ai.security.guardrail.name
+    type: string
+    brief: Human-readable name of the security guardrail service or component.
+    note: >
+      This identifies the security evaluation service or component, not the
+      policy name. For policy names, use `gen_ai.security.policy.name`.
+    examples: ['Azure Content Safety', 'Bedrock Guardrails', 'Custom PII Filter']
+    stability: development
+  - key: gen_ai.security.guardrail.version
```
lmolkova, line 440: see comment on the id - can this come as instrumentation scope version?

model/gen-ai/registry.yaml
```diff
@@ -393,6 +393,354 @@ attributes:
...
+    note: |
+      This identifies the service, component, or library performing the security
+      evaluation, not the policy or configuration being applied.
+
+      For provider-specific guardrail or policy configuration identifiers (for
+      example, an AWS Bedrock guardrail ARN or an Azure Content Safety blocklist
+      ID), use `gen_ai.security.policy.id` instead.
+    examples: ['content-safety', 'pii-filter-v3', 'custom-guardrail-001']
+    stability: development
+  - key: gen_ai.security.guardrail.name
```
lmolkova, line 432: given it's a service / component name, can it be called `gen_ai.guardrail.component.name` ? 
This is more precise and eliminates confusion with policy that's currently clarified on all attributes in this namespace

model/gen-ai/registry.yaml
```diff
@@ -393,6 +393,354 @@ attributes:
...
+    stability: development
+  - key: gen_ai.security.guardrail.version
+    type: string
+    brief: Version of the security guardrail service or component.
+    note: >
+      This is the version of the evaluating service or component, not the policy
+      version. For policy versions, use `gen_ai.security.policy.version`.
+    examples: ['1.0.0', '2024-05-01', 'v2']
+    stability: development
+  - key: gen_ai.security.guardrail.provider.name
```
lmolkova, line 448: please reuse existing attribute for `gen_ai.provider.name` since it covers a wide set of different providers

model/gen-ai/registry.yaml
```diff
@@ -393,6 +393,354 @@ attributes:
...
+      - 'PII detected in output'
+      - 'Prompt injection attempt detected'
+      - 'Action exceeds agent permission scope'
+    stability: development
+  - key: gen_ai.security.verdict.code
+    type: string
+    brief: Provider-specific code for the security verdict.
+    examples: ['112', '403', 'AACS-PII', 'MODEL_ARMOR_SAFETY']
+    stability: development
+  - key: gen_ai.security.action.type
```
lmolkova, line 523: how `gen_ai.security.action.type` is different from `gen_ai.security.verdict.type` ? 

E.g. for allow

> The caller allowed the content or action to proceed.
> The content or action is permitted without modification.


When instrumentation would populate one vs another? why both?

I see 

>     note: >
      Use this attribute when the guardrail verdict is enforced by the caller or
      framework and the action is known. For example, a guardrail may return a
      `warn` verdict while the caller chooses to `block` the operation.

below, but how would run_guardrail span know about both ? Does it cover both the user review and the guardrail?

model/gen-ai/registry.yaml
```diff
@@ -393,6 +393,354 @@ attributes:
...
+    type: string[]
+    brief: Non-content evidence descriptors for a security finding.
+    note: |
+      This attribute is intended for low-cardinality, non-content descriptors
+      that help locate or explain a finding without recording the evaluated
+      content itself.
+
+      > [!WARNING]
+      > This attribute MUST NOT contain sensitive user content, PII, secrets, or
+      > other high-risk data.
```
lmolkova, line 666: how would instrumentation can guarantee it?

model/gen-ai/registry.yaml
```diff
@@ -393,6 +393,354 @@ attributes:
...
+    brief: Numeric risk or confidence score.
+    note: >
+      The value SHOULD be in the range `[0.0, 1.0]` when the guardrail provider
+      reports a normalized score. Instrumentations SHOULD document the scoring
+      scale when reporting provider-specific scores that are not normalized.
+    examples: [0.85, 0.95, 0.42]
+    stability: development
+  - key: gen_ai.security.finding.evidence
+    type: string[]
+    brief: Non-content evidence descriptors for a security finding.
```
lmolkova, line 658: it seems to be very broad and can contain anything that individual systems might want to report. They can as well add custom attributes instead of providing an obscure list of strings. I'd propose to remove this attribute

model/gen-ai/registry.yaml
```diff
@@ -393,6 +393,354 @@ attributes:
...
+  - key: gen_ai.security.policy.name
+    type: string
+    brief: Human-readable name of the policy that produced the verdict.
+    examples: ['PII Protection Policy', 'Financial Advice Restriction']
+    stability: development
+  - key: gen_ai.security.policy.version
+    type: string
+    brief: Version of the policy that produced the verdict.
+    examples: ['1.0', '2024-05-01']
+    stability: development
```
lmolkova, line 692: are there systems that in practice have all 3?

model/gen-ai/registry.yaml
```diff
@@ -393,6 +393,354 @@ attributes:
...
+    type: string
+    brief: Identifier of the policy rule, section, or control that produced the verdict.
+    note: >
+      Use this when `gen_ai.security.policy.id` identifies a broader policy and
+      the guardrail reports a more specific rule, section, control, or detector
+      within that policy.
+    examples: ['rule-pii-email', 'section-4.2', 'detector-jailbreak']
+    stability: development
+  - key: gen_ai.security.content.input.value
+    type: any
```
lmolkova, line 703: this needs JSON schema and formal definition

model/gen-ai/registry.yaml
```diff
@@ -393,6 +393,354 @@ attributes:
...
+      > data. Instrumentations SHOULD NOT capture this by default. Enable via
+      > opt-in configuration only.
+
+      This attribute MAY be truncated. For correlation without full content,
+      consider `gen_ai.security.content.input.hash`.
+    examples:
+      - 'Send an email to customer@example.com'
+    stability: development
+  - key: gen_ai.security.content.output.value
+    type: any
```
lmolkova, line 720: same here, need to define the schema even if it's anything - check out tool call arguments / results for an example

model/gen-ai/registry.yaml
```diff
@@ -393,6 +393,354 @@ attributes:
...
+    type: boolean
+    brief: Whether guardrail processing actually changed the content.
+    note: >
+      This describes the observed content change, not merely that
+      `gen_ai.security.verdict.type` or `gen_ai.security.action.type` was
+      `modify`. For example, a guardrail can run in modify mode but leave the
+      content unchanged.
+    examples: [true, false]
+    stability: development
+  - key: gen_ai.security.external_finding_id
```
lmolkova, line 762: this could be recorded as system-specific attribute instead and then would have clear semantics.

reference/scenarios/openai/scenario.py
```diff
@@ -265,6 +269,114 @@ def get_weather(location: str) -> str:
...
+    }
+    if host and port is not None:
+        guardrail_attrs["server.address"] = host
+        guardrail_attrs["server.port"] = port
+    with _reference_tracer.start_as_current_span(
+        "run_guardrail Azure Content Safety",
+        kind=SpanKind.CLIENT,
+        attributes=guardrail_attrs,
+    ):
+        with urlopen(health_url, timeout=5) as response:
```
lmolkova, line 492: this scenario is not related to openai, are there libraries or common shared components that we can simulate instrumentation for ?

lmolkova (commented):
Could you please  list the systems that were evaluated (that could be instrumented with run_guardrail span) and share research on which properties they have and how they map to attributes being defined.

It seems we don't have any scenarios that would show how auto-instrumentation can record this span - are there any open source libraries or clients for server-side guardrails that can be instrumented and then can we add them to reference scenarios?

model/gen-ai/registry.yaml
aabmass: I was on board with `any` actually. IIUC the guardrail can protect really generic input/outputs e.g. as part of an HTTP forward proxy for any request.

lmolkova: it can still be Any, just have a schema that says so - this helps with validation and future extensibility

lmolkova: this seems to be resolved now - there is a schema

model/gen-ai/registry.yaml
habibam: Done dropped in PR #427.

model/gen-ai/registry.yaml
habibam: Done dropped.

model/gen-ai/registry.yaml
habibam: Done dropped #427 and reused the existing  gen_ai.provider.name  instead

model/gen-ai/registry.yaml
habibam: Agreed, dropped in #427. Evidence payloads inevitably contain customer content matched substrings and offsets into the prompt/response which is exactly the sensitive context that GDPR/DLP rules keep out of telemetry. Even as a "recommended" field, instrumentation has no way to guarantee it's free of personal data, so shipping it would put an unenforceable safety burden on every implementer. Better to leave it out of the spec entirely; producers that need evidence can keep it in their own store and correlate via external_finding_id .

model/gen-ai/registry.yaml
habibam: No it would not, removed in #427 as noted above.

aabmass:
@habibam should we close this one out now?

habibam:
> @habibam should we close this one out now?

Yes, let's close it out thanks @aabmass. #427 is the successor so we can review that one going forward. I've been making incremental changes from the feedback on this PR; I'll tag you and @lmolkova on #427 with a summary of what changed and we can review again. Closing in favor of #427.

habibam:
> @habibam should we close this one out now?

@aabmass can you close this PR? Looks like I don't have the perms to

PR #267, opened 2026-06-09
Add time_budget value for gen_ai.agent.finish_reason

Nik-Reddy:
## Summary

Draft follow-up to #238 for issue #171.

This PR is intentionally a placeholder for the next focused follow-up: adding `time_budget` as a well-known value for `gen_ai.agent.finish_reason`.

## Planned scope

- Add `time_budget` to the `gen_ai.agent.finish_reason` well-known values.
- Document it as the value for agent loop termination caused by an implementation-defined wall-clock budget or timeout.
- Add one reference scenario, likely using CrewAI if it remains the cleanest public SDK timeout boundary.
- Regenerate docs, schema snapshot, and reference reports.

## Out of scope

This follow-up will not include `handoff`, `tool_failure`, `cancelled`, `error`, `max_tokens`, or budget sub-attributes. Those still need separate span-scope, precedence, or SDK-boundary discussion.

Refs #171.
Follows #238.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,11 @@
 
 ### 💡 Enhancements 💡
 
+- Add `gen_ai.agent.finish_reason` attribute on agent invocation spans
+  (`invoke_agent.client`, `invoke_agent.internal`) with initial enum values
+  `completed`, `guardrail`, and `max_iterations`. Records why the agent loop
+  terminated; complements per-LLM-call `gen_ai.response.finish_reasons`.
+  ([#171](https://github.com/open-telemetry/semantic-conventions-genai/issues/171))
 - Add GenAI memory operation span and attributes for memory store lifecycle (create/delete),
   memory record create/update/upsert/search/delete operations, and record counts.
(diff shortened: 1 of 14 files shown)


trask:

Create `changelog.d/267.enhancement.md` containing the change log entry, e.g.:

```md
Add `gen_ai.agent.finish_reason` attribute on agent invocation spans (`invoke_agent.client`, `invoke_agent.internal`) with initial enum values `completed`, `guardrail`, and `max_iterations`. Records why the agent loop terminated; complements per-LLM-call `gen_ai.response.finish_reasons`. ([#171](https://github.com/open-telemetry/semantic-conventions-genai/issues/171))
```

Thanks!

Nik-Reddy:
Closing this in favor of the direction #238 landed on. gen_ai.agent.finish_reason was removed during review and replaced with gen_ai.agent.invocation.status (completed only); failure and termination causes, including a timeout, are represented via error.type on the span rather than a dedicated enum value. That makes the planned time_budget value unnecessary. If a CrewAI timeout reference scenario is still wanted, that would be a small independent follow-up, not a semconv change.

PR #270, opened 2026-06-09
Add `gen_ai.main_agent` entity

aabmass:
Propose GenAI agent entity

## Description

This PR defines the new `gen_ai.main_agent` entity in the GenAI semantic conventions, which allows an Agent process to identify itself via **Resource attributes**. Part of #243 to clarify agent attributes and resource association.

- Defines the `gen_ai.main_agent` entity, with `gen_ai.main_agent.id` as the identifying attribute, and `gen_ai.main_agent.name` and `gen_ai.main_agent.description` as descriptive attributes.
- Updates the markdown generator templates (`snippet.md.j2` and `entity_macros.j2`) to support the Weaver v2 schema structure for entities.
- Adds a new documentation page `docs/registry/entities/gen-ai.md` to host the generated entity table.

## Motivation

- **User journey:** This change introduces `gen_ai.main_agent.id` as a Resource attribute to identify the main agent running in a process. This allows differentiating the logical process-level top-level agent from any subagents or helper agents invoked within the same process (which are tracked via span attributes). See #243.
- **Prior art:** Cloud providers and frameworks use resource-level identifiers to correlate system metrics and telemetry with the logical agent process. For example, [AWS Bedrock uses ARN (a type of URN)](https://docs.aws.amazon.com/bedrock/latest/APIReference/API_agent_Agent.html) for agent resources, and [GCP Agent Registry](https://docs.cloud.google.com/agent-registry/concepts) uses URNs for identifiers. Both of these apply to a "deployment" level, so it is not an instance-level identifier.

## Prototype

N/A. Resource attributes are not part of instrumentation generally.

## Checklist

- [ ] Reference scenarios updated for affected libraries (N/A for this schema-only PR)
- [x] Changelog fragment added under `changelog.d/`


aabmass:
Waiting on https://github.com/open-telemetry/semantic-conventions-genai/pull/273

model/gen-ai/entities.yaml
```diff
@@ -0,0 +1,13 @@
+file_format: definition/2
+entities:
+  - type: gen_ai.agent
+    stability: development
+    brief: >
+      A Generative AI agent.
+    identity:
```
lmolkova, line 21: should we clarify how/when to use this entity ?

```suggestion
    note: >
      This entity SHOULD be used to capture the active agent when it's possible
      to associate an OpenTelemetry SDK instance with a specific agent.

      When multiple top-level agents can be active within the scope of a single
      OpenTelemetry SDK instance, these attributes MAY be recorded on the
      corresponding telemetry items instead.
    identity:
```

E.g. on the `internal_invoke` span we should probably mention that agent.id / name / description / version should be populated on the entity whenever possible 

And once we add `gen_ai.invoke_agent.duration`, we'd mention agent entity in the associated entities.

lmolkova:
> #273

I don't see a problem in bundling these two together, otherwise it's hard to review either of them

model/gen-ai/entities.yaml
aabmass: 👍 I applied the diff.

> we should probably mention that agent.id / name / description / version should be populated on the entity whenever possible

Can I just use the `entity_associations` on the relevant internal spans?

aabmass: Added in 0bc77a0b5982cce5b7a53f9f2d372ea6cbfb677d

templates/registry/markdown/weaver.yaml
```diff
@@ -28,6 +28,22 @@ templates:
...
+      | map({ root_namespace: .[0].root_namespace, entities: . | sort_by(.type) })) catch empty
+    application_mode: single
+    file_name: "entities/README.md"
+  - template: entity_namespace.md.j2
+    filter: >
+      try (semconv_signal("entity"; {"v2":true, "exclude_deprecated":true})
+      | group_by(.root_namespace)
+      | map({ root_namespace: .[0].root_namespace, entities: . | sort_by(.type) })) catch empty
+    application_mode: each
+    file_name: "entities/{{ ctx.root_namespace | kebab_case }}.md"
```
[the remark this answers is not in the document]
lmolkova: either one should work

trask:

Create `changelog.d/270.enhancement.md` containing the change log entry, e.g.:

```md
Add `gen_ai.agent` entity definition and resource attributes, and update generator templates to support entity documentation.
```

Thanks!

templates/registry/markdown/weaver.yaml
```diff
@@ -28,6 +28,22 @@ templates:
...
     file_name: README.md
 
+  # entity registry
+  - template: entity_readme.md.j2
+    filter: >
+      try (semconv_signal("entity"; {"v2":true, "exclude_deprecated":true})
+      | group_by(.root_namespace)
+      | map({ root_namespace: .[0].root_namespace, entities: . | sort_by(.type) })) catch empty
+    application_mode: single
+    file_name: "entities/README.md"
```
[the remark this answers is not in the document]
aabmass: As mentioned in other comment, they both work

CHANGELOG.md
```diff
@@ -17,6 +17,7 @@
 
 ### 💡 Enhancements 💡
 
+- Add `gen_ai.agent` entity definition and resource attributes, and update generator templates to support entity documentation.
 - Add GenAI memory operation span and attributes for memory store lifecycle (create/delete),
```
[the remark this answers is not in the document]
aabmass: fixed the desc

templates/registry/markdown/snippet.md.j2
```diff
@@ -23,15 +25,18 @@
...
+{{ sp.header(span, entity_registry_base_url) }}{{ at.generate(_attrs, attribute_registry_base_url, upstream_docs_base, "Attributes") }}
 {%- elif _instrument -%}
 {%- set metric = {"name": _name, "instrument": _instrument, "unit": (unit | default("")), "stability": _stability, "deprecated": _deprecated, "brief": _brief, "note": _note, "entity_associations": (entity_associations | default([]))} -%}
 {{ mt.generate(metric) }}
 {{ at.generate(_attrs, attribute_registry_base_url, upstream_docs_base, "Attributes") }}
+{%- elif (identity is defined or description is defined) and not _kind and not _instrument -%}
+{%- set entity_obj = {"name": type, "stability": _stability, "deprecated": _deprecated, "brief": _brief, "note": _note, "identity": identity | default([]), "description": description | default([])} -%}
+{{ entity.header(entity_obj) }}{{ entity.all_attribute_table(entity_obj, attribute_registry_base_url, upstream_docs_base) }}
```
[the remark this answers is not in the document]
aabmass: Fixed

model/gen-ai/entities.yaml
```diff
@@ -0,0 +1,20 @@
...
+
+      When multiple top-level agents can be active within the scope of a single
+      OpenTelemetry SDK instance, these attributes MAY be recorded on the
+      corresponding telemetry items instead.
+    identity:
+      - ref: gen_ai.agent.id
+        requirement_level: required
+    description:
+      - ref: gen_ai.agent.name
+      - ref: gen_ai.agent.version
```
aabmass: I think I'm going to remove `gen_ai.agent.version`.

Looking at the examples in https://github.com/open-telemetry/semantic-conventions/pull/3428, `gen_ai.agent.version` was originally only supposed to be on client spans when creating or invoking a **remote** agent.

model/gen-ai/spans.yaml
lmolkova, line 573:
```suggestion
        sampling_relevant: true
        requirement_level:
          recommended: When instrumentation can't populate `gen_ai.agent` entity.
```

see my other comment

```suggestion
          recommended: When available and when instrumentation can't populate `gen_ai.agent` entity.
```

I.e. only server-side / local tool instrumentations (today) can skip it and once https://github.com/open-telemetry/opentelemetry-specification/blob/main/oteps/entities/4665-multiple-resource-in-sdk.md lands in the spec and SDKs, it would be possible to use in other instrumentations.

Softening requirement level because it becomes difficult to require / expect attribute.

lmolkova (approved):
LGTM with a couple of suggestions to clarify entity + agent info attributes on spans

AgentGymLeader:
@lmolkova @aabmass the split here looks right to me: use the `gen_ai.agent` entity for the active local/server-side agent when instrumentation can bind an SDK instance to that agent, and keep span-level `gen_ai.agent.*` as the fallback when the entity cannot be populated.

One wording tweak I would preserve is that the span attributes describe the agent relevant to that telemetry item only when the entity path is not available for that instrumentation. That keeps client/remote invocation from having to fake an entity today, while still making the entity the preferred shape as SDK support catches up.

aabmass:
Hey @AgentGymLeader are you using an AI agent to leave comments directly? We appreciate people contributing but prefer if comments are human driven to avoid noise.

If I'm mistaken, sorry to presume, and feel free to join our Slack and weekly calls to say hello https://github.com/open-telemetry/community#sig-genai-instrumentation!

model/gen-ai/entities.yaml
aabmass: Good to resolve this one?

AgentGymLeader:
@aabmass 
Yeah, you guessed right. I'm Japanese — that's my native language. My own English comes out pretty broken, and to be honest it gets in the way of my thinking, so I write in Japanese first and use AI to help me put it into English. That's how I try to keep pace with native speakers here.

One downside is that the AI sometimes makes it sound stronger or smoother than I actually mean — or just a bit unnatural. If I'm not catching those nuances each time, that's on me to fix. But I read everything and post it myself, so it's me here, not a bot commenting on its own.

Sorry if it came across as noise. If something sounds off, or harsher than I meant, just tell me and I'll fix it.

And thanks for the invite — I'll drop by the Slack for sure. To be honest the weekly calls are a bit intimidating for me, since speaking English live isn't something I can do yet, so I'll mostly stick to text for now. But I'll come say hi in Slack.

aabmass:
No need to apologize, and sorry if I made some false presumptions. I don't want to discourage you from contributing at all, I was just surprised how active you are. Feel free to also approve PRs as a signal you agree with them, by the way.

Thanks for your contributions 😃

model/gen-ai/entities.yaml
lmolkova: sure, please resolve when you're confident it was resolved

model/gen-ai/entities.yaml
```diff
@@ -0,0 +1,19 @@
...
+
+      When multiple top-level agents can be active within the scope of a single
+      OpenTelemetry SDK instance, these attributes MAY be recorded on the
+      corresponding telemetry items instead.
+    identity:
+      - ref: gen_ai.agent.id
+        requirement_level: required
+    description:
+      - ref: gen_ai.agent.name
+      - ref: gen_ai.agent.description
```
trask: I don't recall running into this before, where an attribute is expected to be on both span/metric and its resource with different values, e.g.

span/metric attributes:
- `gen_ai.agent.id` - `89723497823`
- `gen_ai.agent.name` - `my nested agent`

resource attributes:
- `gen_ai.agent.id` - `273894783`
- `gen_ai.agent.name` - `my main agent`

which might(?) cause issues in Prometheus and other backends

aabmass: I agree it's not ideal. Some options


1. `gen_ai.agent.id` - Since https://github.com/open-telemetry/semantic-conventions-genai/pull/242, it's clear that `gen_ai.agent.id` is static and 1:1 with a process and only needs to be on resource.
   - For internal invoke agent spans, we can deprecate it.
   - For remote invoke agent client spans, we can add "peer" to all of these attributes to disambiguate (https://github.com/open-telemetry/semantic-conventions-genai/issues/243#issuecomment-4619229448) -> `gen_ai.agent.peer.id`. That's in line with `service.peer.*` (https://github.com/open-telemetry/semantic-conventions/pull/3097).
1. `gen_ai.agent.name` - This is already widely adopted for naming subagents in the same process.
   - Option 1, rename the telemetry level attribute to indicate that it identifies the local subagent -> `gen_ai.subagent.name`. (this is my preference if we had no tech debt)
   - Option 2, since that would break lots of existing usage of this attribute, rename the entity/resource attribute in this PR with "root" or something similar -> `gen_ai.root_agent.name`

lmolkova: subagents can be internal and won't know if they are subagents or main ones

Ideally it could work like this (in the future once/if https://github.com/open-telemetry/opentelemetry-specification/blob/main/oteps/entities/4665-multiple-resource-in-sdk.md is implemented): 

```
- invoke_agent (main agent entity)
  - llm stuff
  - agent-as-tool 
    - invoke_agent (sub agent entity) # if not entity, we still need to record it, option 1 breaks here
      - llm stuff
      - agent-as-tool 
         - invoke_agent (sub-sub agent entity) # Option 2.1 breaks here
```

In this case nobody uses resource attributes or agent.name on span / metric attributes. Each agent object instance has it's own tracer/meter with specific to this agent entity.

----------------

I see 3 options:

1. don't even introduce entity until we can support different entities for in-process agents. agent attributes always go on signals. 
2. Option 2.2 - main / root agent entity. Main/root agent spans and metrics will have the same info in resource  attrs and signal attrs.
    - Not sure if it'll be useful if/when multi-entity support will happen.
3. Current shape of this PR (same attributes on resource and signals) and tolerate the pain until multiple entities are supported

---

to this point:

> which might(?) cause issues in Prometheus and other backends

Prometheus should be fine - it [doesn't (at least yet) understand](https://github.com/open-telemetry/opentelemetry-specification/blob/efceef79720efc238645666139a7cee1b0145df1/specification/compatibility/prometheus_and_openmetrics.md#resource-attributes-1) resource attributes.

lmolkova: another thought: we can't even assign entity to in-process subagents now - `gen_ai.agent.id` does not make sense for them. 

Maybe we should have `agent` entity (name+version) and `hosted_agent` (the one with id)  entities? This aligns with option 2.2.

aabmass: > subagents can be internal and won't know if they are subagents or main ones

Agreed, the application author/framework has to choose one to put in the resource attributes.

> Maybe we should have `agent` entity (name+version) and `hosted_agent` (the one with id) entities?

Why would hosted agent not have a name?

lmolkova: sorry, I meant that only hosted agent would have an id (plus name and version), sub-/local agents would not have static id

Nik-Reddy:
Could we make the local-agent case explicit in the entity doc? Something like: `Instrumentation MUST NOT generate gen_ai.agent.id from an in-memory/process-local identifier. If no stable id is available, the agent entity is not emitted, and instrumentation continues to report gen_ai.agent.name at the span level when relevant.`

That would give implementers a clear rule and avoid high-cardinality synthetic IDs while keeping the entity reserved for stable agent identities.

model/gen-ai/entities.yaml
aabmass: Chatted offline, two proposal to start:
1. Say the outermost agent (name TBD some options are "server", "main", "hosted", "exposed", "external", "root"). 
1. No entity at all

---

Re conflict with telemetry level, we can use a different key e.g. `gen_ai.server_agent.*` to prevent flattening issues.

lmolkova: SIG discussion:

Now:
  entity (main agent), resource for now: gen_ai.server.agent
    - gen_ai.server.agent.id
    - gen_ai.server.agent.name
    ...
    
  on signals (internal):
    - gen_ai.agent.name

  on signals (remote): - different problem
    - gen_ai.agent.name # current agent executing operation
    - gen_ai.<another>.agent.name # remote peer


---- 

Future:
  - gen_ai.agent (any agent) and gen_ai.server.agent (externally visible one)
  - remove attributes from signals

aabmass:
@Nik-Reddy sure, let me copy what we used in https://github.com/open-telemetry/semantic-conventions-genai/pull/242

model/gen-ai/entities.yaml
aabmass: Updated to `gen_ai.server.agent.*` throughout and updated the descriptions to make it all clear that this is an external/server agent exposed to others.

Do you mind taking another look ?

AgentGymLeader (approved):

model/gen-ai/entities.yaml
```diff
@@ -0,0 +1,25 @@
+file_format: definition/2
+entities:
+  - type: gen_ai.server.agent
```
trask: I kind of like `gen_ai.hosted_agent`.

It feels a bit more descriptive to me.

But also, if we apply the pattern discussed in https://github.com/open-telemetry/semantic-conventions/issues/3472 and https://github.com/open-telemetry/semantic-conventions/issues/3791, we can model interactions between hosted agents as:

- `client.gen_ai.hosted_agent.name`
- `server.gen_ai.hosted_agent.name`

(which starts to get very confusing with `server.gen_ai.hosted.server.agent`)

model/gen-ai/entities.yaml
```diff
@@ -0,0 +1,25 @@
...
+    stability: development
+    brief: >
+      The Generative AI server agent exposed to external callers.
+    note: >
+      This entity identifies the externally visible agent service (for example,
```
trask: why the limitation on it being externally visible? and what does externally visible mean? maybe there's another way to phrase? AI gave me:

> an agent identity that exists as a stable resource outside a single in-process execution, and that callers/operators can recognize as the agent being invoked.

aabmass: > and that callers/operators can recognize as the agent being invoked.

That's the idea, callable by something external. I like "exposed" or "exposed as a server".

Do you generally agree with this as the entity definition, and we just need to settle on a name?

model/gen-ai/entities.yaml
aabmass: IMO hosted has the connotation of being a managed service or some cloud offering, but I don't want to bikeshed if we can agree on something.

> But also, if we apply the pattern discussed in [open-telemetry/semantic-conventions#3472](https://github.com/open-telemetry/semantic-conventions/issues/3472) and [open-telemetry/semantic-conventions#3791](https://github.com/open-telemetry/semantic-conventions/issues/3791), we can model interactions between hosted agents as:

Could we just have `{client,server}.gen_ai.agent.name`? That would be analogous to `server.service.name`.

trask: > IMO hosted has the connotation of being a managed service or some cloud offering

👍

maybe then `host_agent` or `main_agent`?

> Could we just have `{client,server}.gen_ai.agent.name`? That would be analogous to `server.service.name`.

It's not clear whether `server.gen_ai.agent.name` on a CLIENT span refers to the host agent or the remote agent it's talking to.

(in general, this pattern takes a resource/entity attribute and shows how to apply it on a cross-resource/cross-entity interaction)

model/gen-ai/entities.yaml
trask: what about an agent that wakes up every hour and kicks off some multi-agent orchestration? it seems from the definition that wouldn't be considered "exposed as a server", but not sure why we wouldn't want it to be an "entity"

model/gen-ai/entities.yaml
aabmass: Updated to `main_agent` and tweaked the wording.

model/gen-ai/entities.yaml
aabmass: Mentioned this in the wording, lmk if it looks OK.

model/gen-ai/spans.yaml
```diff
@@ -621,7 +631,7 @@ spans:
       - ref: gen_ai.agent.name
         brief: The human-readable name of the agent performing the planning.
         requirement_level:
-          conditionally_required: When available.
+          recommended: When available and when instrumentation can't populate `gen_ai.main_agent` entity.
```
lmolkova: sorry I think I proposed this clarification at some point, but I think it does not make sense in the current world of single resource per process and especially when we have different attributes for main agent and 'current' agent.

Assuming multi-tenant entities and entity API becomes a reality, we'll need to do a lot of changes and we can add clarification then.

So proposing to remove it from everywhere for now


```suggestion
          recommended: When available.
```

model/gen-ai/registry.yaml
```diff
@@ -306,6 +306,37 @@ attributes:
     brief: The version of the GenAI agent.
     examples: ["1.0.0", "2025-05-01"]
     stability: development
+  - key: gen_ai.main_agent.id
+    type: string
+    brief: The unique and stable identifier of the GenAI main agent resource.
```
```suggestion
    brief: The unique and stable identifier of the top-level Generative AI agent in the process
```

model/gen-ai/registry.yaml
```diff
@@ -306,6 +306,37 @@ attributes:
...
+      Instrumentation MUST NOT generate `gen_ai.main_agent.id` from an in-memory or
+      process-local identifier. If no stable identifier is available, the agent entity
+      is not emitted, and instrumentation continues to report `gen_ai.agent.name` at
+      the span level when relevant.
+
+    examples: ['asst_5j66UpCpwteGg4YSxUnt7lPY', 'arn:aws:bedrock:us-east-1:123:agent/42', 'urn:agent:projects-123:projects:123:locations:us-east1:aiplatform:reasoningEngines:456']
+    stability: development
+  - key: gen_ai.main_agent.name
+    type: string
+    brief: The human-readable name of the GenAI main agent.
```
```suggestion
    brief: The human-readable name of the top-level Generative AI agent in the process.
```

model/gen-ai/registry.yaml
```diff
@@ -306,6 +306,37 @@ attributes:
...
+    note: >
+      In Agent-to-Agent (A2A) communication, this maps to the `name` property declared
+      in the Agent Card.
+    examples: ["Math Tutor", "Fiction Writer"]
+    stability: development
+  - key: gen_ai.main_agent.description
+    type: string
+    brief: The free-form description of the GenAI main agent.
```
```suggestion
    brief: The free-form description of the top-level Generative AI agent in the process.
```

model/gen-ai/entities.yaml
```diff
@@ -0,0 +1,26 @@
...
+      declared in an Agent-to-Agent (A2A) Agent Card). It SHOULD be used when it is
+      possible to associate an OpenTelemetry SDK instance with that main agent.
+
+      As noted on [gen_ai.main_agent.id](/docs/registry/attributes/gen-ai.md),
+      transient in-memory or process-local identifiers MUST NOT be used. If no stable
+      identifier is available, the main agent entity MUST NOT be emitted.
+
+      When internal helper agents, subagents, or orchestrators are active within the
+      service process, their attributes SHOULD be recorded on signal-level telemetry
+      items (such as [gen_ai.agent.name](/docs/registry/attributes/gen-ai.md)) instead.
```
lmolkova: this also applies to the main agent, the framework-level instrumentation has no knowledge if its the 'main' one. 


```suggestion
      In-process instrumentations that record telemetry for `invoke_agent` operation 
      should use `gen_ai.agent.*` attributes such as   
      [gen_ai.agent.name](/docs/registry/attributes/gen-ai.md)) instead.
```

lmolkova (approved):

changelog.d/270.enhancement.md
```diff
@@ -0,0 +1 @@
+Add `gen_ai.main_agent` entity definition and resource attributes, and update generator templates to support entity documentation.
```
lmolkova, line 1: could you please update PR title and description too?

model/gen-ai/entities.yaml
aabmass: Going to resolve this since the thread is long. We're now at `gen_ai.main_agent.*` to avoid coupling the entity to something that runs a server e.g. for cron jobs.

model/gen-ai/entities.yaml
aabmass: We went with `gen_ai.main_agent.*`, going to resolve this thread since it's also outdated.

aabmass:
From SIG: let's try a scenario using an A2A server with resource attributes being called by a client

aabmass:
@trask I added some reference scenarios in bfe9b01 and updated the reports to render the entities in a009bee720bcb03d50c7c808eab34626681bbb7c. LMK if it makes sense.

model/gen-ai/spans.yaml
```diff
@@ -1,762 +1,772 @@
...
+    attributes:
+      - ref_group: attributes.gen_ai.invoke_agent.common
+      - ref: gen_ai.operation.name
+        sampling_relevant: true
+      - ref: gen_ai.request.model
+        sampling_relevant: true
+      - ref: gen_ai.agent.name
+        sampling_relevant: true
+        requirement_level:
+          recommended: When instrumentation can't populate `gen_ai.main_agent` entity.
```
trask: why? and can instrumentation know this?

reference/scenarios/a2a/scenario.py
```diff
@@ -0,0 +1,146 @@
...
+        async for _ in client.send_message(req):
+            pass
+
+
+def main():
+    print("=== A2A Agent Showcase (`gen_ai.main_agent` entity + ASGI transport) ===")
+
+    resource = Resource.create(
+        {
+            "gen_ai.main_agent.id": "urn:agent:travel-assistant:v1",
```
trask: ## Reference evaluation

> Evaluated with the proposed [`reference` skill](https://github.com/open-telemetry/semantic-conventions-genai/pull/377) against the authoritative PR diff.

### Findings

1. **Blocking — A2A cannot provide the required entity identity.**  
   [`gen_ai.main_agent.id`](https://github.com/aabmass/semantic-conventions-genai/blob/a009bee720bcb03d50c7c808eab34626681bbb7c/reference/scenarios/a2a/scenario.py#L132-L138) is a fabricated URN. A2A `AgentCard` exposes `name` and `description`, but no stable identifier. Because the model requires `id` and prohibits transient or generated identifiers, this scenario must not claim `gen_ai.main_agent` coverage.

2. **Blocking — no credible reference remains for the new entity associations.**  
   The model associates `gen_ai.main_agent` with internal invoke-agent, execute-tool, workflow, and plan spans, but A2A is the only updated scenario and cannot emit the required identity. Add coverage from a supporting library with a stable identifier or record the associations as honest capture gaps.

3. **High — A2A falsely claims execute-tool support.**  
   The [`execute_tool` span](https://github.com/aabmass/semantic-conventions-genai/blob/a009bee720bcb03d50c7c808eab34626681bbb7c/reference/scenarios/a2a/scenario.py#L47-L56) wraps only `pass`; every tool attribute is synthetic. Remove this coverage or invoke a real tool boundary.

4. **High — invoke-agent attributes are not captured from the operation.**  
   [`conversation.id`, input, and output](https://github.com/aabmass/semantic-conventions-genai/blob/a009bee720bcb03d50c7c808eab34626681bbb7c/reference/scenarios/a2a/scenario.py#L35-L70) are hardcoded despite direct sources in `context.context_id`, `context.message`, and the emitted `Message`. The span also closes before `enqueue_event`, so it does not wrap the complete operation.

5. **Medium — the claimed client span is internal.**  
   [`start_as_current_span`](https://github.com/aabmass/semantic-conventions-genai/blob/a009bee720bcb03d50c7c808eab34626681bbb7c/reference/scenarios/a2a/scenario.py#L108-L126) omits `kind=SpanKind.CLIENT`, so the report classifies it as `invoke_agent_internal`.

### Convention changes

- Add the `gen_ai.main_agent` entity.
- Require `gen_ai.main_agent.id`; recommend `name` and `description`.
- Associate the entity with internal invoke-agent, execute-tool, workflow, and plan spans.
- Change relevant `gen_ai.agent.name` requirement levels.

### Field classification

| Field | Classification | Current-call source |
| --- | --- | --- |
| `gen_ai.main_agent.id` | capture gap | A2A `AgentCard` has no stable ID; the URN is test scaffolding |
| `gen_ai.main_agent.name` | direct, implementation needs fixing | `card.name` |
| `gen_ai.main_agent.description` | direct, implementation needs fixing | `card.description` |

### Libraries not updated

| Library | Evaluation |
| --- | --- |
| `aws-bedrock-agent` | `gen_ai.main_agent.id` is direct from the stable `agentId` invocation argument; add reference coverage, noting that the current scenario emits client spans rather than the new internal associations |
| `agent-framework` | weak; its scenario uses an application-supplied `"weather-agent"` ID with no demonstrated stability |
| `azure-ai-foundry`, `openai-assistants` | honest capture gap in the current scenarios; provider IDs are returned only after ephemeral agent creation, after the OTel resource is initialized |
| `autogen`, `crewai`, `google-adk`, `openai-agents`, `pydantic-ai`, `langchain` | honest capture gap; current examples expose names or process-local objects but no stable main-agent identity |

### Capture gaps

- **A2A:** needs a stable identifier exposed by the Agent Card or server configuration; none is available.
- **Internal framework spans:** need a stable, non-process-local main-agent identifier available when the OTel resource is created.
- **Provider-created agents:** need an existing persistent agent identity available before telemetry-provider initialization.

### Validation

Reviewed the current `gh pr diff`, the A2A SDK `AgentCard` and `RequestContext` surfaces, and existing agent/workflow/plan scenarios. No test suite was run. The PR currently conflicts with `main`.

lmolkova: addressed by showing how main_agent.id can be captured either
- from a hosting platform via env var or any other platform-specific config
- from AgentCard.url which can reasonably satisfy id requirements too.

removed manual spans, just shows how to set the entity and spans come from native ADK

lmolkova:
WRT relationship to #355 - entertaining the idea of using `gen_ai.main_agent.name` instead of`gen_ai.root_operation.name` in #385:

- main_agent is limited to in-process propagation. It's interesting to have origin agent (or maybe end-to-end task name), but it's additive and should not break 'my local boss' identification. I.e. if we do this, it should come through some other attribute(s)
- since main_agent is local, the worst case is that main agent would show up as entity and signal attribute, but they will have the same value
- entity can only be set by infra / application code and can only apply to cases when main agent is global for the whole process.
- signal attributes can be set by all agentic / inference instrumentations regardless of hosting model

So with the caveat that in both cases main agent is "local root" I like this. As a side-effect we should probably reconsider having individual span type and metrics for workflow - it's a variation of agentic scenario, but this should not be a blocker for this PR.

reference/scenarios/a2a/scenario.py
aabmass: This thread is really old and Liudmila pushed some fixes, resolving.

Issue #281, opened 2026-06-10
Add metrics to reference scenarios


lmolkova:
We don't have any  yet

AgentGymLeader:
Took a first pass at this — opened a draft PR (#283) that adds metrics to the `anthropic` scenario to start: token usage and operation duration, with a small Metrics section in the README sitting next to the Spans and Events ones.

One thing I wasn't sure about: whether you'd want metrics as their own section like I did it here, or folded into the existing span reports. Easy to switch either way. And happy to extend it to the other libraries once the shape feels right.

Necmttn:
One reference-scenario edge case I would include: token usage should still join back to the agent session when it does not arrive on the same span as the model call.

PR #283 looks like a good first metrics pass, and I saw the per-data-point double-counting fix. The next useful acceptance case is cross-signal correlation:

- usage arrives as metrics or logs/events with a stable conversation/session id
- the report shows how that usage joins back to the agent/session span
- token counts stay separate from derived USD cost and pricing provenance
- transcript-derived usage and OTLP-derived usage are not silently double-counted
- cache, reasoning, and tool-token dimensions remain visible when providers emit them

That would make the reference scenario cover the case many agent harnesses hit in practice: spans, token usage, and transcript records are produced by different parts of the system, but downstream tools still need one session-level usage view.

---

_Generated with [ax](https://github.com/Necmttn/ax)._

PR #283, opened 2026-06-10
Add GenAI client metrics to the anthropic reference scenario

AgentGymLeader:
## Description

Adds the first metrics coverage to the reference scenarios. The `anthropic` chat scenario now emits the two core GenAI client metrics — `gen_ai.client.token.usage` and `gen_ai.client.operation.duration` — alongside its existing spans and events, and the tooling captures and reports them. Addresses #281, starting with `anthropic` as the first library.

## Motivation

The reference scenarios validate which libraries produce which telemetry. Spans and events were covered, but metrics had no coverage at all: there was no meter helper in the shared framework and no Metrics report. This adds the missing third signal so the reference data and the README status reflect metric support, not just spans and events.

## Prototype

The `anthropic` scenario emits the metrics; the committed `data.json` and the new README **Metrics** section are regenerated from a real run (`uv run run-scenario anthropic && uv run update-reports`). A new Metrics report section is rendered parallel to the existing Spans and Events tables.

## Open questions

- Started with `anthropic` and the two client metrics as a minimal slice — happy to fan out to other libraries in follow-ups.
- Rendered metrics as their own report section; happy to fold them into the per-span reports instead if you prefer a different shape.

## Checklist

- [x] Reference scenarios updated for the affected library (`anthropic`)
- [ ] Towncrier fragment — not needed (reference tooling, no convention change)

### DIFF
--- a/reference/reports/gen-ai-client-token-usage-metric.md
+++ b/reference/reports/gen-ai-client-token-usage-metric.md
@@ -1,0 +1,27 @@
+# Client Token Usage Metric
+
+> **[Semantic Convention](../../docs/gen-ai/gen-ai-metrics.md#metric-gen_aiclienttokenusage)**
+
+## Required
+
+| Attribute | Supporting Libraries |
+| --- | --- |
+| gen_ai.operation.name | [anthropic] |
+| gen_ai.provider.name | [anthropic] |
+| gen_ai.token.type | [anthropic] |
+
+## Conditionally Required
+
+| Attribute | Supporting Libraries |
+| --- | --- |
+| gen_ai.request.model | [anthropic] |
+| server.port | [anthropic] |
+
+## Recommended
+
+| Attribute | Supporting Libraries |
+| --- | --- |
+| gen_ai.response.model | [anthropic] |
+| server.address | [anthropic] |
+
+[anthropic]: ../scenarios/anthropic/scenario.py
--- a/reference/scenarios/anthropic/scenario.py
+++ b/reference/scenarios/anthropic/scenario.py
@@ -71,8 +71,11 @@ def run_chat():
         if port is not None:
             metric_attributes["server.port"] = port
         if resp.usage:
+            cache_creation = getattr(resp.usage, "cache_creation_input_tokens", None) or 0
+            cache_read = getattr(resp.usage, "cache_read_input_tokens", None) or 0
+            total_input = resp.usage.input_tokens + cache_creation + cache_read
             token_usage_histogram.record(
-                resp.usage.input_tokens,
+                total_input,
                 attributes={**metric_attributes, "gen_ai.token.type": "input"},
             )
             token_usage_histogram.record(
--- a/reference/scenarios/anthropic/scenario.py
+++ b/reference/scenarios/anthropic/scenario.py
@@ -18,7 +18,7 @@ from reference_shared import (
 )
 
 MOCK_BASE_URL = os.environ["MOCK_LLM_URL"]
-TOKEN_BUCKET_BOUNDARIES = [1, 4, 16, 64, 256, 1024, 4096]
+TOKEN_BUCKET_BOUNDARIES = [1, 4, 16, 64, 256, 1024, 4096, 16384, 65536, 262144, 1048576, 4194304, 16777216, 67108864]
 
 _reference_tracer = reference_tracer()
 _reference_meter = reference_meter()
--- a/reference/scenarios/anthropic/scenario.py
+++ b/reference/scenarios/anthropic/scenario.py
@@ -244,6 +244,42 @@ def run_chat_with_document_input():
         print(f"    -> {resp.content[0].text[:60]}")
 
 
+def run_chat_error():
+    """Scenario: error path — records operation duration with error.type.
+
+    Connects to an unreachable address so the Anthropic SDK raises
+    APIConnectionError, exercising the error.type attribute on the
+    gen_ai.client.operation.duration metric.
+    """
+    import anthropic
+
+    print("  [chat_error] error path (reference implementation)")
+    request_model = "claude-sonnet-4-20250514"
+    # Point to a port that is not listening so the SDK raises APIConnectionError.
+    bad_url = "http://127.0.0.1:1"
+    client = anthropic.Anthropic(base_url=bad_url, api_key="mock-key")
+
+    metric_attributes = {
+        "gen_ai.operation.name": "chat",
+        "gen_ai.provider.name": "anthropic",
+        "gen_ai.request.model": request_model,
+    }
+    start_time = time.perf_counter()
+    try:
+        client.messages.create(
+            model=request_model,
+            max_tokens=100,
+            messages=[{"role": "user", "content": "Say hello."}],
+        )
+    except anthropic.APIError as exc:
+        elapsed = time.perf_counter() - start_time
+        _operation_duration_histogram.record(
+            elapsed,
+            attributes={**metric_attributes, "error.type": type(exc).__qualname__},
+        )
+        print(f"    -> error recorded: {type(exc).__qualname__}")
+
+
 def main():
     print("=== Reference Implementation: Anthropic Reference Implementation ===")
 
@@ -251,6 +287,7 @@ def main():
 
     run_chat()
     run_chat_with_document_input()
+    run_chat_error()
 
     flush_and_shutdown(tp, lp, mp)
 
--- a/reference/scenarios/anthropic/scenario.py
+++ b/reference/scenarios/anthropic/scenario.py
@@ -6,11 +6,13 @@ against a mock Anthropic server, with manual OTel spans.
 
 import json
 import os
+import time
 
 from reference_shared import (
     flush_and_shutdown,
     mock_server_host_port,
     reference_event_logger,
+    reference_meter,
     reference_tracer,
     setup_otel,
 )
@@ -29,6 +31,9 @@ def run_chat():
     request_max_tokens = 100
     messages = [{"role": "user", "content": "Say hello."}]
     client = anthropic.Anthropic(base_url=MOCK_BASE_URL, api_key="mock-key")
+    meter = reference_meter()
+    token_usage_histogram = meter.create_histogram("gen_ai.client.token.usage", unit="{token}")
+    operation_duration_histogram = meter.create_histogram("gen_ai.client.operation.duration", unit="s")
 
     host, port = mock_server_host_port(MOCK_BASE_URL)
     span_attributes = {
@@ -46,11 +51,35 @@ def run_chat():
             "gen_ai.input.messages",
             json.dumps([{"role": m["role"], "parts": [{"type": "text", "content": m["content"]}]} for m in messages]),
         )
+        start_time = time.perf_counter()
         resp = client.messages.create(
             model=request_model,
             max_tokens=request_max_tokens,
             messages=messages,
         )
+        elapsed = time.perf_counter() - start_time
+        metric_attributes = {
+            "gen_ai.operation.name": "chat",
+            "gen_ai.provider.name": "anthropic",
+            "gen_ai.request.model": request_model,
+        }
+        response_model = getattr(resp, "model", None)
+        if response_model:
+            metric_attributes["gen_ai.response.model"] = response_model
+        if host:
+            metric_attributes["server.address"] = host
+        if port is not None:
+            metric_attributes["server.port"] = port
+        if resp.usage:
+            token_usage_histogram.record(
+                resp.usage.input_tokens,
+                attributes={**metric_attributes, "gen_ai.token.type": "input"},
+            )
+            token_usage_histogram.record(
+                resp.usage.output_tokens,
+                attributes={**metric_attributes, "gen_ai.token.type": "output"},
+            )
+        operation_duration_histogram.record(elapsed, attributes=metric_attributes)
         span.set_attribute("gen_ai.response.model", resp.model)
         span.set_attribute("gen_ai.response.id", resp.id)
         span.set_attribute("gen_ai.response.finish_reasons", [resp.stop_reason])
--- a/reference/scenarios/anthropic/scenario.py
+++ b/reference/scenarios/anthropic/scenario.py
@@ -18,8 +18,16 @@ from reference_shared import (
 )
 
 MOCK_BASE_URL = os.environ["MOCK_LLM_URL"]
+TOKEN_BUCKET_BOUNDARIES = [1, 4, 16, 64, 256, 1024, 4096]
 
 _reference_tracer = reference_tracer()
+_reference_meter = reference_meter()
+_token_usage_histogram = _reference_meter.create_histogram(
+    "gen_ai.client.token.usage",
+    unit="{token}",
+    explicit_bucket_boundaries_advisory=TOKEN_BUCKET_BOUNDARIES,
+)
+_operation_duration_histogram = _reference_meter.create_histogram("gen_ai.client.operation.duration", unit="s")
 
 
 def run_chat():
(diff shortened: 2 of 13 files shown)


reference/src/semconv_genai/parse_results.py
```diff
@@ -137,14 +130,21 @@ def _summarize_samples(
...
-                        signals.metric_attrs[metric_name] = set(attr_names)
-                    else:
-                        signals.metric_attrs[metric_name].intersection_update(attr_names)
-                    signals.metric_any_attrs.setdefault(metric_name, set()).update(attr_names)
+                if isinstance(metric_name, str) and metric_name:
+                    data_points = _metric_data_points(metric)
+                    observations = data_points if data_points else [metric]
+                    for observation in observations:
+                        metric_attrs = _attributes_by_name(observation, include_attr)
+                        attr_names = set(metric_attrs)
```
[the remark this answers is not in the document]
AgentGymLeader: Good catch, fixed in 9a6b13c. `_attributes_by_name` now skips attributes whose name is empty or not a string, same as `_attribute_names` already does. So a nameless attr can't sneak in under an empty-string key and skew the metric attr sets.

reference/src/semconv_genai/parse_results.py
```diff
@@ -67,35 +68,27 @@ def _attribute_names(
...
+        if isinstance(raw_data, dict):
+            raw_points = raw_data.get("data_points")
+    if not isinstance(raw_points, list):
+        return []
+    return [dp for dp in raw_points if isinstance(dp, dict)]
+
+
+def _classify_genai_metric(metric_name: str, metric_attrs: dict[str, object]) -> set[str]:
+    del metric_attrs
+    return {metric_name} if metric_name.startswith("gen_ai.") else set()
```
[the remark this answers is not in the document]
AgentGymLeader: Fixed in 9a6b13c. Dropped the local permissive default. `parse_result_dir` now defaults `classify_metric` to the shared `classify.classify_metric`, which is what the pipeline and data-file callers already pass. No new import cycle; `classify` only pulls from `semconv_model`.

reference/src/semconv_genai/parse_results.py
```diff
@@ -380,6 +382,7 @@ def parse_result_dir(
     result_dir: Path,
     library: str,
     classify_span: ClassifySpan,
+    classify_metric: ClassifyMetric = _classify_genai_metric,
```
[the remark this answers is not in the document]
AgentGymLeader: Same as the line-85 thread, addressed in 9a6b13c: the default `classify_metric` now uses the shared `classify.classify_metric`. One path instead of two.

reference/src/semconv_genai/data_files.py
```diff
@@ -171,6 +178,28 @@ def _build_event_type_present_names(result: ScenarioResult) -> dict[str, list[st
...
+def _metric_type_present_attributes(
+    result: ScenarioResult,
+    metric_name: str,
+    level: RequirementLevel,
+) -> set[str]:
+    """Return attrs present for a metric type at the requested requirement level."""
+    all_present = _present_attributes(result)
+    if level is RequirementLevel.REQUIRED:
+        return result.detected.metric_attrs.get(metric_name, all_present)
+    return result.detected.metric_any_attrs.get(metric_name, all_present)
```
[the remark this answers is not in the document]
AgentGymLeader: Good catch — agreed. The fallback to the full present-attribute set overstated coverage for metrics with no observed datapoints. Both branches now fall back to an empty set, so a metric attribute is only counted as present when it was actually observed. (`_present_attributes` stays for span/event types, which are observed differently.)

Fixed in f533cde.

reference/scenarios/anthropic/scenario.py
```diff
@@ -46,11 +51,35 @@ def run_chat():
...
+        response_model = getattr(resp, "model", None)
+        if response_model:
+            metric_attributes["gen_ai.response.model"] = response_model
+        if host:
+            metric_attributes["server.address"] = host
+        if port is not None:
+            metric_attributes["server.port"] = port
+        if resp.usage:
+            token_usage_histogram.record(
+                resp.usage.input_tokens,
```
JWinermaSplunk: Should we record the input token metric with the same total input as calculated below, including cache_creation and cache_read tokens according to existing anthropic semconv?
i.e. `total_input = resp.usage.input_tokens + cache_creation + cache_read`

AgentGymLeader: yeah, that lines up better. The input metric currently only picks up the non-cached `input_tokens`, but I'll record the same total_input (`input_tokens + cache_creation + cache_read`) we calculate below, matching the existing anthropic semconv. I'll update the scenario and push — thanks for catching it 🙏

reference/reports/gen-ai-client-token-usage-metric.md
```diff
@@ -0,0 +1,27 @@
+# Client Token Usage Metric
+
+> **[Semantic Convention](../../docs/gen-ai/gen-ai-metrics.md#metric-gen_aiclienttokenusage)**
+
+## Required
```
MikeGoldsmith, line 5: `error.type` is missing here because the scenario only exercises the happy path, so Weaver never observes it. Same gap exists in the operation-duration report. Could we extend the anthropic scenario to exercise an error path (e.g. wrap messages.create in try/except and record the metric with error.type)? Otherwise the rendered report under-states the metric's actual attribute surface.

reference/src/semconv_genai/parse_results.py
```diff
@@ -137,14 +130,21 @@ def _summarize_samples(
...
+                if isinstance(metric_name, str) and metric_name:
+                    data_points = _metric_data_points(metric)
+                    observations = data_points if data_points else [metric]
+                    for observation in observations:
+                        metric_attrs = _attributes_by_name(observation, include_attr)
+                        attr_names = set(metric_attrs)
+                        for metric_type in classify_metric(metric_name, metric_attrs):
+                            if not metric_type.startswith("gen_ai."):
+                                continue
+                            signals.metrics[metric_type] = signals.metrics.get(metric_type, 0) + 1
```
MikeGoldsmith: The count increment needs to move outside the observations loop. As-is, metrics with multiple data points (e.g., gen_ai.client.token.usage with both input and output records) get double-counted, since this line runs once per data point instead of once per metric.

reference/scenarios/anthropic/scenario.py
```diff
@@ -29,6 +31,9 @@ def run_chat():
     request_max_tokens = 100
     messages = [{"role": "user", "content": "Say hello."}]
     client = anthropic.Anthropic(base_url=MOCK_BASE_URL, api_key="mock-key")
+    meter = reference_meter()
+    token_usage_histogram = meter.create_histogram("gen_ai.client.token.usage", unit="{token}")
+    operation_duration_histogram = meter.create_histogram("gen_ai.client.operation.duration", unit="s")
```
MikeGoldsmith: We should create the histogram before run_chat is called. 

I think we should be setting bucket bounds for this too.

reference/src/semconv_genai/parse_results.py
```diff
@@ -137,14 +130,21 @@ def _summarize_samples(
...
-                    signals.metrics[metric_name] = signals.metrics.get(metric_name, 0) + 1
-                    attr_names = _metric_attribute_names(metric, include_attr)
-                    if metric_name not in signals.metric_attrs:
-                        signals.metric_attrs[metric_name] = set(attr_names)
-                    else:
-                        signals.metric_attrs[metric_name].intersection_update(attr_names)
-                    signals.metric_any_attrs.setdefault(metric_name, set()).update(attr_names)
+                if isinstance(metric_name, str) and metric_name:
+                    data_points = _metric_data_points(metric)
+                    observations = data_points if data_points else [metric]
```
MikeGoldsmith: I think we should warn and return if data_points is not a dict. Just expecting metric to be one data point could cause issues attributes.

reference/src/semconv_genai/data_files.py
```diff
@@ -52,6 +53,12 @@
     "gen_ai.evaluation.result",
 ]
 
+# Display order for metric types in reports.
+METRIC_TYPE_ORDER = [
```
MikeGoldsmith: I think these might already be in `METRIC_SPECS`.

reference/src/semconv_genai/parse_results.py
```diff
@@ -67,35 +73,22 @@ def _attribute_names(
...
-                continue
-            if include_attr is not None and not include_attr(attr):
-                continue
-            names.add(name)
-    return names
+        raw_data = metric.get("data")
+        if isinstance(raw_data, dict):
+            raw_points = raw_data.get("data_points")
+    if not isinstance(raw_points, list):
+        return []
```
MikeGoldsmith: If metric is not valid, we should log a warning.

MikeGoldsmith (changes requested):
Nice work @AgentGymLeader  - I think this is a good start.

I've left a couple of comments & suggestions. I think it's worth adding tests for the new parsing/counting paths so regressions get caught too.

AgentGymLeader:
Thanks for the review. Addressed the comments:

- Double-counting: each metric is now counted once per metric instead of once per data point, so `gen_ai.client.token.usage` with input + output records no longer counts as 2. Added a unit test for this and the other parsing paths.
- `_metric_data_points` now warns and skips when `data_points` isn't a dict or list, instead of assuming a single shape.
- Metrics with an empty or invalid name, or that classify to nothing, now log a warning instead of being dropped silently.
- `METRIC_TYPE_ORDER` now derives from `METRIC_SPECS` so it isn't duplicated.
- Histograms are created at module load (before `run_chat`), and `gen_ai.client.token.usage` sets explicit bucket boundary advice.

On `error.type`: you're right that the happy-path-only scenario hides it. Since we're about to move the anthropic scenario to util-genai (Liudmila's thread, and I'm taking anthropic as the first package), I'd rather add the error path as part of that migration so it isn't instrumented twice. Okay with you, or would you prefer it in this PR?

reference/src/semconv_genai/parse_results.py
AgentGymLeader: Fixed in `ec5276b`. The count increment is now outside the data-points loop, so each metric only increments once regardless of how many data points it has.

reference/src/semconv_genai/parse_results.py
AgentGymLeader: Fixed in `ec5276b`. Added an `isinstance(data_points, list)` check that warns and skips if `data_points` isn't a list.

reference/src/semconv_genai/parse_results.py
AgentGymLeader: Fixed in `ec5276b`. Unknown metric names now log a warning and skip instead of silently disappearing.

reference/src/semconv_genai/data_files.py
AgentGymLeader: You're right, removed the local redefinition in `ec5276b`. Both metrics are only in `METRIC_SPECS` now.

reference/scenarios/anthropic/scenario.py
AgentGymLeader: Fixed in `ec5276b`. Both histograms are now module-level so they're created before any scenario function runs. Added `TOKEN_BUCKET_BOUNDARIES` as advisory explicit bucket bounds on the token histogram.

reference/reports/gen-ai-client-token-usage-metric.md
AgentGymLeader: Fixed in `547614e`. Added `run_chat_error()` to the anthropic scenario — it points to an unreachable address, catches the `APIConnectionError`, and records `gen_ai.client.operation.duration` with `error.type` set to the exception class name. The report now shows `error.type | [anthropic]` for the duration metric.

Left `gen_ai.client.token.usage` without `error.type` — the spec doesn't list it as an attribute there, and a failed call produces no tokens to record. Happy to add it if you think it should be there too.

AgentGymLeader:
Tests added: `ec5276b` adds `reference/tests/test_parse_results.py` covering the double-count fix, the non-list `data_points` guard, and the unrecognized-metric warning path. `547614e` adds a fourth test for the error-path metric attribute detection.

AgentGymLeader:
@MikeGoldsmith — I've addressed all the review comments (fixes in `ec5276b`, `cc7d366`, and `547614e`, each noted inline), and CI is green again. Ready for another look when you get a chance. One open question: I left `error.type` off `gen_ai.client.token.usage` since the spec doesn't list it there, but happy to add it if you'd prefer.

reference/tests/test_parse_results.py
JWinermaSplunk, line 1: I am unsure that adding unit tests is something we want to do at the moment. I think correcting the functionality in `src/semconv_genai` should be enough, @MikeGoldsmith thoughts?

reference/src/semconv_genai/semconv_model.py
```diff
@@ -214,8 +217,15 @@ def _from_yaml(
     ),
 }
 
-# No METRIC_SPECS yet: gen_ai.* metrics are deliberately omitted from the reference
```
JWinermaSplunk, line 217: Just noting for future reference, I am curious about this comment as we can still derive token usage and operation duration from the spans themselves, can we clarify the reason to change this rationale?

reference/scenarios/anthropic/scenario.py
```diff
@@ -6,18 +6,28 @@
...
     flush_and_shutdown,
     mock_server_host_port,
     reference_event_logger,
+    reference_meter,
     reference_tracer,
     setup_otel,
 )
 
 MOCK_BASE_URL = os.environ["MOCK_LLM_URL"]
+TOKEN_BUCKET_BOUNDARIES = [1, 4, 16, 64, 256, 1024, 4096]
```
JWinermaSplunk: Can we follow these bucket boundaries for token usage?
https://github.com/open-telemetry/semantic-conventions-genai/blob/main/docs/gen-ai/gen-ai-metrics.md?plain=1#L55

JWinermaSplunk (commented):
Left a few comments, mainly wondering about the previous comment, as we can still derive token usage and operation duration from the existing span.

AgentGymLeader:
@JWinermaSplunk thanks for the review. Addressed below:

**Bucket boundaries:** switched the token-usage histogram to the standard `ExplicitBucketBoundaries` from `docs/gen-ai/gen-ai-metrics.md` (`[1, 4, 16, 64, 256, 1024, 4096, 16384, 65536, 262144, 1048576, 4194304, 16777216, 67108864]`). Pushed.

**On deriving these from the spans:** you're right that token usage and duration are derivable from the spans. The reason to also emit them as metrics is that the two serve different roles. A metric is a pre-aggregated, low-cardinality series (counts, rates, percentiles) that's cheap to query for dashboards, alerting, and SLOs. Deriving the same aggregates from spans at query time is expensive, and under sampling (head or tail) the span-derived numbers get biased or lossy since you're aggregating over a partial set, whereas metrics are recorded independently of sampling. So it's not about replacing the span; spans stay the source of per-request detail, and the metric just makes the aggregate signal sampling-safe and cheap to query.

**Unit tests:** happy to go either way here. I'll leave them in for now since they pass, but glad to drop them if you and @MikeGoldsmith would rather keep the reference tooling test-free at this stage.

AgentGymLeader:
Any chance you could take another look when you get a sec, @MikeGoldsmith? All 6 points from your review are addressed in ec5276b, cc7d366, and 547614e (double-count fix, `data_points` guard, invalid-metric warning, `METRIC_TYPE_ORDER` dedup, histogram timing, error-path tests), and CI is green.

There's one open question @JWinermaSplunk left to you: whether to keep the unit tests in the reference tooling. I left them in for now, but happy to drop them if you'd rather the tool stay test-free.

AgentGymLeader:
The metric tooling here ended up landing in #324 (the util-genai migration), which is approved now. That makes this PR redundant, so I'll close it once #324 merges.

On the unit tests — agreed, I didn't carry those into #324. The scenario `data.json` plus the run check already covers the metric coverage there, so a separate test file isn't needed.

trask:
@AgentGymLeader does it make sense to move this to draft until #324 and #415 are settled?

AgentGymLeader:
Yeah, makes sense — moving it to draft. I'll close it as superseded once #324 and #415 land.

PR #289, opened 2026-06-14
Remove gen_ai.provider.name from internal agent spans

lmolkova:
Part of #178. 

Provider name does not apply to local agents - they are created by frameworks. Capturing a provider of a model client they use does not seem necessary.

So this PR removes provider name, and it does not introduce any replacement yet.


reference/scenarios/openai-agents/scenario.py
```diff
@@ -58,14 +58,9 @@ def get_weather(ctx: ToolContext[None], location: str) -> str:
     print("  [agent_run] agent with tool calling (reference implementation)")
     agent_span_attributes = {
         "gen_ai.operation.name": "invoke_agent",
-        "gen_ai.provider.name": "openai",
         "gen_ai.request.model": request_model,
         "gen_ai.agent.name": agent.name,
     }
-    if host:
-        agent_span_attributes["server.address"] = host
```
lmolkova, line 66: should not appear on internal span

JWinermaSplunk (commented):
Does the pydantic-ai scenario.py need updating as well?
https://github.com/open-telemetry/semantic-conventions-genai/blob/main/reference/scenarios/pydantic-ai/scenario.py#L229

reference/scenarios/openai-agents/data.json
```diff
@@ -1,19 +1,16 @@
 {
   "spans": {
-    "invoke_agent_client": [
+    "invoke_agent_internal": [
```
trask, line 3: was confused why this switched, sent https://github.com/open-telemetry/semantic-conventions-genai/pull/296 to encode this better going forwards

trask (approved):

reference/scenarios/openai-agents/data.json
lmolkova: thanks!

MikeGoldsmith (approved):

JWinermaSplunk (approved):

Issue #192, opened 2026-05-23
GenAI semconv needs vendor-neutral reasoning parts and continuity tokens


mikeldking:
### Area(s)

area:gen-ai

### What's missing?

Reasoning-capable models now return more than assistant text and tool calls.

They can return ordered reasoning or thinking parts, plus opaque continuity values that the next request must carry forward. Those values are not ordinary metadata. They are part of the provider protocol for continuing a reasoning turn, especially when reasoning happens before a tool call.

Today, the GenAI semantic conventions mostly model assistant output as text plus tool calls. That leaves no vendor-neutral place to represent:

- a reasoning/thinking part in the assistant message
- the original order between reasoning, text, and tool calls
- opaque continuity tokens or signatures attached to those parts
- signatures attached directly to a tool call
- reasoning-specific token accounting and request-side reasoning controls

The practical problem is replay.

A trace can show that a model called a tool, and it can show the tool result, but it may not contain enough information to reconstruct the next provider request. If a tool runner rebuilds the next turn from only the tool call and tool result, it can silently drop the reasoning part that preceded the tool call.

That makes the trace incomplete in a way that matters operationally. You cannot reliably replay the conversation from telemetry, compare provider behavior, debug a failed tool loop, or prove what was sent on turn N+1 from the trace of turn N.

This is not about standardizing or exposing private chain-of-thought text. It is about preserving the provider-visible message structure and the opaque continuity values required to continue the conversation, subject to normal redaction and privacy policy.

### Cross-vendor pattern

The providers use different names and wire formats, but the shape is similar.

**OpenAI Responses API**

- Assistant output can include an `output[]` item with `type: "reasoning"`.
- Reasoning continuity can be carried by `encrypted_content` and the reasoning item identity.
- Reasoning token usage can be reported separately, for example in `output_tokens_details.reasoning_tokens`.
- Request-side controls include fields such as `reasoning.effort` and reasoning summary settings.

**Anthropic Messages API**

- Assistant content can include `thinking` and `redacted_thinking` blocks.
- Continuity is represented through values such as `signature` for visible thinking, or redacted data for redacted thinking.
- Thinking tokens are counted in output tokens, but are not always exposed as a separate usage bucket.
- Request-side controls include thinking type, token budget, and display settings.

**Google Gemini API**

- Response `parts[]` can include thinking parts, for example with `thought: true`.
- Continuity is represented with `thoughtSignature`.
[...]

PR #324, opened 2026-06-19
Migrate anthropic reference scenario to opentelemetry-util-genai

AgentGymLeader:
This PR migrates the Anthropic reference scenario from the hand-written OTel span/event API to [`opentelemetry-util-genai`](https://pypi.org/project/opentelemetry-util-genai/) (1.0b0). This is the first reference scenario to take that dependency; the other 28 still use the hand-written API.

## Scope

Migration only. The metric parse/report changes moved to #415 — my [earlier claim](https://github.com/open-telemetry/semantic-conventions-genai/pull/324#discussion_r3485241060) that they couldn't come out without breaking the anthropic check was wrong: with them reverted `run-scenario anthropic` exits 0 and `data.json` regenerates without a `metrics` block.

## What changed

`scenario.py` uses `handler.inference()` as a context manager instead of raw tracer/span calls:

```python
from opentelemetry.util.genai.handler import get_telemetry_handler
from opentelemetry.util.genai.types import Text, Blob, InputMessage, OutputMessage

handler = get_telemetry_handler(tracer_provider=tp, meter_provider=mp, logger_provider=lp)

with handler.inference(provider="anthropic", request_model=model, ...) as inv:
    inv.max_tokens = request_max_tokens       # -> gen_ai.request.max_tokens
    inv.input_messages = [InputMessage(...)]  # -> gen_ai.input.messages
    # ... SDK call ...
    inv.response_model_name = resp.model      # -> gen_ai.response.model
    inv.output_messages = [OutputMessage(...)]
```

The handler owns span creation, attribute serialization, event emission, and metric recording — the scenario only populates the invocation fields.

## Scenarios covered

| Scenario | What it exercises |
|---|---|
| `run_chat` | Basic chat + `gen_ai.request.reasoning.level` via `inv.attributes` |
| `run_compaction` | `client.beta.messages.create` with `context_management`, `gen_ai.conversation.compacted` |
| `run_chat_with_document_input` | `Blob(modality="document", content=pdf_bytes)` |

`run_chat_error` is a follow-up, not part of this PR.

## Attribute parity

The new `data.json` includes all attributes from the hand-written version plus the two added since our branch point:

- `gen_ai.request.reasoning.level` (added in #258) — set via `inv.attributes["gen_ai.request.reasoning.level"]` since `InferenceInvocation` has no first-class property for it yet
- `gen_ai.conversation.compacted` (added in #162) — set via `inv.attributes["gen_ai.conversation.compacted"]`

Compaction blocks have no util-genai part type, so they are emitted as plain dict parts inside `InputMessage`/`OutputMessage`; `dataclasses.asdict` recurses into them, so the span and the event carry the same shape. They stay type-only: Anthropic exposes an encrypted, opaque compaction state here, never the summary that `CompactionPart.content` documents.

Zero attributes removed compared to the hand-written baseline.

Two changes an attribute-name table can't show. The messages on `gen_ai.client.inference.operation.details` were JSON strings before; util-genai serializes them structurally for events and keeps the JSON string on spans (`get_content_attributes(for_span=...)`). In the structured form a document blob's `content` is raw bytes, where the span still carries base64. `run_chat_with_document_input` also emits that event now — before this PR it produced a span only.


reference/scenarios/anthropic/scenario.py
```diff
@@ -200,185 +127,204 @@ def run_compaction_reference():
...
+    )
+    inv.max_tokens = 1
+    inv.input_messages = [InputMessage(role="user", parts=[Text(content="Trigger error")])]
+
+    try:
+        client.messages.create(
+            model="",
+            max_tokens=1,
+            messages=[{"role": "user", "content": "Trigger error"}],
         )
```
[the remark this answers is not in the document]
AgentGymLeader: Thanks, good catch. Fixed: the error scenario now records an empty request model so the telemetry matches the actual failing call rather than the intended model name.

reference/src/semconv_genai/semconv_model.py
```diff
@@ -214,8 +217,15 @@ def _from_yaml(
...
+        _groups,
+        "metric.gen_ai.client.token.usage",
+        label="Client Token Usage",
+    ),
+    "gen_ai.client.operation.duration": _from_yaml(
+        _groups,
+        "metric.gen_ai.client.operation.duration",
+        label="Client Operation Duration",
+    ),
+}
```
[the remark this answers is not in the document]
AgentGymLeader: This is intentional for now: METRIC_SPECS mirrors EVENT_SPECS and is scoped to the metrics the reference scenarios actually emit. Broadening it toward the full registry, so other valid metrics are recognized rather than warned, is a fair follow-up; for this PR I've kept it aligned with the in-scope scenarios. Thanks for flagging it.

reference/tests/test_parse_results.py
lmolkova, line 1: do we need tests for reference scenarios?

reference/scenarios/anthropic/scenario.py
```diff
@@ -200,185 +127,204 @@ def run_compaction_reference():
...
-            "gen_ai.input.messages",
-            input_messages_json,
-        )
-        resp = client.beta.messages.create(
-            model=request_model,
-            max_tokens=request_max_tokens,
-            messages=messages,
-            context_management={
-                "edits": [
+    inv = handler.start_inference(
```
lmolkova:
```suggestion
   with handler.inference(...) as inv:
```

reference/scenarios/anthropic/scenario.py
```diff
@@ -200,185 +127,204 @@ def run_compaction_reference():
...
+        provider="anthropic",
+        request_model=request_model,
+        server_address=host,
+        server_port=port,
+    )
+    inv.max_tokens = request_max_tokens  # -> gen_ai.request.max_tokens
+
+    # Compaction blocks have no util-genai type yet; set gen_ai.input.messages manually.
+    inv.attributes["gen_ai.input.messages"] = json.dumps(  # -> gen_ai.input.messages
```
lmolkova:
```suggestion
    inv.input_messages = typed_object
```

Please use actual instrumentation as an example of how to use utils https://github.com/open-telemetry/opentelemetry-python-genai/blob/main/instrumentation/opentelemetry-instrumentation-genai-anthropic/src/opentelemetry/instrumentation/genai/anthropic/patch.py#L81

reference/scenarios/anthropic/pyproject.toml
```diff
@@ -5,6 +5,7 @@ requires-python = ">=3.12"
 dependencies = [
     "anthropic==0.107.1",
     "genai-reference-shared",
+    "opentelemetry-util-genai==0.4b0",
```
lmolkova: let's take unreleased version, there should be an easy way https://github.com/open-telemetry/opentelemetry-python-genai/tree/main/util/opentelemetry-util-genai

reference/scenarios/anthropic/scenario.py
```diff
@@ -200,185 +127,204 @@ def run_compaction_reference():
...
+        if block_type == "text" and getattr(block, "text", None):
+            output_parts.append({"type": "text", "content": block.text})
+        elif block_type == "compaction":
+            part = {"type": "compaction"}
+            compaction_content = getattr(block, "content", None)
+            if compaction_content:
+                part["content"] = compaction_content
+            output_parts.append(part)
+    if output_parts:
+        inv.attributes["gen_ai.output.messages"] = json.dumps(  # -> gen_ai.output.messages
```
lmolkova:
```suggestion
        inv.output_messages = ...
```

reference/scenarios/anthropic/scenario.py
```diff
@@ -200,185 +127,204 @@ def run_compaction_reference():
...
+    handler = get_telemetry_handler(
+        tracer_provider=tp,
+        meter_provider=mp,
+        logger_provider=lp,
+    )
+
+    run_chat(handler)
+    run_compaction(handler)
+    run_chat_with_document_input(handler)
+    run_chat_error(handler)
```
lmolkova: please keep the PR focused on change, avoid adding unrelated changed (e.g. new scenarios)

reference/scenarios/anthropic/scenario.py
```diff
@@ -200,185 +127,204 @@ def run_compaction_reference():
...
+    try:
+        client.messages.create(
+            model="",
+            max_tokens=1,
+            messages=[{"role": "user", "content": "Trigger error"}],
         )
-        span.set_attribute("gen_ai.output.messages", output_messages)
-        print(f"    -> {resp.content[0].text[:60]}")
+    except Exception as exc:  # noqa: BLE001
+        inv.fail(Error(type=type(exc), message=str(exc)))
```
lmolkova:
```suggestion
        inv.fail(exc)
```

reference/tests/test_parse_results.py
AgentGymLeader: reference doesn't carry unit tests elsewhere, so I dropped this. The metric-counting logic it touched is exercised by `run-scenario` through `parse_results.py`.

reference/scenarios/anthropic/scenario.py
AgentGymLeader: Applied in 7f43e37.

reference/scenarios/anthropic/scenario.py
AgentGymLeader: Applied. Moved to the context-manager form and matched the anthropic `patch.py` idiom.

reference/scenarios/anthropic/pyproject.toml
AgentGymLeader: Switched to the unreleased util-genai, pinned to a commit on python-genai (subdirectory `util/opentelemetry-util-genai`) for a reproducible build. Happy to track a branch/tag instead if you'd prefer.

reference/scenarios/anthropic/scenario.py
AgentGymLeader: Applied in 7f43e37.

reference/scenarios/anthropic/scenario.py
AgentGymLeader: Agreed. Moved the new `run_chat_error` scenario to a follow-up so this one stays on the migration (chat/compaction/document already existed and are just migrated). Kept the metric parse/report changes because the migrated scenarios now emit token-usage/duration metrics that `run-scenario` validates against `data.json`, so dropping them breaks the anthropic check. Happy to split those too if you'd prefer.

reference/scenarios/anthropic/scenario.py
AgentGymLeader: This lived in `run_chat_error`, which moved to the follow-up PR, so the `inv.fail(exc)` suggestion is applied there.

AgentGymLeader:
Thanks for the review. Pushed 7f43e37 addressing the points, with replies inline. Moved the new error scenario to a focused follow-up.

reference/scenarios/anthropic/scenario.py
```diff
@@ -235,150 +167,128 @@ def run_compaction_reference():
...
+            if block_type == "text" and getattr(block, "text", None):
+                output_parts.append({"type": "text", "content": block.text})
+            elif block_type == "compaction":
+                part = {"type": "compaction"}
+                compaction_content = getattr(block, "content", None)
+                if compaction_content:
+                    part["content"] = compaction_content
+                output_parts.append(part)
+        if output_parts:
+            inv.attributes["gen_ai.output.messages"] = json.dumps(  # -> gen_ai.output.messages
```
lmolkova: no need to json dump and please set typed object

reference/scenarios/anthropic/scenario.py
```diff
@@ -200,25 +124,33 @@ def run_compaction_reference():
...
-            input_messages_json,
+    with handler.inference(
+        provider="anthropic",
+        request_model=request_model,
+        server_address=host,
+        server_port=port,
+    ) as inv:
+        inv.max_tokens = request_max_tokens  # -> gen_ai.request.max_tokens
+
+        # Compaction blocks have no util-genai type yet; set gen_ai.input.messages manually.
```
lmolkova: there is a [GenericPart](https://github.com/open-telemetry/opentelemetry-python-genai/blob/1c19a61de020966f24a63b7163c4f993a3b05060/util/opentelemetry-util-genai/src/opentelemetry/util/genai/types.py#L31) that can capture anything, but I think one in the utils is not right - it has "generic" type, but semconv version of it does not require it https://github.com/open-telemetry/semantic-conventions-genai/blob/c321d7eb4443ae1d1d88c2e24eda849f62049008/docs/gen-ai/non-normative/models.py#L254

I think it's ok to keep it in this PR for now, but would be nice to 1) add compaction part to util-genai 2) fix GenericPart ther

lmolkova (approved):

reference/scenarios/anthropic/scenario.py
AgentGymLeader: Same root problem as the input side. The response in this scenario comes back as a compaction block (stop_reason "compaction"), and `GenericPart` would serialize it with type "generic" rather than "compaction", so the typed path can't produce the right shape yet. Kept it manual per your other comment. Happy to switch this over once util-genai can represent compaction.

reference/scenarios/anthropic/scenario.py
AgentGymLeader: Agreed on both. A first-class compaction part, plus letting GenericPart carry the actual type string, would let this scenario go fully typed. If it'd help I can take a stab at those two on the util-genai side. No worries if they're already on someone's list.

reference/scenarios/anthropic/scenario.py
```diff
@@ -200,25 +124,33 @@ def run_compaction_reference():
...
+        inv.attributes["gen_ai.input.messages"] = json.dumps(  # -> gen_ai.input.messages
+            [
+                {
+                    "role": "assistant",
+                    "parts": [
+                        {
+                            "type": "compaction",
+                            "content": "opaque encrypted compaction state from a prior turn",
```
JWinermaSplunk: I don't believe we should be defining content here as we define `CompactionPart.content` as `The unencrypted compacted conversation summary, when available.` in https://github.com/open-telemetry/semantic-conventions-genai/blob/7f43e37cb21034e9a341f587bbcb9c55a2713e86/docs/gen-ai/non-normative/models.py#L175-L192. Adding `CompactionPart.encrypted_content` or something similar is definitely something to look into in another issue.

JWinermaSplunk (commented):
Can we go ahead and rebase and regenerate the `uv.lock`? Believe `opentelemetry-instrumentation==0.63b1` requires `opentelemetry-semantic-conventions==0.63b1` but we're currently on `opentelemetry-semantic-conventions==0.64b0`. Think we should consider upgrading `opentelemetry-instrumentation` or allow it to resolve a compatible version.

AgentGymLeader:
`CompactionPart.content` is the unencrypted summary, and Anthropic only hands us the encrypted opaque state here, so rather than make one up I left the compaction parts type-only. An encrypted-content field is probably its own issue.

Also rebased on main and dropped the git pin for the released `opentelemetry-util-genai==1.0b0` — that gets `opentelemetry-instrumentation` and `opentelemetry-semantic-conventions` onto a compatible pair. `uv.lock` + generated files refreshed.

JWinermaSplunk (approved):
LGTM, can we go ahead and update the package version to `opentelemetry-util-genai==1.0b0` in the description and file a follow up issue(s) for Compaction in either this repo and/or the python genai repo based on any actionable findings here? Encryption via `encrypted_content` and support within the handler for `CompactionPart` are the two that came to mind, if not already present. Thanks!

AgentGymLeader:
Thanks @JWinermaSplunk — done. The description now points at `opentelemetry-util-genai==1.0b0`.

On compaction: the opaque continuity value (`encrypted_content`) is the same shape as #192's vendor-neutral continuity tokens, so I noted the compaction case there rather than opening a duplicate. The handler-side `CompactionPart` support is a util-genai concern, so I filed a focused follow-up: https://github.com/open-telemetry/opentelemetry-python-genai/issues/270

reference/scenarios/anthropic/scenario.py
```diff
@@ -1,191 +1,118 @@
...
-def response_has_compaction_block(response):
-    """Return whether Anthropic reported compaction in response content or usage."""
-    if getattr(response, "stop_reason", None) == "compaction":
-        return True
-    for block in getattr(response, "content", []) or []:
-        if getattr(block, "type", None) == "compaction":
-            return True
-    usage = getattr(response, "usage", None)
-    return any(getattr(iteration, "type", None) == "compaction" for iteration in getattr(usage, "iterations", []) or [])
+os.environ["OTEL_SEMCONV_STABILITY_OPT_IN"] = "gen_ai_latest_experimental"
```
lmolkova: should no longer be necessary

lmolkova (approved):

reference/README.md
```diff
@@ -46,6 +46,8 @@ Run `uv run update-reports` to regenerate.
 
 | Metric | Libraries |
 | --- | --- |
+| [Client Token Usage](reports/gen-ai-client-token-usage-metric.md) | agent-framework, anthropic, google-adk |
+| [Client Operation Duration](reports/gen-ai-client-operation-duration-metric.md) | agent-framework, anthropic, google-adk |
```
trask: Anchoring PR-title/description feedback here, since this PR adds no `changelog.d/` fragment and these two rows are the visible surface of the part the description never mentions.

**The description no longer matches the code.** It walks through `handler.start_inference(...)` / `inv.stop()`, but head uses `handler.inference()` as a context manager. It sets out `OTEL_SEMCONV_STABILITY_OPT_IN` and devotes a whole "Private `_initialize()` call" section to justifying a private API call — neither exists at head, which sets only `OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT` and `OTEL_INSTRUMENTATION_GENAI_EMIT_EVENT`. That section explicitly invites reviewer input ("Happy to explore alternatives"), and the PR is already approved, so there's a real chance it was approved against a concern that no longer applies. Worth rewriting before merge so the record reflects what's actually being merged.

**"in line with the handler-based pattern used by the other reference libraries"** is not accurate — `git grep -l opentelemetry-util-genai` at the merge base returns nothing. This is the first scenario to take that dependency, which makes it a precedent for the other 28 rather than a catch-up. That's worth stating plainly, since it's the more interesting claim.

**"All four scenarios are migrated"** — the table lists three, and `scenario.py` defines three.

**Neither the title nor the description mentions the metric-coverage expansion**, which is these two rows. Adding `gen_ai.client.token.usage` and `gen_ai.client.operation.duration` to `METRIC_SPECS` reverses a documented maintainer decision and is the largest change in the PR by blast radius: two new report files, this README table, `data.json` churn in agent-framework and google-adk, and a rewrite of metric parsing/classification. Under a title scoped to "migrate anthropic reference scenario", that is easy to approve without noticing. See my comment on `semconv_model.py` — my preference is to split it out; if it stays, the title and description need to cover it.

reference/scenarios/anthropic/scenario.py
```diff
@@ -200,25 +122,28 @@ def run_compaction_reference():
...
+        # Compaction blocks have no util-genai type yet; set gen_ai.input.messages manually.
+        inv.attributes["gen_ai.input.messages"] = json.dumps(  # -> gen_ai.input.messages
+            [
+                {
+                    "role": "assistant",
+                    "parts": [{"type": "compaction"}],
+                },
+                {"role": "user", "parts": [{"type": "text", "content": "Continue this long conversation."}]},
+            ]
         )
```
trask: Two problems here.

**1. The emitted value is hardcoded rather than derived.** `"Continue this long conversation."` now appears twice in this file — once in `messages` and once in this literal — so the telemetry can silently drift from the request that was actually sent. The pre-migration code derived this from `messages` via `input_messages(messages)`. `.github/instructions/reference-scenarios.instructions.md`: *"For attributes whose value is not truly static for the scenario, do not hardcode the emitted value... If the same non-static value is needed in both the SDK call and span attributes, bind it once locally and reuse it."* Note `_has_compaction_block_in_input(messages)` on line 163 still reads the real messages, so half the emission is derived and half is a literal.

**2. Setting this through `inv.attributes` corrupts the event, not just the span.** In util-genai 1.0b0, `InferenceInvocation._maybe_create_event()` builds event attributes with `get_content_attributes(for_span=False)` — a structured `list[dict]` — and then does `attributes.update(self.attributes)` **last**. So `inv.attributes` overrides the event too, and `gen_ai.client.inference.operation.details` ends up with `gen_ai.input.messages` as a JSON **string** here while `run_chat` and `run_chat_with_document_input` emit it as a structured list. Same library, same event, two different value shapes. `data.json` only records attribute *names*, so CI cannot catch this.

Both are fixable without any new type: `dataclasses.asdict` recurses into plain dicts, so a dict part inside `InputMessage` serializes to exactly the semconv compaction shape on both signals (verified: `asdict(InputMessage(role="assistant", parts=[{"type": "compaction"}]))` -> `{"role": "assistant", "parts": [{"type": "compaction"}]}`).

```suggestion
        # Compaction blocks have no util-genai part type, but `dataclasses.asdict`
        # recurses into plain dicts, so a dict part reaches the span and the event
        # in the same semconv shape.
        compaction_message, user_message = messages
        inv.input_messages = [  # -> gen_ai.input.messages
            InputMessage(role=compaction_message["role"], parts=[{"type": "compaction"}]),
            InputMessage(role=user_message["role"], parts=[Text(content=user_message["content"])]),
        ]
```

Heads-up: with this and the `gen_ai.output.messages` suggestion (line 185) applied, `import json` on line 9 becomes unused and ruff F401 will fail — drop it in the same change.

reference/scenarios/anthropic/scenario.py
```diff
@@ -235,150 +160,127 @@ def run_compaction_reference():
...
+                output_parts.append({"type": "text", "content": block.text})
+            elif block_type == "compaction":
+                # Anthropic only ever returns `encrypted_content` here, never the
+                # unencrypted summary CompactionPart.content documents, so this part
+                # stays type-only.
+                output_parts.append({"type": "compaction"})
+        if output_parts:
+            inv.attributes["gen_ai.output.messages"] = json.dumps(  # -> gen_ai.output.messages
+                [{"role": "assistant", "parts": output_parts, "finish_reason": resp.stop_reason}]
+            )
```
trask: Same event-shape issue as `gen_ai.input.messages` (line 137): routing this through `inv.attributes` overrides the structured event payload that `get_content_attributes(for_span=False)` produced, so this event carries a JSON string while the other two scenarios carry a list.

`output_parts` is already a list of plain dicts, and `asdict` recurses into those, so handing it to `inv.output_messages` keeps the compaction part type-only while letting the handler serialize the span and the event consistently. `inv.finish_reasons` is already set on line 168, so `_get_finish_reasons()` still resolves to `resp.stop_reason`.

```suggestion
        if output_parts:
            inv.output_messages = [  # -> gen_ai.output.messages
                OutputMessage(role="assistant", parts=output_parts, finish_reason=resp.stop_reason),
            ]
```

reference/scenarios/anthropic/scenario.py
```diff
@@ -1,191 +1,113 @@
...
+    """Scenario: server-side compaction via context_management beta.
+
+    gen_ai.conversation.compacted is set via inv.attributes since
+    InferenceInvocation does not have a first-class property for it yet.
+    Compaction blocks in input/output messages are serialised manually
+    via inv.attributes["gen_ai.input.messages"] because opentelemetry-util-genai
+    does not have a CompactionPart type. The compaction parts stay type-only
+    (no `content`): Anthropic only exposes an encrypted, opaque compaction
+    state here, never the unencrypted summary CompactionPart.content
+    documents, so it would be dishonest to fabricate one.
```
trask: If the `gen_ai.input.messages` (line 137) and `gen_ai.output.messages` (line 185) suggestions are taken, this paragraph needs to follow. Also worth noting: util-genai 1.0b0 *does* ship `GenericPart` as its opt-in for provider-specific parts — it just isn't usable here, because it serializes to `{"type": "generic", "value": ...}` rather than the semconv compaction shape. Saying so preempts the obvious reviewer question.

```suggestion
    Compaction blocks have no util-genai part type (`GenericPart` would serialize
    as `{"type": "generic", ...}`, not the semconv shape), so they are emitted as
    plain dict parts inside `InputMessage`/`OutputMessage`; `dataclasses.asdict`
    recurses into them, so the span and the event carry the same shape. The
    compaction parts stay type-only (no `content`): Anthropic only exposes an
    encrypted, opaque compaction state here, never the unencrypted summary
    CompactionPart.content documents, so it would be dishonest to fabricate one.
```

reference/src/semconv_genai/parse_results.py
```diff
@@ -67,35 +76,43 @@ def _attribute_names(
...
+                data_point,
+            )
+        return data_points
+
+    _LOGGER.warning(
+        "Skipping metric %r with invalid data_points: expected dict or list, got %s",
+        metric.get("name", ""),
+        type(raw_points).__name__,
+    )
+    return None
```
trask: I checked this against the pinned Weaver (`v0.24.2`) instead of assuming, and **all three non-list branches here are unreachable**. `SampleMetric` is declared as:

```rust
pub struct SampleMetric {
    pub name: String,
    pub instrument: SampleInstrument,
    pub unit: String,
    pub data_points: Option<DataPoints>,   // <- no skip_serializing_if
    pub live_check_result: Option<LiveCheckResult>,
}

#[serde(untagged)]
pub enum DataPoints {
    Number(Vec<SampleNumberDataPoint>),
    Histogram(Vec<SampleHistogramDataPoint>),
    ExponentialHistogram(Vec<SampleExponentialHistogramDataPoint>),
}
```

- `data_points` has no `#[serde(skip_serializing_if)]`, so the key is **always present** (`null` when empty). `raw_points is marker` never happens, so `return [metric]` is dead.
- `DataPoints` is an untagged enum over `Vec`s, so it always serializes as an **array**. The `isinstance(raw_points, dict)` branch is dead.
- `SampleMetric` has **no `data` field at all**, so the `metric["data"]["data_points"]` lookup is dead, and the docstring's "either known Weaver metric shapes" doesn't describe anything Weaver emits.

Confirmed empirically as well: I ran `uv run run-scenario google-adk` against Weaver v0.24.2 and every one of the 16 metric samples in `live_check.json` had keys exactly `['data_points', 'instrument', 'live_check_result', 'name', 'unit']`, with `data_points` a list. No `data` key, no dict, no absent key.

**The shape that is real is the one this function mishandles.** `otlp_data_to_data_points` in `src/registry/otlp/conversion.rs` ends in `_ => None`, so a Summary metric (or one with no data set) serializes as `"data_points": null`. That falls past all three checks into the final warning and is reported as `invalid data_points ... got NoneType` — a valid Weaver serialization logged as invalid. `_summarize_samples` already treats an empty result as "nothing to record" (`if observations is None or not observations: continue`), so `None` should simply map to `[]`.

```suggestion
def _metric_data_points(metric: dict[str, object]) -> list[dict[str, object]] | None:
    """Return the metric's data points, or None if `data_points` is unusable."""
    raw_points = metric.get("data_points")
    if raw_points is None:
        return []
    if not isinstance(raw_points, list):
        _LOGGER.warning(
            "Skipping metric %r with invalid data_points: expected list, got %s",
            metric.get("name", ""),
            type(raw_points).__name__,
        )
        return None

    data_points: list[dict[str, object]] = []
    for data_point in raw_points:
        if isinstance(data_point, dict):
            data_points.append(data_point)
            continue
        _LOGGER.warning(
            "Skipping non-dict data point for metric %r: %r",
            metric.get("name", ""),
            data_point,
        )
    return data_points
```

I applied this on top of this branch and checked it: `ruff check src`, `ruff format --check src`, and the existing `tests/` suite all pass.

reference/src/semconv_genai/parse_results.py
```diff
@@ -137,14 +154,41 @@ def _summarize_samples(
...
+                    continue
+
+                metric_attrs_by_type: dict[str, set[str]] = {}
+                metric_any_attrs_by_type: dict[str, set[str]] = {}
+                for observation in observations:
+                    metric_attrs = _attributes_by_name(observation, include_attr)
+                    attr_names = set(metric_attrs)
+                    for metric_type in classify_metric(metric_name, metric_attrs):
+                        if not metric_type.startswith("gen_ai."):
+                            continue
```
trask: This guard is unreachable. `classify_metric` returns `{metric_name} if metric_name in METRIC_SPECS else set()`, and every key in `METRIC_SPECS` is `gen_ai.*`-prefixed, so `metric_type.startswith("gen_ai.")` is always true. The parallel span path has no equivalent guard either. If the invariant matters, it belongs in `classify_metric` rather than being re-checked at each call site.

```suggestion
                    for metric_type in classify_metric(metric_name, metric_attrs):
```

reference/src/semconv_genai/parse_results.py
```diff
@@ -137,14 +154,41 @@ def _summarize_samples(
...
+                        if metric_type not in metric_attrs_by_type:
+                            metric_attrs_by_type[metric_type] = set(attr_names)
+                        else:
+                            metric_attrs_by_type[metric_type].intersection_update(attr_names)
+                        metric_any_attrs_by_type.setdefault(metric_type, set()).update(attr_names)
+
+                if not metric_attrs_by_type:
+                    _LOGGER.warning("Skipping unrecognized metric: %s", metric_name)
+                    continue
```
trask: This turns "metric we don't track" into a warning on **every sample**. The previous code filtered with `if metric_name.startswith("gen_ai.")` and silently ignored everything else; now any non-GenAI metric that reaches the results directory — and any `gen_ai.*` metric not yet in `METRIC_SPECS` — logs during `make generate-all`. Untracked non-GenAI metrics are expected, not anomalies.

```suggestion
                if not metric_attrs_by_type:
                    if metric_name.startswith("gen_ai."):
                        _LOGGER.warning("Skipping untracked GenAI metric: %s", metric_name)
                    continue
```

reference/src/semconv_genai/parse_results.py
```diff
@@ -380,6 +426,7 @@ def parse_result_dir(
     result_dir: Path,
     library: str,
     classify_span: ClassifySpan,
+    classify_metric: ClassifyMetric = _default_classify_metric,
```
trask: This default is dead: both call sites — `data_files.write_generated_scenario_data` and `pipeline.run_one_library` — pass `classify_metric` explicitly in this PR, and those are the only two callers. It also makes the signature asymmetric (`classify_span` is required) and forces `parse_results` to import from `classify` purely to supply a default it never uses.

```suggestion
    classify_metric: ClassifyMetric,
```

With this applied, `from .classify import classify_metric as _default_classify_metric` on line 11 becomes unused and must be removed too (ruff F401).

reference/src/semconv_genai/data_files.py
```diff
@@ -184,10 +181,9 @@ def _metric_type_present_attributes(
     level: RequirementLevel,
 ) -> set[str]:
     """Return attrs present for a metric type at the requested requirement level."""
-    all_present = _present_attributes(result)
     if level is RequirementLevel.REQUIRED:
-        return result.detected.metric_attrs.get(metric_name, all_present)
-    return result.detected.metric_any_attrs.get(metric_name, all_present)
+        return result.detected.metric_attrs.get(metric_name, set())
+    return result.detected.metric_any_attrs.get(metric_name, set())
```
trask: This is a semantics change bundled into the migration, not a mechanical refactor: the fallback went from "every attribute present anywhere in the scenario" to "nothing". For a metric that is counted only via `statistics` (no sample data points in `detected.metric_attrs`), the report now shows no supporting attributes instead of over-reporting all of them.

The new behavior is almost certainly the correct one, but it silently changes what the reports claim about libraries whose metrics are not sample-backed. Worth a line in the PR description, and ideally a test — `reference/tests/test_metrics.py` only covers the case where `metric_attrs` is populated.

reference/tests/test_metrics.py
```diff
@@ -37,8 +37,12 @@ def _result_with_tool_calls_metric() -> ScenarioResult:
...
 def test_metric_specs_expose_recommended_agent_name():
     assert METRIC_SPECS, "expected at least one tracked metric"
-    for name, spec in METRIC_SPECS.items():
-        assert "gen_ai.agent.name" in spec.recommended, name
+    # Only the per-invocation agent metrics are agent-scoped; gen_ai.client.*
+    # metrics (token usage, operation duration) are not dimensioned by
+    # gen_ai.agent.name, so this checks the invoke_agent metrics specifically
+    # rather than every entry in METRIC_SPECS.
+    for name in (_INFERENCE_CALLS, _TOOL_CALLS):
+        assert "gen_ai.agent.name" in METRIC_SPECS[name].recommended, name
```
trask: Loosening this assertion is reasonable, but this file is the only test in the repo that touches the metrics path, and it covers none of the parsing added in this PR. `_metric_data_points` and `classify_metric` are new, and `_summarize_samples` now intersects attributes across multiple data points — all untested. The `data.json` round-trip test can't catch shape regressions because it starts from an already-parsed `DetectedSignals`.

These three cover the branches that matter:

```python
import logging

from semconv_genai.classify import classify_metric
from semconv_genai.parse_results import _summarize_samples

_TOKEN_USAGE = "gen_ai.client.token.usage"


def _samples(metric: dict) -> list[dict]:
    return [{"samples": [{"metric": metric}]}]


def _no_spans(*_args) -> set[str]:
    return set()


def test_metric_attrs_intersect_across_data_points():
    metric = {
        "name": _TOKEN_USAGE,
        "data_points": [
            {"attributes": [{"name": "gen_ai.operation.name"}, {"name": "server.port"}]},
            {"attributes": [{"name": "gen_ai.operation.name"}]},
        ],
    }
    _, signals = _summarize_samples(_samples(metric), _no_spans, classify_metric)
    assert signals.metric_attrs[_TOKEN_USAGE] == {"gen_ai.operation.name"}
    assert signals.metric_any_attrs[_TOKEN_USAGE] == {"gen_ai.operation.name", "server.port"}


def test_null_data_points_is_skipped_without_warning(caplog):
    metric = {"name": _TOKEN_USAGE, "data_points": None}
    with caplog.at_level(logging.WARNING):
        _, signals = _summarize_samples(_samples(metric), _no_spans, classify_metric)
    assert signals.metrics == {}
    assert caplog.records == []


def test_untracked_metric_is_ignored():
    metric = {"name": "http.client.request.duration", "data_points": [{"attributes": []}]}
    _, signals = _summarize_samples(_samples(metric), _no_spans, classify_metric)
    assert signals.metrics == {}
```

I ran these against this branch. The first and third pass as-is. `test_null_data_points_is_skipped_without_warning` reproduces the real Weaver `"data_points": null` shape and **fails today** with:

```
WARNING semconv_genai.parse_results:parse_results.py:104 Skipping metric
'gen_ai.client.token.usage' with invalid data_points: expected dict or list, got NoneType
```

It passes once `_metric_data_points` is fixed — see my comment on `parse_results.py`.

trask (commented):

reference/src/semconv_genai/parse_results.py
AgentGymLeader: @trask Checked this against v0.24.2's `SampleMetric` too, and I get a slightly different result on one of the three.

`DataPoints` is `#[serde(untagged)]` over three `Vec<...>` variants, so you're right that the dict branch and the final warning are unreachable for a present value.

But `data_points: Option<DataPoints>` carries no `skip_serializing_if`, so `None` serializes as `"data_points": null` rather than being omitted. That makes the `marker` branch dead, while a null lands on the final warning path instead. So two of the three go and the last one still has a real input shape behind it.

A live run of the anthropic scenario gives 4 `data_points`, all lists, all Number instruments, so it doesn't exercise the null case either way.

AgentGymLeader:
@trask thanks — all four checked out, and all four are addressed.

1. **Description vs code.** Rewritten. The `_initialize()` section is gone; that code no longer exists at head, so the section was still inviting review input on something already removed.
2. **"in line with the other reference libraries."** Wrong, as you measured — `git grep -l opentelemetry-util-genai` returns nothing on main. The description now says this is the first of 29, with the other 28 still on the hand-written API.
3. **"All four scenarios."** Three. Fixed, and `run_chat_error` is called out as a follow-up.
4. **Metric coverage expansion.** Split out to #415, which covers the `METRIC_SPECS` change and the maintainer decision it reverses.

This PR is now migration only. The six threads on `parse_results.py`, `data_files.py` and `test_metrics.py` moved with it — five are applied in #415, and the open `_metric_data_points` question is re-posted there so a reply lands next to the code.

reference/README.md
AgentGymLeader: All four addressed — details in https://github.com/open-telemetry/semantic-conventions-genai/pull/324#issuecomment-5141184946.

reference/scenarios/anthropic/scenario.py
AgentGymLeader: Applied, both points. Checked the override at head: `_maybe_create_event()` does `attributes.update(self._get_message_attributes(for_span=False))` and then `attributes.update(self.attributes)` last, so `inv.attributes` did win over the structured payload. Parts are derived from `messages` now, and `import json` is gone.

reference/scenarios/anthropic/scenario.py
AgentGymLeader: Applied as suggested.

reference/scenarios/anthropic/scenario.py
AgentGymLeader: Applied verbatim, including the `GenericPart` note.

reference/scenarios/anthropic/scenario.py
```diff
@@ -1,191 +1,113 @@
...
-        {
-            "role": "assistant",
-            "parts": parts,
-            "finish_reason": getattr(response, "stop_reason", None) or "stop",
-        }
-    ]
+import anthropic
+from opentelemetry.util.genai.handler import get_telemetry_handler
+from opentelemetry.util.genai.types import Blob, InputMessage, OutputMessage, Text
+from reference_shared import flush_and_shutdown, mock_server_host_port, setup_otel
```
trask, line 9: These don't need to be set before the imports. `get_content_capturing_mode()` and `should_emit_event()` both read `os.environ` on every call, and the only value cached at construction time is `TelemetryHandler._capture_content` — built in `TelemetryHandler.__init__`, which this scenario only reaches via `get_telemetry_handler()` inside `main()`. Setting them at the top of `main()` is enough, and it lets the imports be a normal single block.

Keep both assignments though. I checked and `OTEL_INSTRUMENTATION_GENAI_EMIT_EVENT=true` is not redundant: if the ambient environment has it set to `false`, that explicitly wins over `SPAN_AND_EVENT` inside `should_emit_event()` and no events are emitted at all.

```suggestion
"""Reference implementation for Anthropic using opentelemetry-util-genai."""

import base64
import os

import anthropic
from opentelemetry.util.genai.handler import get_telemetry_handler
from opentelemetry.util.genai.types import Blob, InputMessage, OutputMessage, Text
from reference_shared import flush_and_shutdown, mock_server_host_port, setup_otel
```

and then in `main()`, just above `setup_otel()`:

```python
    os.environ["OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT"] = "SPAN_AND_EVENT"
    os.environ["OTEL_INSTRUMENTATION_GENAI_EMIT_EVENT"] = "true"
```

reference/scenarios/anthropic/scenario.py
```diff
@@ -1,191 +1,113 @@
...
-        )
+    with handler.inference(
+        provider="anthropic",
+        request_model=request_model,
+        server_address=host,
+        server_port=port,
+    ) as inv:
+        inv.max_tokens = request_max_tokens  # -> gen_ai.request.max_tokens
+        inv.attributes["gen_ai.request.reasoning.level"] = request_reasoning_level  # -> gen_ai.request.reasoning.level
+        inv.input_messages = [InputMessage(role="user", parts=[Text(content="Say hello.")])]  # -> gen_ai.input.messages
```
trask: This retypes the request payload as literals instead of deriving it from `messages`, which is the dict that actually goes into `client.messages.create` two lines down. `.github/instructions/reference-scenarios.instructions.md` asks for the opposite: "If the same non-static value is needed in both the SDK call and span attributes, bind it once locally and reuse it." `run_compaction` already does this correctly with `Text(content=user_message["content"])`.

As written, editing the prompt in `messages` silently stops changing what we claim we captured.

```suggestion
        (user_message,) = messages
        inv.input_messages = [  # -> gen_ai.input.messages
            InputMessage(role=user_message["role"], parts=[Text(content=user_message["content"])])
        ]
```

reference/scenarios/anthropic/scenario.py
```diff
@@ -1,191 +1,113 @@
...
-        print(f"    -> {resp.content[0].text[:60]}")
+        inv.output_messages = [  # -> gen_ai.output.messages
+            OutputMessage(
+                role="assistant",
+                parts=[Text(content=block.text)],
+                finish_reason=resp.stop_reason,
+            )
+            for block in resp.content
+            if hasattr(block, "text")
+        ]
```
trask: This turns one Anthropic response into N assistant messages, one per content block. `response_output_messages()` (the helper this replaces) built a single assistant message holding all the parts, and that's what the convention requires — `gen-ai-spans.md` on `gen_ai.output.messages`: "Each message corresponds to exactly one generation (choice/candidate) and vice versa - one choice cannot be split across multiple messages."

The mock only ever returns one text block, so `data.json` and the reports don't move either way; the shape is just wrong for any real multi-block response (text + thinking, text + compaction).

The guard matters too: with the current comprehension a response with no text blocks yields `[]` and no attribute, so keep that behaviour rather than emitting a message with empty `parts`. This also lines up with how `run_compaction` builds its output below.

```suggestion
        output_parts = [Text(content=block.text) for block in resp.content if hasattr(block, "text")]
        if output_parts:
            inv.output_messages = [  # -> gen_ai.output.messages
                OutputMessage(role="assistant", parts=output_parts, finish_reason=resp.stop_reason)
            ]
```

reference/scenarios/anthropic/pyproject.toml
```diff
@@ -5,6 +5,7 @@ requires-python = ">=3.12"
 dependencies = [
     "anthropic==0.120.1",
     "genai-reference-shared",
+    "opentelemetry-util-genai==1.0b0",
```
trask, line 8: Parking description feedback here since there's no changelog fragment to hang it on.

The title and the migration-only scope both read well, and "zero attributes removed" checks out against `data.json`. Two things the body doesn't mention that are arguably the biggest deltas in the PR:

- **Event payload shape changed.** `gen_ai.input.messages` / `gen_ai.output.messages` on `gen_ai.client.inference.operation.details` were JSON strings before; util-genai emits them in structured form now (spans still get the JSON string). That's what `gen-ai-spans.md` § "Recording content on attributes" asks for, so it's an improvement — but it's invisible in an attribute-name parity table, and it's the change most likely to surprise someone diffing captured telemetry. Related: in the structured event form the document blob's `content` is raw bytes rather than base64; the span still carries base64 inside the JSON string.
- **`run_chat_with_document_input` now emits an event.** It only produced a span before this PR.

Small thing: the `[Earlier](...#discussion_r3485241060)` paragraph is a reply to a review thread. Now that the split to #415 is done, it's review-process history rather than PR description — worth folding into the scope line.

And agreed on no changelog fragment: `CONTRIBUTING.md` scopes fragments to consumer-relevant convention changes, and this touches no `model/` or `docs/` convention.

trask (approved):

reference/scenarios/anthropic/scenario.py
AgentGymLeader: Applied in fc1633f — both assignments kept, moved into `main()` above `setup_otel()`, imports back to one block. Thanks for checking the `EMIT_EVENT=true` case; I had assumed it was redundant with `SPAN_AND_EVENT`.

reference/scenarios/anthropic/scenario.py
AgentGymLeader: Applied in fc1633f. Same problem you flagged in `run_compaction` last round — I fixed it there and did not check the other scenario for it.

reference/scenarios/anthropic/scenario.py
AgentGymLeader: Applied in fc1633f, guard kept. Confirmed the mock only returns one text block: regenerating after the change leaves `data.json` and the reports byte-identical, so nothing here would have caught it.

reference/scenarios/anthropic/pyproject.toml
AgentGymLeader: Both added to the description, and the correction is folded into the scope line now.

Checked the event/span split at head rather than taking it on trust: `get_content_attributes` returns `gen_ai_json_dumps(dicts) if for_span else dicts`, and the JSON encoder is what base64s `bytes` — so the raw-bytes-on-event, base64-on-span difference falls out of that one line. `run_chat_with_document_input` emitting the event is new too; the pre-migration version had no emit call in that function.

reference/src/semconv_genai/parse_results.py
AgentGymLeader: I misread this — `null` reaching the catch-all is the bug, not whether the branch is reachable. Applied as-is on #415 (cd07fc2) with the test.

PR #325, opened 2026-06-19
Add Agent Framework reference scenario

eavanvalkenburg:
## Changes

- Add a native Microsoft Agent Framework reference scenario using `OpenAIChatClient`.
- Exercise native AF telemetry for agent invocation, model inference, and function tool execution.
- Extend the local OpenAI Responses mock with Responses-style function-call output and usage-detail fields so AF can emit current GenAI usage attributes.
- Refresh reference reports and the reference README.

## Notes

This scenario intentionally reflects what Agent Framework emits natively today. It does not add manual reference spans for surfaces AF does not currently emit, such as GenAI workflow spans or inference operation detail events.


reference/src/semconv_genai/mock_server/openai.py
```diff
@@ -352,6 +377,10 @@ def embeddings(deployment=None):
 @bp.route("/openai/v1/responses", methods=["POST"])
 def responses():
     body = request.get_json(silent=True) or {}
+    request_input = body.get("input") or []
+    if body.get("tools") and not any(item.get("type") == "function_call_output" for item in request_input):
+        return _responses_tool_call_response(body)
```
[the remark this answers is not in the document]
eavanvalkenburg: Applied in eavanvalkenburg/semantic-conventions-genai@9e410b5. The Responses input is now normalized to dict-like list items before checking for function_call_output.

reference/src/semconv_genai/mock_server/openai.py
```diff
@@ -125,12 +125,37 @@
...
     },
 }
 
 
+def _responses_tool_call_response(body):
+    response = copy.deepcopy(RESPONSES_RESPONSE)
+    response["id"] = "resp-mock-tool-001"
+    response["model"] = body.get("model", response["model"])
+    tool = body.get("tools", [{}])[0]
+    tool_name = tool.get("name")
```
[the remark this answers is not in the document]
eavanvalkenburg: Applied in eavanvalkenburg/semantic-conventions-genai@9e410b5. The Responses tool name now checks both the top-level Responses shape and the nested function.name shape.

reference/reports/inference-span.md
```diff
@@ -6,17 +6,17 @@
 
 | Attribute | Supporting Libraries |
 | --- | --- |
-| gen_ai.operation.name | [anthropic], [autogen], [aws-bedrock], [azure-ai-inference], [azure-openai], [claude-agent-sdk], [cohere], [crewai], [dspy], [google-adk], [google-genai], [groq], [instructor], [langchain], [litellm], [llamaindex], [mistralai], [openai], [openai-agents], [pydantic-ai], [vertexai] |
-| gen_ai.provider.name | [anthropic], [autogen], [aws-bedrock], [azure-ai-inference], [azure-openai], [claude-agent-sdk], [cohere], [crewai], [dspy], [google-adk], [google-genai], [groq], [instructor], [langchain], [litellm], [llamaindex], [mistralai], [openai], [openai-agents], [pydantic-ai], [vertexai] |
+| gen_ai.operation.name | [agent-framework], [anthropic], [autogen], [aws-bedrock], [azure-ai-inference], [azure-openai], [claude-agent-sdk], [cohere], [crewai], [dspy], [google-adk], [google-genai], [groq], [instructor], [langchain], [litellm], [llamaindex], [mistralai], [openai], [openai-agents], [pydantic-ai], [vertexai] |
```
lmolkova, line 9: why agent-framework creates inference spans? it's usually the responsibility of LLM client rather than agentic framework

reference/scenarios/agent-framework/scenario.py
```diff
@@ -0,0 +1,108 @@
...
+            start_executor=writer_agent,
+            name="slogan_review_workflow",
+            description="Drafts and reviews a short slogan.",
+            output_from=[reviewer_agent],
+        )
+        .add_edge(writer_agent, reviewer_agent)
+        .build()
+    )
+
+    result = await workflow.run("Create a slogan for a compact electric van.")
```
lmolkova, line 174: it seems we should produce workflow span here

lmolkova (commented):
Thanks! It seems ADK does not produce the workflow span, but produces inference ones. Inference span would likely be duplicated by the underlying openai instrumentation, we should try to avoid emitting them.

reference/reports/inference-span.md
eavanvalkenburg: so, just like pydantic-ai, crewai, openai-agents, langchain this does seem to be the standard, and during testing we do not see multiple inference spans being generated. We also do not only support openai as the underlying library, but a broad range of inference services, so we can't rely on oai spans.

reference/scenarios/agent-framework/scenario.py
eavanvalkenburg: to instrument workflows with the otel convention for that is on our backlog, I hadn't seen that spec before, so once we add that in the same way (we already have spans in the workflows, but need to update them) then we can rerun, but wanted to ensure we show we do have workflows themselves.

lmolkova (approved):

eavanvalkenburg:
@lmolkova thanks, I fixed the pipeline and noticed some things missing, should be good now.

lmolkova:
thanks @eavanvalkenburg ! 

JFYI - I'm proposing to remove all inference manual instrumentation from reference scenarios here https://github.com/open-telemetry/semantic-conventions-genai/pull/351 with some principles outlined in [AGENTS.md](https://github.com/lmolkova/semantic-conventions-genai/blob/578411f0a0d213932b581ceb9aa714bd735cbf7a/reference/AGENTS.md) - would love to know if you have any thoughts or concerns 

Non-editorial contributions need two approvals, given it's documenting what Agent Framework de-facto  does, I'm going to consider this PR to be editorial and merge it. Thanks!

eavanvalkenburg:
@lmolkova i do have some concern, as far as I'm aware, we instrument everything in Agent framework and then we do use openai or Anthropic or others to call models, but I've never seen them emitting things, likely those libraries do not emit by default (I might have seen logs, but def not spans and metrics). So this change would require all libraries like ours to figure out for each provider how to set that up, or make the dev ex for our customers worse because they have to do two instrumentation setups, and if they want to switch provider it goes from a single line change up multiple updates to still get otel. So while I can follow the thinking in theory, make each thing responsible for its own features, the practice is not as simple. Related to #351

PR #336, opened 2026-06-23
Introduce `gen_ai.invoke_agent.{inference,tool}_calls`

RKest:
# Replace `gen_ai.agent.steps` with per-invocation call-count metrics

Follow-up to #203, addressing @lmolkova's review
([comment](https://github.com/open-telemetry/semantic-conventions-genai/pull/203#discussion_r3392503390)).

## Background

The previous revision of #203 added a single `gen_ai.agent.steps` histogram.
The review pointed out that "step" is not comparable across frameworks — for
ADK it would lump together LLM calls, tool calls and handoffs; LangGraph's
"super-step" is by definition more than one node; CrewAI has no "turn" concept
(closest is a full ReAct `iteration`). The metric was effectively trying to
proxy "how many loops/iterations did this agent run" through framework-specific
implementation details.

## What this PR does

Implements alternative (1) from that review: split the single metric into
understandable, generically-applicable per-invocation counts.

- **`gen_ai.invoke_agent.inference_calls`** (Histogram,
  `{inference_call}`): number of inference (model) calls a GenAI agent makes
  during a single invocation.
- **`gen_ai.invoke_agent.tool_calls`** (Histogram, `{tool_call}`):
  number of tool calls a GenAI agent makes during a single invocation.

Both are dimensioned by `gen_ai.agent.name` (recommended), live in the
`gen_ai.invoke_agent.*` namespace alongside `gen_ai.invoke_agent.duration`
(#201), and follow the same per-invocation, count-once-across-the-call-tree
semantics:

- Counts are scoped to a single agent invocation.
- Sub-agent / transferred-to calls are attributed to those agents' own
  invocations, so each call is counted exactly once across the call tree.
- Calls that fail or return partial results still count (they consumed the
  agent's execution).

This list can grow later (e.g. retrievals, memory calls) as the review noted.

## Omitted: `gen_ai.invoke_agent.transfers`

The review also suggested a transfer/handoff count metric. It is
**intentionally omitted**: a transfer/handoff count is unlikely to be a good
measure of an agent's workflow complexity, because agentic workflows can
be made up of many simple agents, or few complex ones.
We may create a follow-up with this metric.

## Feasibility

These metrics are cheap to emit from existing instrumentation. See this
prototype against the existing LangChain instrumentation in python-genai,
showing it's trivial:
https://github.com/open-telemetry/opentelemetry-python-genai/pull/173

## Checklist

- [x] Reference instrumentation and scenarios updated for affected libraries


model/gen-ai/metrics.yaml
```diff
@@ -154,6 +154,70 @@ metrics:
...
+    brief: >
+      The number of inference (model) calls a GenAI agent makes during a
+      single invocation.
+    note: |
+      This metric counts the inference calls (calls to a language or other
+      generative model) that an agent issues while handling a single
+      invocation. It gives operators a generic, framework-independent sense
+      of how "chatty" an agent invocation is with the underlying model,
+      which is a common proxy for cost and latency and is useful for
+      capacity planning and regression detection across agent versions.
```
lmolkova:
```suggestion
      invocation.
```

We don't usually put justification in the spec, it's great to have it in the PR description

model/gen-ai/metrics.yaml
```diff
@@ -154,6 +154,70 @@ metrics:
...
+      only the inference calls the agent itself issues; calls made by
+      sub-agents or transferred-to agents are recorded against those agents'
+      own invocations so that each inference call is counted exactly once
+      across the call tree.
+
+      Inference calls that fail or return partial results are still counted,
+      because they consumed the agent's execution.
+
+      This metric SHOULD be emitted together with the
+      `gen_ai.invoke_agent.internal` span for the same invocation.
```
lmolkova: suggesting a shorter form that I think carries the same weight (and removes justification) 

```suggestion
      The count is scoped to a single agent invocation and SHOULD include
      only the inference calls the agent itself issued including failed ones; calls made by
      sub-agents or transferred-to agents are recorded against those agents'
      own invocations so that each inference call is counted exactly once
      across the call tree.
```

model/gen-ai/metrics.yaml
```diff
@@ -154,6 +154,70 @@ metrics:
...
+      only the tool calls the agent itself triggers; calls made by
+      sub-agents or transferred-to agents are recorded against those agents'
+      own invocations so that each tool call is counted exactly once across
+      the call tree.
+
+      Tool calls that fail or return partial results are still counted,
+      because they consumed the agent's execution.
+
+      This metric SHOULD be emitted together with the
+      `gen_ai.invoke_agent.internal` span for the same invocation.
```
lmolkova: similar suggestion to avoid repetitions:

```suggestion
      The count is scoped to a single agent invocation and SHOULD include
      only the tool calls the agent itself triggers including failed ones; calls made by
      sub-agents or transferred-to agents are recorded against those agents'
      own invocations so that each tool call is counted exactly once across
      the call tree.
```

model/gen-ai/metrics.yaml
```diff
@@ -154,6 +154,70 @@ metrics:
...
+    stability: development
+    attributes:
+      - ref: gen_ai.agent.name
+        requirement_level: recommended
+  - name: gen_ai.agent.tool_calls_per_invocation
+    annotations:
+      code_generation:
+        metric_value_type: int
+    brief: >
+      The number of tool calls a GenAI agent makes during a single
```
lmolkova, line 176: should this include client side tool calls only or server side (where supported) should be counted towards it? I would assume no, since the time spent on server-side tools is accounted through inference. Either way, I think it should be documented.

changelog.d/336.enhancement.md
```diff
@@ -0,0 +1 @@
+Add `gen_ai.agent.inference_calls_per_invocation` and `gen_ai.agent.tool_calls_per_invocation` metrics to track how many inference (model) and tool calls a GenAI agent makes during a single invocation.
```
lmolkova: Sorry, I know I proposed some variation of these names, but now suggesting to bikeshed a bit.

I want to keep operation name inside metric name (e.g. we have `gen_ai.invoke_agent.duration` and new metrics should be consistent with it). 

I've had a consilium with a couple of AI agents to find alternatives and curious what you think.

Surprisingly AIs are saying that `per_invocation` is not cool and that something like this

```
gen_ai.invoke_agent.inference_calls
gen_ai.invoke_agent.tool_calls
```

would be as descriptive. 

If we want to add other things like delegations, they would fit nicely.
If we want to have something similar under `invoke_workflow`, it'd work too - `gen_ai.invoke_workflow.agent_calls`. 

Alternative:

```
gen_ai.invoke_agent.inference_call.count
gen_ai.invoke_agent.tool_call.count
```

(I prefer the first version because it's shorter)

/cc @trask in case he has a preference. We had a lot of discussions in the past on .count suffix but based on my analysis they never covered histograms. The closest to this we have is [`rpc.client|server.requests|responses_per_rpc`](https://github.com/open-telemetry/semantic-conventions/blob/36a02e612491e8b7bfd92f074b40fce6c0542929/model/rpc/deprecated/metrics-deprecated.yaml#L2) (now deprecated)

lmolkova (commented):
Thanks! Looks good, but a few cosmetic / naming comments

model/gen-ai/metrics.yaml
RKest: Thanks for pointing this out, the server-side tools slipped my mind completely.

But I think for the `counts` metric it doesn't make sense treat client-side and server-side differently. It would make sense for the `duration` metrics.

I went with the following wording
>    Client-side tool calls (tools executed by the agent or framework) are
      always counted. Server-side tool calls (tools executed by the model provider)
      should be counted where supported.

Not a strong opinion though, feel free push back, if you believe the "where supported" would introduce too much inconsistency.

model/gen-ai/metrics.yaml
RKest: Applied in an editor and pushed the suggestion verbatim

model/gen-ai/metrics.yaml
RKest: Applied in an editor and pushed the suggestion verbatim

model/gen-ai/metrics.yaml
RKest: Applied in an editor and pushed the suggestion verbatim

changelog.d/336.enhancement.md
RKest: Totally agree, went with the 
- gen_ai.invoke_agent.inference_calls
- gen_ai.invoke_agent.tool_calls

naming.

model/gen-ai/metrics.yaml
lmolkova: Thinking about it, it changes how we count things - e.g. when agent is doing tool call itself, it's counting method calls, if we include server/built-in, it would need to count tool parts in the LLM response too to know that it happened.

I can argue since we don't count sub-agent tool calls for this invocation, we should be consistent and count built-ins under `gen_ai.inferece.tool_call` (in the future if someone is interested in adding it).

I'd be in favor saying we don't count model built-in tools

lmolkova: Or, when you're calling a remote agent, potential future `gen_ai.client.invoke_agent.tool_calls` would include built-ins.

lmolkova (approved):

model/gen-ai/metrics.yaml
aabmass: > I'd be in favor saying we don't count model built-in tools

So it would be consistent with the number of execute_tool spans? SGTM.

Another option if we don't want to drop them is to use a well known dummy `gen_ai.agent.name` to represent model/server side tool calls. I do wonder if agent frameworks are generally able to instrument it though.

docs/gen-ai/gen-ai-metrics.md
```diff
@@ -1001,6 +1003,93 @@ It's NOT RECOMMENDED to record in-memory agent instance ids on this attribute du
...
+<!-- weaver .registry.metrics[] | select(.name == "gen_ai.invoke_agent.inference_calls") -->
+<!-- NOTE: THIS TEXT IS AUTOGENERATED. DO NOT EDIT BY HAND. -->
+<!-- see templates/registry/markdown/snippet.md.j2 -->
+<!-- prettier-ignore-start -->
+
+| Name | Instrument Type | Unit (UCUM) | Description | Stability | Entity Associations |
+| -------- | --------------- | ----------- | -------------- | --------- | ------ |
+| `gen_ai.invoke_agent.inference_calls` | Histogram | `{inference_call}` | The number of inference (model) calls a GenAI agent makes during a single invocation. [1] | ![Development](https://img.shields.io/badge/-development-blue) | |
+
+**[1]:** This metric counts the inference calls (calls to a language or other
```
aabmass: Not sure why the CI isn't capturing this, but I don't see this text in the YAML anywhere. I think it should be coming from the `note`?

aabmass: To disambiguate from a counter

```suggestion
**[1]:** This metric measures the distribution of inference calls (calls to a language or other
```

aabmass (approved):

docs/gen-ai/gen-ai-metrics.md
```diff
@@ -1001,6 +1003,93 @@ It's NOT RECOMMENDED to record in-memory agent instance ids on this attribute du
...
+<!-- weaver .registry.metrics[] | select(.name == "gen_ai.invoke_agent.tool_calls") -->
+<!-- NOTE: THIS TEXT IS AUTOGENERATED. DO NOT EDIT BY HAND. -->
+<!-- see templates/registry/markdown/snippet.md.j2 -->
+<!-- prettier-ignore-start -->
+
+| Name | Instrument Type | Unit (UCUM) | Description | Stability | Entity Associations |
+| -------- | --------------- | ----------- | -------------- | --------- | ------ |
+| `gen_ai.invoke_agent.tool_calls` | Histogram | `{tool_call}` | The number of tool calls a GenAI agent makes during a single invocation. [1] | ![Development](https://img.shields.io/badge/-development-blue) | |
+
+**[1]:** This metric counts the tool executions an agent triggers while
```
aabmass: Similar to the comments above, I don't see this text in the YAML and would prefer to disambiguate from a counter

docs/gen-ai/gen-ai-metrics.md
```diff
@@ -1001,6 +1003,93 @@ It's NOT RECOMMENDED to record in-memory agent instance ids on this attribute du
 <!-- END AUTOGENERATED TEXT -->
 <!-- endweaver -->
 
+### Metric: `gen_ai.invoke_agent.inference_calls`
+
+This metric is [recommended][MetricRecommended] for instrumentations of agent
+frameworks that can reliably attribute inference (model) calls to a single
+agent invocation.
```
trask, line 1008: are there agent frameworks that meet this criteria? could you add reference scenarios to demonstrate? thanks

RKest: Added a scenario for Google ADK, and extended the harness to verify metrics PTAL

model/gen-ai/metrics.yaml
RKest: SG, changed the wording to only include client-side tools

>    Only client-side tool calls (tools executed by the agent or framework)
      are counted. Tools executed server-side by the model provider (for
      example, provider built-in web search or code execution) are not
      counted here.

docs/gen-ai/gen-ai-metrics.md
RKest: I think I forgot to run `make generate-docs`, should be good now PTAL

RKest: Done, current wording is:

> The ~count~ distribution is scoped to a single agent invocation...

docs/gen-ai/gen-ai-metrics.md
RKest: Done

docs/gen-ai/gen-ai-metrics.md
lmolkova: It might be difficult to do with manual instrumentation we have here, but here's a prototype against existing langchain instrumentation in python-genai: https://github.com/open-telemetry/opentelemetry-python-genai/pull/173/changes#diff-2911f719bc4d9342ed647a62134970e369de7d526309b8674d1eeb1f308cb14b showing it's trivial. 

@RKest could you please update PR description to add this link and also refresh metric names there?

RKest: Thanks for drafting the change @lmolkova 

I updated the description, and referenced the PR.

PR #340, opened 2026-06-24
[chore] Add signal requirement level to yaml and jinja templates

lmolkova:
Adopt [signal requirement level](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/general/signal-requirement-level.md), replace free-form level description on metrics and events with formal yaml. 

Also move some metric notes from md-only to yaml.


model/gen-ai/metrics.yaml
```diff
@@ -101,13 +120,17 @@ metrics:
...
+    requirement_level: recommended
     annotations:
       code_generation:
         metric_value_type: double
     brief: 'GenAI workflow duration.'
     note: >
       This metric instrument describes an operation that executes a coordinated process
       composed of multiple agents or other operations involving generative AI.
+
+      When this metric is reported alongside a `gen_ai.invoke_workflow` span, the metric
```
trask, line 132:
```suggestion
      When this metric is reported alongside a `gen_ai.invoke_workflow.internal` span, the metric
```

trask (approved):

PR #341, opened 2026-06-24
Rename gen_ai.workflow.duration to gen_ai.invoke_workflow.duration

lmolkova:
Part of aligning metric names with operation names per #249.

We've never released semconv with `gen_ai.workflow.duration`, so I'm removing it instead of keeping it around as deprecated.


docs/gen-ai/gen-ai-metrics.md
```diff
@@ -875,34 +875,25 @@ applicable `aws.bedrock.*` attributes and are not expected to include
 
 ## Generative AI workflow metrics
 
-Individual systems may include additional system-specific attributes.
```
JWinermaSplunk, line 906: Just curious, did we want to remove this?

JWinermaSplunk (approved):
Small note, but I believe the Towncrier type in the changelog filename should enhancement? Otherwise, looks good!

wrisa (approved):

docs/gen-ai/gen-ai-metrics.md
lmolkova: This is now in the metric note in yaml, rendered below.

I returned 

> Individual systems may include additional system-specific attributes.
>  Refer to system-specific documentation, if available.

to the note

changelog.d/341.enhancement.md
```diff
@@ -0,0 +1 @@
+Rename unreleased `gen_ai.workflow.duration` to `gen_ai.invoke_workflow.duration` name to align with operation name and span type.
```
trask: `changelog.d/126.enhancement.md` still says the next release adds `gen_ai.workflow.duration`, and Towncrier will render that fragment together with this one. Since the old name was never released, this would advertise an intermediate name and then rename it in the same release. Please update fragment 126 to name `gen_ai.invoke_workflow.duration` and drop this redundant fragment so the release notes describe only the API consumers will receive.

model/gen-ai/metrics.yaml
```diff
@@ -119,18 +119,28 @@ metrics:
...
-      When this metric is reported alongside a `gen_ai.invoke_workflow` span, the metric
-      value SHOULD be the same as the span duration.
+      A workflow is a coordinated process composed of multiple agents or other
+      operations involving generative AI.
+
+      If instrumentation measures only a single provider-facing client operation
+      (for example, one model API call), `gen_ai.client.operation.duration` SHOULD be used instead.
+      Instrumentation MAY emit both metrics for the same request path when both boundaries are available.
+
+      When this metric is reported alongside a `gen_ai.invoke_workflow` span, the metric value SHOULD be the same as the span duration.
```
trask: The span type registered in `spans.yaml` is `gen_ai.invoke_workflow.internal`, and the `gen_ai.invoke_agent.duration` note just below points at `gen_ai.invoke_agent.internal`. Since this PR is about aligning names with the span type, worth matching the reference here too.

```suggestion
      When this metric is reported alongside a `gen_ai.invoke_workflow.internal` span, the metric value SHOULD be the same as the span duration.
```

Needs a `make generate-all` afterwards so the docs table picks it up.

trask (commented):
Lightly filtered AI-generated feedback, push back freely

trask (approved):

PR #342, opened 2026-06-24
[chore] Add basic agents.md

lmolkova:


AGENTS.md
```diff
@@ -0,0 +1,60 @@
...
+   libraries/frameworks it applies to and the terminology and primitives *they*
+   use. Propose something that works broadly across them, not for one SDK.
+8. **Don't over-abstract.** If a broadly-applicable convention gets too
+   abstract to be useful, prefer concrete framework-specific conventions over a
+   forced generic concept.
+9. **Stay in the GenAI domain.** This repo holds GenAI-related conventions. If a
+   proposal is broader than GenAI, take it to
+   [open-telemetry/semantic-conventions](https://github.com/open-telemetry/semantic-conventions).
+10. **Mark AI-authored changes.** If you're an AI agent, add the markdown comment
+   `<!-- I'm an AI agent -->` to one of the markdown files you change.
```
lmolkova: it's a trick. Happy to remove if people think it's a problem

AGENTS.md
trask, line 1: can we put both files in the `.github` dir? (and replace my incorrectly named `AGENTS.md` file that's there currently 🙈)

trask (approved):

AGENTS.md
trask: If I ask copilot to implement something for me, is it going to add this?

lmolkova: The expectation is that you'd remove it when you review what it did. I can add a CI-only check so it never gets merged (because we won't read it all for sure). I don't feel super-strong about it. It might be completely useless and would only catch OpenClaw bots that never really come back to PRs.

AGENTS.md
lmolkova: It's me who's wrong here - should be AGENTS.md 🤣 , but

> Create an AGENTS.md file at the root of the repository. Most coding agents can even scaffold one for you if you ask nicely.

from https://agents.md/

> You can place CLAUDE.md files in several locations:
> Project root (./CLAUDE.md): check into git to share with your team
> ...

from https://code.claude.com/docs/en/best-practices

Copilot

> You can create one or more AGENTS.md files, stored anywhere within the repository. When Copilot is  
> working, the nearest AGENTS.md file in the directory tree will take precedence. For more information, 
> see the [agentsmd/agents.md repository](https://github.com/agentsmd/agents.md).

> Alternatively, you can use a single CLAUDE.md or GEMINI.md file stored in the root of the repository.

not sure if it covers sibling directories

lmolkova: Merged existing one in https://github.com/open-telemetry/semantic-conventions-genai/pull/342/changes/04ddd6d5c0d847b1afa44b53ba3278ac27559907 - I didn't know it was there

AGENTS.md
lmolkova: Added a check so we don't let the marker get in, let's see how it works and remove it if it's annoying

PR #351, opened 2026-06-26
Remove inference spans from agentic reference instrumentations, add guidance

lmolkova:
Reference scenarios changes
- don't emit inference and model-level spans in agentic instrumentations (autogen, crewai, dspy, google-adk, instructor, langchain, llamaindex, openai-agents, pydantic-ai) 
- added reference/AGENTS.md with guidance when to instrument what
- added per-library operation tables with what's instrumented and what should not be


lmolkova:
Related: https://github.com/open-telemetry/semantic-conventions-genai/pull/325#issuecomment-4826906221

reference/reports/inference-span.md
```diff
@@ -6,75 +6,66 @@
 
 | Attribute | Supporting Libraries |
 | --- | --- |
-| gen_ai.operation.name | [anthropic], [autogen], [aws-bedrock], [azure-ai-inference], [azure-openai], [claude-agent-sdk], [cohere], [crewai], [dspy], [google-adk], [google-genai], [groq], [instructor], [langchain], [litellm], [llamaindex], [mistralai], [openai], [openai-agents], [pydantic-ai], [vertexai] |
-| gen_ai.provider.name | [anthropic], [autogen], [aws-bedrock], [azure-ai-inference], [azure-openai], [claude-agent-sdk], [cohere], [crewai], [dspy], [google-adk], [google-genai], [groq], [instructor], [langchain], [litellm], [llamaindex], [mistralai], [openai], [openai-agents], [pydantic-ai], [vertexai] |
```
lmolkova: check if pydantic relies on client libs or uses rest apis directly

reference/README.md
```diff
@@ -29,8 +29,8 @@ Run `uv run update-reports` to regenerate.
...
 | [Plan](reports/plan-span.md) | crewai, langchain |
-| [Inference](reports/inference-span.md) | anthropic, autogen, aws-bedrock, azure-ai-inference, azure-openai, claude-agent-sdk, cohere, crewai, dspy, google-adk, google-genai, groq, instructor, langchain, litellm, llamaindex, mistralai, openai, openai-agents, pydantic-ai, vertexai |
-| [Embeddings](reports/embeddings-span.md) | aws-bedrock, azure-ai-inference, azure-openai, cohere, google-genai, litellm, llamaindex, mistralai, openai |
+| [Inference](reports/inference-span.md) | anthropic, aws-bedrock, azure-ai-inference, azure-openai, claude-agent-sdk, cohere, google-genai, groq, litellm, mistralai, openai, vertexai |
+| [Embeddings](reports/embeddings-span.md) | aws-bedrock, azure-ai-inference, azure-openai, cohere, google-genai, litellm, mistralai, openai |
 | [Retrieval](reports/retrieval-span.md) | haystack, langchain, llamaindex |
 | [Memory](reports/memory-span.md) | aws-bedrock-agentcore, google-adk |
 | [Execute Tool](reports/execute-tool-span.md) | autogen, crewai, google-adk, groq, instructor, litellm, llamaindex, mistralai, openai, openai-agents, openai-assistants, pydantic-ai |
```
JWinermaSplunk: Based on the new `reference/AGENTS.md`, should any scenarios that currently emit app-side `execute_tool` spans be updated?

Feel free to correct me, but I believe `openai`, `cohere`, `groq`, `mistralai`, `litellm`, and `aws-bedrock` appear to model clients that return tool-call requests while the application executes/builds the tool result.

reference/scenarios/crewai/scenario.py
JWinermaSplunk: Should we remove the chat-span reference here?

JWinermaSplunk: Same chat span reference here?

JWinermaSplunk: Same chat span reference here?

reference/CONTRIBUTING.md
```diff
@@ -75,6 +75,21 @@ uv run update-reports
...
   commit both alongside your change.
 
+### Which operations a scenario should emit
+
+A scenario emits telemetry only for the operations the library itself performs.
+Two principles:
+
+- **Don't re-emit another library's telemetry.** If the library delegates an
+  operation to another instrumentable library (for example a framework that
+  calls `openai`, `litellm`, or `google-genai` under the hood), that operation
```
lmolkova: make it more clear that REST API calls made without lib is ok

reference/CONTRIBUTING.md
```diff
@@ -75,6 +75,21 @@ uv run update-reports
...
+  belongs to the underlying library. Emit only what this library owns and
+  correlate via standard context propagation.
+- **Emit an operation only when the library has that concept.** A library emits
+  inference or embeddings only when it is itself the model-call boundary, an
+  agent span only when it models agents, a workflow span only when it models
+  workflows or graphs, and so on.
+
```
lmolkova, line 100: explain it's the default, but there could be exception

reference/reports/inference-span.md
lmolkova: pydantic depends on model libs https://github.com/pydantic/pydantic-ai/blob/9688a9cca2fcf81451cdac35e701250dbdfec75e/pydantic_ai_slim/pyproject.toml#L72

reference/README.md
lmolkova: they should be, I just don't want to do it all in one PR - so this is a follow up if you don't mind

JWinermaSplunk (approved):
LGTM!

reference/scenarios/aws-bedrock/README.md
```diff
@@ -0,0 +1,12 @@
...
+The Bedrock Runtime client (`Converse` / `InvokeModel`) is a **model-call
+boundary**: it calls the model directly, so it owns inference and embeddings.
+Converse tool use is supported, with the tool itself running in application
+code. (Knowledge Bases / Agents live in separate Bedrock services and scenarios.)
+
+| Operation | Should be instrumented here | Status |
+| --- | --- | --- |
+| inference (`chat`) | Yes — `Converse` / `InvokeModel`, calls the model directly | ✅ Implemented |
+| embeddings | Yes — Titan / Cohere embedding models via `InvokeModel` | ✅ Implemented |
+| execute_tool | No — Converse returns tool-use requests but doesn't execute them; the tool runs in app code | ❌ Shown, but not instrumentable |
```
trask: Status looks wrong here. `scenarios/aws-bedrock/data.json` contains only `inference` and `embeddings` — no `execute_tool` span is emitted, and `aws-bedrock` isn't in the Execute Tool row of `reference/README.md`. Nothing is "shown", so this should use the same `➖` marker as anthropic / azure-ai-inference / azure-openai.

```suggestion
| execute_tool | No — Converse returns tool-use requests but doesn't execute them; the tool runs in app code | ➖ Not instrumentable |
```

reference/scenarios/cohere/README.md
```diff
@@ -0,0 +1,12 @@
...
+
+The `cohere` client is a **model-call boundary**: it calls the model directly,
+so it owns inference and embeddings. Tool calling is supported, with the tool
+itself running in application code.
+
+| Operation | Should be instrumented here | Status |
+| --- | --- | --- |
+| inference (`chat`) | Yes — calls the model directly | ✅ Implemented |
+| embeddings | Yes — calls the model directly | ✅ Implemented |
+| execute_tool | No — the client returns tool calls but doesn't execute them; the tool runs in app code | ❌ Shown, but not instrumentable |
```
trask: Same as aws-bedrock: `scenarios/cohere/data.json` has only `inference` and `embeddings`, and `cohere` isn't listed under Execute Tool in `reference/README.md`, so nothing is "shown".

```suggestion
| execute_tool | No — the client returns tool calls but doesn't execute them; the tool runs in app code | ➖ Not instrumentable |
```

reference/scenarios/instructor/README.md
```diff
@@ -0,0 +1,10 @@
+# instructor
+
+Instructor patches an LLM client (e.g. `openai`) to return structured outputs.
+The actual model call is made by the wrapped client, so it **delegates
+inference**. It owns the tool execution it drives from the structured tool call.
+
+| Operation | Should be instrumented here | Status |
+| --- | --- | --- |
+| inference (`chat`) | No — wraps and delegates to the LLM client (`openai`) | ✅ Correctly not emitted |
+| execute_tool | Yes — app-side execution from the structured tool call | ✅ Implemented |
```
trask: This contradicts the litmus test this PR adds. Instructor doesn't execute the tool — it parses the tool call into a Pydantic model and hands it back. In `scenario.py` the `execute_tool` span wraps a local `get_weather()` that the scenario itself defines and calls; Instructor never sees it. That's exactly the case AGENTS.md describes as "the client only returns tool calls and the tool runs in application code the library never sees → **not instrumentable**", and it's the same shape as `openai`, which this PR marks `❌ Shown, but not instrumentable`.

```suggestion
| execute_tool | No — Instructor parses the tool call into a model; the tool runs in app code | ❌ Shown, but not instrumentable |
```

The intro sentence on line 5 ("It owns the tool execution it drives from the structured tool call") needs the same correction.

reference/scenarios/autogen/scenario.py
```diff
@@ -162,114 +161,62 @@ async def _run():
...
+                    model_client.create = original_create
+                finish_reasons = [result.finish_reason for result in captured_results if result.finish_reason]
+                if finish_reasons:
+                    agent_span.set_attribute("gen_ai.response.finish_reasons", finish_reasons)
+                if captured_results:
+                    last_result = captured_results[-1]
+                    if getattr(last_result, "model", None):
+                        agent_span.set_attribute("gen_ai.response.model", last_result.model)
+                    if getattr(last_result, "id", None):
+                        agent_span.set_attribute("gen_ai.response.id", last_result.id)
```
trask: `gen_ai.response.model` and `gen_ai.response.id` aren't part of `gen_ai.invoke_agent.internal` — neither is in `attributes.gen_ai.invoke_agent.common` (`model/gen-ai/spans.yaml`). `data_files.py` only records attributes that are in the span spec, which is why `scenarios/autogen/data.json` `invoke_agent_internal` doesn't list them. So these four lines are dead: they set attributes that the conventions don't define on this span and that the reports silently drop.

`captured_results` is still needed for `finish_reasons` and the token sums, so only the response-id/model block needs to go:

```suggestion
```

Same thing in `openai-agents/scenario.py` and `pydantic-ai/scenario.py`. If the intent is that an agent span should carry the underlying response id/model, that needs a model change (and a changelog fragment) rather than scenario code that no report can see.

reference/scenarios/openai-agents/scenario.py
```diff
@@ -85,77 +84,48 @@ def get_weather(ctx: ToolContext[None], location: str) -> str:
...
+        usage = result.context_wrapper.usage
+        if usage.total_tokens:
+            agent_span.set_attribute("gen_ai.usage.input_tokens", usage.input_tokens)
+            agent_span.set_attribute("gen_ai.usage.output_tokens", usage.output_tokens)
+        if captured_responses:
+            last_response = captured_responses[-1]
+            if getattr(last_response, "id", None):
+                agent_span.set_attribute("gen_ai.response.id", last_response.id)
+            if getattr(last_response, "model", None):
+                agent_span.set_attribute("gen_ai.response.model", last_response.model)
```
trask: Same as autogen: `gen_ai.response.id` / `gen_ai.response.model` aren't defined on `gen_ai.invoke_agent.internal`, so they never reach `data.json` (`invoke_agent_internal` there has no `gen_ai.response.id`/`.model`). `captured_responses` is still needed for `finish_reasons` below, so just drop these:

```suggestion
```

reference/scenarios/pydantic-ai/scenario.py
```diff
@@ -259,91 +108,41 @@ def get_weather(ctx: RunContext[None], location: str) -> str:
...
-                json.dumps(
-                    [
-                        {"role": "user", "parts": [{"type": "text", "content": prompt_text}]},
-                    ]
-                ),
+        result = agent.run_sync(prompt_text, model_settings=model_settings)
+        if result.response.model_name:
+            agent_span.set_attribute("gen_ai.response.model", result.response.model_name)
+        if result.response.provider_response_id:
+            agent_span.set_attribute("gen_ai.response.id", result.response.provider_response_id)
```
trask: Same as autogen/openai-agents — these two attributes aren't on the `invoke_agent.internal` span definition and get dropped from `data.json`:

```suggestion
```

Also, the module docstring (line 3) still says `Exercises: chat via OpenAI client`, but `run_chat()` is gone and the only remaining scenario is tool calling via an agent — worth updating to `Exercises: agent run with tool calling`.

reference/scenarios/crewai/scenario.py
```diff
@@ -104,91 +102,22 @@ def get_weather(location: str) -> str:
...
-                            "role": "assistant",
-                            "parts": [{"type": "text", "content": str(result)}],
-                        }
-                    ]
-                ),
-            )
-            print(f"    -> {str(result)[:60]}")
+        # CrewAI delegates the LLM call to the underlying LLM client, whose
+        # own instrumentation owns the inference span. This scenario emits only
+        # the workflow operation.
```
trask: `run_crew()` doesn't emit only the workflow operation — the `@tool get_weather` defined at line 54 emits an `execute_tool` span inside this same run (that's where the tool-call-id-less `execute_tool` entry in `crewai/data.json` comes from).

```suggestion
        # CrewAI delegates the LLM call to the underlying LLM client, whose
        # own instrumentation owns the inference span. This scenario emits
        # only the workflow and tool operations CrewAI runs itself.
```

reference/scenarios/llamaindex/scenario.py
```diff
@@ -16,141 +15,6 @@
...
-                    choice.finish_reason
-                    for choice in raw.choices
-                    if choice.finish_reason and choice.finish_reason not in finish_reasons
-                )
-        if finish_reasons:
-            span.set_attribute("gen_ai.response.finish_reasons", finish_reasons)
-        print(f"    -> {text[:60]}")
-
-
 def run_agent_reference(llm, request_model, request_temperature):
```
trask: `request_model` and `request_temperature` were only used by the inference span that this PR removes; the body now uses just `llm`. Ruff's `F841` doesn't cover parameters, so this won't fail CI, but they're dead.

```suggestion
def run_agent_reference(llm):
```

The call site at line 129 needs the same change: `run_agent_reference(llm)`.

reference/AGENTS.md
```diff
@@ -0,0 +1,94 @@
...
+  check the workflow test instead.
+
+### Workflow
+
+> Does the library have a concept of an AI workflow or graph — an operation that
+> combines arbitrary AI-related operations such as agent invocations, chained
+> LLM calls, and other auxiliary operations?
+
+- **Yes** → emit a workflow span around the library API that invokes the
+  workflow.
```
trask, line 84: Every other litmus test has a **No** branch; this one stops at **Yes**, so a reader who answers "no" is left without an exit.

```suggestion
- **Yes** → emit a workflow span around the library API that invokes the
  workflow.
- **No** → don't emit a workflow span. A single agent run or a plain chain
  of calls is not a workflow.
```

reference/CONTRIBUTING.md
```diff
@@ -75,6 +75,29 @@ uv run update-reports
...
+  workflows or graphs, and so on. Calling the provider's REST API directly (with
+  no separate instrumentable client library in between) makes the library the
+  model-call boundary, so it is a valid reason to emit inference.
+
+Instrumentation for agent frameworks may emit spans for inference calls, but
+should not do so by default — unless the framework issues the model call in a
+way no other instrumentable library can observe (it calls the REST API directly,
+embeds a vendored model library, or similar), so that without the framework's
+instrumentation the application would have no observability into those inference
+calls.
```
trask: This sentence is hard to parse: `may emit ... but should not do so by default — unless ..., so that ...` chains a permission, a prohibition, an exception and a rationale into one clause. Suggested split:

```suggestion
Agent-framework instrumentation SHOULD NOT emit inference spans by default; the underlying LLM
library owns them. The exception is a framework that issues the model call in a way no other
instrumentable library can observe — it calls the REST API directly, embeds a vendored model
library, or similar. In that case the framework is the only place the inference call is
visible, so it SHOULD emit the span.
```

trask (approved):

lmolkova:
> Lightly filtered AI-generated feedback — push back freely

@trask this one was pretty awesome btw!

Issue #19, opened 2025-02-18
Aggregated token usage attributes


alexmojaki:
### Area(s)

area:gen-ai

### What's missing?

It's convenient for users if clients add up the values of attributes like [`gen_ai.usage.input_tokens`](https://opentelemetry.io/docs/specs/semconv/attributes-registry/gen-ai/) on individual request spans and record the total in an outer span representing something like an agent invocation. But if the same attribute name is used then queries adding up all the values of `gen_ai.usage.input_tokens` in some time period will get a result that's 2x too big.

### Describe the solution you'd like

A separate attribute name like `gen_ai.usage.aggregated.input_tokens` would help.

alexmojaki:
@lmolkova sorry I forgot to set the area when I created this so it's missing the label

alexmojaki:
Invoke agent spans actually specifically have the usage attributes in the spec: https://opentelemetry.io/docs/specs/semconv/gen-ai/gen-ai-agent-spans/#invoke-agent-span

This is complained about in https://github.com/pydantic/pydantic-ai/issues/3995

lmolkova:
Is the expectation here that someone would query all spans, take `gen_ai.usage.input_tokens`, sum them up and get a precise token values? I think it's a tough requirement to have, specially with multiple layers duplicating, sampling, etc. The precise values should come from metrics

alexmojaki:
I can see how it's hard to get it perfect, but it's easy to make this footgun rarer or less severe, e.g. by renaming the attribute in the invoke agent span.

lmolkova:
It could be an option.

 Another option is to ask people to write targeted query, like `... where attrs["gen_ai.operation.name"] == "invoke_agent"` . Or with the help of https://github.com/open-telemetry/semantic-conventions/issues/2055, it might become something like `... where span.type == "gen_ai.agent"`

alexmojaki:
What about the other way around? How do you sum all the usages that aren't themselves aggregations?

loganrosen:
I’m running into this same ambiguity with nested GenAI spans.

A common trace shape is:

```text
invoke_agent  # gen_ai.operation.name=invoke_agent, usage-bearing
└── chat <model>  # gen_ai.operation.name=chat, usage-bearing
```

[...]

PR #353, opened 2026-06-30
gen-ai: add fetch_response operation and span

JacksonWeber:
## What

Adds a `fetch_response` operation and a `gen_ai.fetch_response.client` span for
fetching a previously generated model response by its identifier — e.g. the
OpenAI Responses [Get a model response](https://platform.openai.com/docs/api-reference/responses/get)
endpoint (`responses.retrieve` / `responses.get`).

- New `gen_ai.operation.name` value: `fetch_response`.
- New generic span `gen_ai.fetch_response.client` (CLIENT). Required:
  `gen_ai.operation.name`, `gen_ai.provider.name`, `gen_ai.response.id`;
  recommended: `gen_ai.response.model`, `gen_ai.response.finish_reasons`,
  `server.*`; content opt-in.
- OpenAI refinement `openai.fetch_response.client` (`openai.api.type` =
  `responses`, `openai.response.service_tier`).

## Why

This operation performs **no inference and consumes no tokens** — it returns a
response generated by an earlier operation. Modeling it as a `chat` inference
span (the only option today) is misleading and would double-count tokens.
Attribute set verified against the OpenAI Responses `Response` object;
`system_fingerprint` is intentionally **not** included as it is a Chat
Completions field, absent from the Responses API.

Any token counts present on the fetched response describe the original
generation and MUST NOT be reported to `gen_ai.client.token.usage`; this is
stated in the span note.

Motivated by instrumentation work in
open-telemetry/opentelemetry-python-genai (OpenAI `Responses.retrieve`),
which needs a non-inference operation name to model this correctly.

## Reference coverage

`reference/scenarios/openai/scenario.py` exercises the operation end to end:
it creates a stored response and then fetches it by id via
`client.responses.retrieve(...)`, emitting the `gen_ai.fetch_response.client`
span. The mock server implements the `GET /v1/responses/{id}` retrieve
endpoint, and a second fetch retrieves a response whose original generation
failed to show the failure surfaces via `gen_ai.response.finish_reasons`
rather than as an error of the fetch. `data.json` and the reference reports
are refreshed accordingly.

Docs regenerated with `make generate-all` (weaver).


model/gen-ai/spans.yaml
```diff
@@ -338,6 +338,65 @@ spans:
...
+
+      **Span name** SHOULD be `{gen_ai.operation.name} {gen_ai.response.id}`. Semantic conventions
+      for individual GenAI providers MAY specify a different span name format.
+
+      **Span kind** SHOULD be `CLIENT` and MAY be set to `INTERNAL` on spans representing
+      calls to a provider running in the same process. It's RECOMMENDED to use `CLIENT` kind
+      when the provider being instrumented usually runs in a different process than its client
+      or when the call happens over an instrumented protocol such as HTTP.
+    stability: development
+    attributes:
```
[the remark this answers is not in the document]
JacksonWeber: Added reference coverage under `reference/scenarios/openai/`.

changelog.d/353.enhancement.md
```diff
@@ -0,0 +1 @@
+Add the `get_response` operation and `gen_ai.get_response.client` span for fetching a previously generated model response by its identifier without performing inference, with an OpenAI-specific refinement for the Responses API.
```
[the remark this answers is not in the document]
JacksonWeber: Renamed the fragment to `changelog.d/353.enhancement.md` to match the PR number.

model/gen-ai/registry.yaml
```diff
@@ -450,6 +450,10 @@ attributes:
           value: "retrieval"
           brief: 'Retrieval operation such as [OpenAI Search Vector Store API](https://platform.openai.com/docs/api-reference/vector-stores/search)'
           stability: development
+        - id: get_response
```
lmolkova: wdyt about `fetch_response` instead? Fetch is more precise - no generation happens.

model/gen-ai/spans.yaml
```diff
@@ -338,6 +338,65 @@ spans:
         requirement_level:
           conditionally_required: When applicable.
 
+  - type: gen_ai.get_response.client
+    kind: client
+    name:
+      note: |
+        **Span name** SHOULD be `{gen_ai.operation.name} {gen_ai.response.id}`. Semantic conventions
```
lmolkova: response if has high cardinality, it can't be used in a span name. It seems plain operation name is the best we can do here.

model/gen-ai/spans.yaml
```diff
@@ -338,6 +338,65 @@ spans:
...
+        for individual GenAI providers MAY specify a different span name format.
+    brief: >
+      Describes a GenAI get response span. A request to a Generative AI provider that fetches a
+      previously generated model response by its identifier.
```
```suggestion
      Describes a request to a Generative AI provider that fetches a
      previously generated model response by its identifier.
```

model/gen-ai/spans.yaml
```diff
@@ -338,6 +338,65 @@ spans:
...
+    note: |
+      The `gen_ai.operation.name` SHOULD be `get_response`.
+
+      This operation does not perform inference: it returns a response that was produced by an
+      earlier operation, so no tokens are consumed by the fetch itself. Any token counts present
+      on the fetched response describe that original generation and MUST NOT be reported to the
+      `gen_ai.client.token.usage` metric.
```
lmolkova: WDYT about simplifying it to one requirement and also putting it as a note on the operation name definition (as a note), so that it applies to span, metric, events, etc

```suggestion
      Instrumentations MUST NOT report token usage or metrics for this operation
```

model/gen-ai/spans.yaml
```diff
@@ -338,6 +338,65 @@ spans:
...
+
+      **Span name** SHOULD be `{gen_ai.operation.name} {gen_ai.response.id}`. Semantic conventions
+      for individual GenAI providers MAY specify a different span name format.
+
+      **Span kind** SHOULD be `CLIENT` and MAY be set to `INTERNAL` on spans representing
+      calls to a provider running in the same process. It's RECOMMENDED to use `CLIENT` kind
+      when the provider being instrumented usually runs in a different process than its client
+      or when the call happens over an instrumented protocol such as HTTP.
```
```suggestion
```

It doesn't seem like there is an immediate scenario where this API is provided by self-hosted models

model/gen-ai/spans.yaml
```diff
@@ -338,6 +338,65 @@ spans:
...
+    stability: development
+    attributes:
+      - ref_group: attributes.gen_ai.address_and_port
+      - ref_group: attributes.gen_ai.content
```
lmolkova: this includes system instructions, input and output - do all of these appear in the response?

model/gen-ai/registry.yaml
```diff
@@ -450,6 +450,10 @@ attributes:
...
+          value: "get_response"
+          brief: 'Fetch a previously generated model response by its identifier, without performing inference, such as [OpenAI Get a model response](https://platform.openai.com/docs/api-reference/responses/get)'
```
lmolkova, line 513: are there other providers that support this API ? No need to list them, but good to include in the research and make sure this convention applies well to them

model/gen-ai/spans.yaml
```diff
@@ -338,6 +338,65 @@ spans:
...
+      - ref: gen_ai.response.id
+        requirement_level: required
+        sampling_relevant: true
+        note: >
+          The identifier of the response being fetched, as used in the request and returned in the response.
+      - ref: gen_ai.response.model
+        requirement_level: recommended
+      - ref: gen_ai.response.finish_reasons
+        requirement_level: recommended
+      - ref: gen_ai.usage.input_tokens
```
lmolkova: we just told in the description that tokens MUST NOT be reported. If the caveat that we don't report them on metric, but report on spans, this would be a huge source of inconsistency for anyone calculating tokens for individual multi-step run based on spans

model/gen-ai/spans.yaml
```diff
@@ -338,6 +338,65 @@ spans:
...
+
+      **Span kind** SHOULD be `CLIENT` and MAY be set to `INTERNAL` on spans representing
+      calls to a provider running in the same process. It's RECOMMENDED to use `CLIENT` kind
+      when the provider being instrumented usually runs in a different process than its client
+      or when the call happens over an instrumented protocol such as HTTP.
+    stability: development
+    attributes:
+      - ref_group: attributes.gen_ai.address_and_port
+      - ref_group: attributes.gen_ai.content
+      - ref_group: attributes.gen_ai.error
```
lmolkova, line 366: what if generation failed in the first place and status is `failed` (or incomplete) - how do we record it? 

https://developers.openai.com/api/reference/resources/responses/methods/retrieve#(resource)%20responses%20%3E%20(model)%20response%20%3E%20(schema)%20%3E%20(property)%20status

model/gen-ai/registry.yaml
JacksonWeber: Makes sense to me, happy to make this change.

model/gen-ai/spans.yaml
JacksonWeber: Good point, I'll change this to just use the `gen_ai.operation.name` to reduce the cardinality.

model/gen-ai/spans.yaml
JacksonWeber: Looks like the `Response` object returned only carries `instructions`, `output`, and `tools`, but not `input`. Updated the PR accordingly.

model/gen-ai/registry.yaml
JacksonWeber: Looks like this applies to Azure OpenAI and AWS Bedrock as well as both mirror the OpenAI `Response` object.

model/gen-ai/spans.yaml
JacksonWeber: Makes sense. I removed `gen_ai.usage.input_tokens`/`output_tokesn` from the response span entirely.

model/gen-ai/spans.yaml
JacksonWeber: `get_response` fetch will succeed even when the response it returns has a failed original generation, so we don't mark the span as an error in that case. I've added notes to the span and `finish_reasons` attribute to make it more clear.

model/gen-ai/registry.yaml
```diff
@@ -450,6 +450,10 @@ attributes:
           value: "retrieval"
           brief: 'Retrieval operation such as [OpenAI Search Vector Store API](https://platform.openai.com/docs/api-reference/vector-stores/search)'
           stability: development
+        - id: fetch_response
+          value: "fetch_response"
+          brief: 'Fetch a previously generated model response by its identifier, without performing inference, such as [OpenAI Get a model response](https://platform.openai.com/docs/api-reference/responses/get)'
```
lmolkova, line 513:
```suggestion
          brief: 'Fetch a previously generated model response by its identifier, without performing inference, such as [OpenAI Get a model response](https://platform.openai.com/docs/api-reference/responses/get)'
          note: > 
             Instrumentations SHOULD NOT report token usage (as attributes or metrics) for this operation. 
```

Let's move it here so it applies to spans and metrics

model/gen-ai/spans.yaml
```diff
@@ -341,6 +341,67 @@ spans:
...
+      note: |
+        **Span name** SHOULD be `{gen_ai.operation.name}`. Semantic conventions
+        for individual GenAI providers MAY specify a different span name format.
+    brief: >
+      Describes a request to a Generative AI provider that fetches a
+      previously generated model response by its identifier.
+    note: |
+      The `gen_ai.operation.name` SHOULD be `fetch_response`.
+
+      Instrumentations MUST NOT report token usage or metrics for this operation.
```
lmolkova: left a comment suggesting to move it to operation name definition instead of spans

model/gen-ai/spans.yaml
```diff
@@ -341,6 +341,67 @@ spans:
...
+      - ref: gen_ai.operation.name
+        requirement_level: required
+        sampling_relevant: true
+      - ref: server.address
+        sampling_relevant: true
+      - ref: server.port
+        sampling_relevant: true
+      - ref: gen_ai.response.id
+        requirement_level: required
+        sampling_relevant: true
```
lmolkova: it's just  a random string, it does not help with sampling


```suggestion
```

model/gen-ai/spans.yaml
```diff
@@ -341,6 +341,67 @@ spans:
...
+      Describes a request to a Generative AI provider that fetches a
+      previously generated model response by its identifier.
+    note: |
+      The `gen_ai.operation.name` SHOULD be `fetch_response`.
+
+      Instrumentations MUST NOT report token usage or metrics for this operation.
+
+      `error.type` and the span status describe failures of the fetch itself. Successfully
+      fetching a response whose original generation failed or was left incomplete is not an
+      error of this operation; that outcome is conveyed through `gen_ai.response.finish_reasons`.
```
```suggestion
```

it seems it's already captured on the finish reason, not needed here

model/gen-ai/spans.yaml
```diff
@@ -341,6 +341,67 @@ spans:
...
+        requirement_level: opt_in
+      - ref: gen_ai.response.model
+        requirement_level: recommended
+      - ref: gen_ai.response.finish_reasons
+        requirement_level: recommended
+        note: >
+          Conveys the outcome of the original generation as recorded on the fetched response,
+          derived from its status: a completed generation maps to its stop reason, an incomplete
+          one to why it was cut short (for example `length` or `content_filter`), and a failed or
+          cancelled generation to `error`.
```
lmolkova, line 413:
```suggestion
        note: >
          Conveys the outcome of the original generation as recorded on the fetched response,
          derived from its status: a completed generation maps to its stop reason, an incomplete
          one to why it was cut short (for example `length` or `content_filter`), and a failed or
          cancelled generation to `error`.
          
          Semantic conventions for individual GenAI
          providers SHOULD document how to 
          populate this attribute.
```

and can you please add a short sentence as note on openai finish_reason describing how `incomplete_details` map to finish reason?

thanks!

lmolkova (approved):
LGTM, just a few cosmetic suggestions

model/gen-ai/registry.yaml
JacksonWeber: Moved the note over to the registry instead of spans so it'll apply to both spans and metrics.

model/gen-ai/spans.yaml
JacksonWeber: Addressed by the above update.

model/gen-ai/spans.yaml
JacksonWeber: Removed this unnecessary information.

model/gen-ai/spans.yaml
JacksonWeber: Added a note to the finish_reason.

lmolkova (approved):
LGTM

reference/scenarios/openai/scenario.py
```diff
@@ -593,6 +593,115 @@ def run_responses_with_prompt_template_reference(client):
...
+        if service_tier is not None:
+            span.set_attribute("openai.response.service_tier", service_tier)
+        # Content carried on the fetched response: system instructions and output
+        # messages. The original input messages are NOT part of the fetched
+        # response, so `gen_ai.input.messages` is not set here.
+        instructions = getattr(fetched, "instructions", None)
+        if instructions:
+            span.set_attribute(
+                "gen_ai.system_instructions",
+                json.dumps([{"parts": [{"type": "text", "content": instructions}]}]),
```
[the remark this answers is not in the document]
JacksonWeber: Updated to fix.

model/gen-ai/spans.yaml
```diff
@@ -344,6 +344,60 @@ spans:
         requirement_level:
           conditionally_required: When applicable.
 
+  - type: gen_ai.fetch_response.client
```
aabmass, line 347: I think the [Gemini Interactions API](https://ai.google.dev/api/interactions-api#getInteractionById) is equivalent, please let me know if that's right. I think there may be some conflicts with this proposal, I add comments below.

model/gen-ai/registry.yaml
```diff
@@ -459,6 +459,12 @@ attributes:
           value: "retrieval"
           brief: 'Retrieval operation such as [OpenAI Search Vector Store API](https://platform.openai.com/docs/api-reference/vector-stores/search)'
           stability: development
+        - id: fetch_response
```
aabmass, line 511: In the case of the Gemini Interactions API, should we add another value here like `get_interaction`?

We did that for `generate_content` vs `chat`/`text_completion`

reference/scenarios/openai/scenario.py
```diff
@@ -683,6 +683,115 @@ def run_responses_with_prompt_template_reference(client):
...
+            span.set_attribute(
+                "gen_ai.system_instructions",
+                json.dumps([{"type": "text", "content": instructions}]),
+            )
+        output_messages = responses_output_messages(fetched)
+        if output_messages:
+            span.set_attribute("gen_ai.output.messages", json.dumps(output_messages))
+        # Token usage is intentionally NOT recorded on this span: no inference
+        # happens on a fetch, and the fetched response's token counts belong to
+        # the original generation (already accounted for on that operation).
```
aabmass, line 791: What about background execution with polling? IIUC, both OpenAI responses and Gemini interactions support this, and you would typically read the token usage when polling/fetching since the inference is happening in the background.

- https://developers.openai.com/api/docs/guides/background?lang=curl
- https://ai.google.dev/gemini-api/docs/background-execution#rest_3

model/gen-ai/spans.yaml
```diff
@@ -344,6 +344,60 @@ spans:
...
+        requirement_level: required
+        sampling_relevant: true
+      - ref: gen_ai.operation.name
+        requirement_level: required
+        sampling_relevant: true
+      - ref: server.address
+        sampling_relevant: true
+      - ref: server.port
+        sampling_relevant: true
+      - ref: gen_ai.response.id
```
aabmass, line 377: The Interactions API also supports a [`last_event_id`](https://ai.google.dev/api/interactions-api#:~:text=interaction%20to%20retrieve.-,last_event_id,-string) which acts like a cursor. Does the Responses API have something similar?

eternalcuriouslearner (approved):

model/gen-ai/registry.yaml
lmolkova: yes, I think so - different operation name, but the same span type and metric name. Span type will need to eventually land on otlp in some form (regardless of this PR)

reference/scenarios/openai/scenario.py
lmolkova: what would be the alternative? get instrumentation does not know how it's being used. It can be called multiple times on the same content or never called at all.

model/gen-ai/spans.yaml
```diff
@@ -344,6 +344,60 @@ spans:
...
+          Only content carried on the fetched response is recorded. A fetched response contains
+          the system instructions, output messages, and tool definitions but not the original
+          input messages, so `gen_ai.input.messages` is not set on this span.
+      - ref: gen_ai.output.messages
+        requirement_level: opt_in
+      - ref: gen_ai.tool.definitions
+        requirement_level: opt_in
+      - ref: gen_ai.response.model
+        requirement_level: recommended
+      - ref: gen_ai.response.finish_reasons
```
aabmass, line 407: One more note on background generation, there is a separate status for both APIs when polling that indicates if the response is done or not. See https://developers.openai.com/api/docs/guides/background?lang=curl#polling-background-responses

Not sure if capturing it in finish_reason is appropriate, we might need another attribute.

reference/scenarios/openai/scenario.py
aabmass: Ya good point, but how would you recover the token usage if it's not recorded here? I guess it's either double count or don't count at all.

lmolkova: we don't ? 

- we can't record usage metrics here for sure.
- we could record span attributes only assuming people will filter by span type / operation name. But we can't really make any meaningful aggregations on them. Per-response is still valuable. 

A layer above what polls and waits can record token usage. It can be a convenience (start-and-wait-for-lro-to-complete) API that records full inference / agent call or an application. Another option - server side needs to report usage. It's the only reliable place (given network issues and retries, client side is always under-counting).

---
I think @JacksonWeber had them attributes on  span initially and I suggested to remove. I'm open to adding them back with a note that these attributes must not be used for aggregation (and still we get into the same problem as https://github.com/open-telemetry/semantic-conventions-genai/issues/19). My motivation asking to remove was to untangle trivial addition of a span from complex token usage questions.

model/gen-ai/spans.yaml
JacksonWeber: Yeah, I think that's correct. I'll address each of the comments below.

reference/scenarios/openai/scenario.py
aabmass: The issue IMO is that both OpenAI and Gemini use the same RPC to represent "background inference" and "retrieve an older response". @JacksonWeber did you intend to solve both of these with this PR? Maybe we should just propose opening a follow up issue for background/async inference.

> A layer above what polls and waits can record token usage. It can be a convenience (start-and-wait-for-lro-to-complete) API that records full inference / agent call or an application.

I don't think either of the client SDKs have such a layer today.

model/gen-ai/registry.yaml
JacksonWeber: I think that makes sense, I'll work on updating this PR and see what you think.

model/gen-ai/spans.yaml
JacksonWeber: Yeah, RAPI has the equivalent. `GET /v1/responses/{id}` takes `stream: true` with `starting_after` which is basically OpenAI's version of `last_event_id`

model/gen-ai/spans.yaml
JacksonWeber: That's a good point, finish_reasons might be the wrong home for this. Status is meant to capture the lifecycle of the brackground run whereas finish_reasons is why the model stopped once it's procuded output.

I think we want a dedicated attribute here, something like `gen_ai.response.status` with a small cross-provider enum that maps cleanly onto the different statuses. What do you think?

reference/scenarios/openai/scenario.py
JacksonWeber: My intent with this PR was just the "retrieve an existing response by id" case, not to solve background/async inference. I think the fact that both providers overload the same GET for polling an in-progress run is what's been muddying the water here. Happy to keep this PR scoped to just the original intent with a future PR for background inference.

aabmass: Great, do you mind filing a follow up issue for this?

model/gen-ai/spans.yaml
aabmass: Seems relatively harmless to capture it in the span attributes and potentially useful. WDYT?

model/gen-ai/registry.yaml
aabmass: It also looks like the Response API calls this operation "retrieve" https://developers.openai.com/api/reference/resources/responses/methods/retrieve and the Python SDK also uses this. Is there a strong reason to use "fetch"?

JacksonWeber: This was in response to Liudmila's comment: https://github.com/open-telemetry/semantic-conventions-genai/pull/353#discussion_r3525316935. I'm not opposed to using the `retrieve` language here as it also doesn't imply creation. Does that work for you as well, @lmolkova?

reference/scenarios/openai/scenario.py
JacksonWeber: @aabmass I've filed the related issue here: https://github.com/open-telemetry/semantic-conventions-genai/issues/403

model/gen-ai/registry.yaml
lmolkova: we have retrieval operation name which means like vector DB or search call. It'd be confusing to have `retrieve` for response and `retrieval` for data retrieval

lmolkova: maybe the least bad option would be to return to `get_response` ? `fetch` is impossible to mix with create | generate, but `get` (unless you're hardcore REST person) may be misleading. This was my intent behind fetch - it's not used by any provider and the least misleading. I still prefer to keep it, but it's not a strong optinion.

model/gen-ai/spans.yaml
JacksonWeber: Added it here under the name `stream_cursor`. Let me know if that works for you.

model/gen-ai/spans.yaml
JacksonWeber: Happy to add this, what do you think, @lmolkova?

model/gen-ai/spans.yaml
aabmass: Thanks!

model/gen-ai/registry.yaml
aabmass: That makes sense and I don't have a strong opinion, so let's stick with fetch

aabmass (approved):

model/gen-ai/spans.yaml
JacksonWeber: Just noticed the thumbs up, I've added the above proposal.

aabmass: Yup LGTM

aabmass (approved):

model/gen-ai/registry.yaml
```diff
@@ -205,6 +214,45 @@ attributes:
...
+          stability: development
+    brief: The lifecycle status of a generated response, as reported by the provider when the response is fetched or polled.
+    note: >
+      This attribute captures the lifecycle state of a (possibly background or long-running)
+      generation, such as whether it is queued, still running, or has reached a terminal state.
+      It is distinct from `gen_ai.response.finish_reasons`, which describes why the model stopped
+      once it began producing output.
+
+      The value SHOULD be the provider's response status mapped onto the closest member of this
+      enum. Semantic conventions for individual GenAI providers SHOULD document how their status
```
```suggestion
      The value SHOULD be the provider's response status mapped onto the closest member of this
      enum or a provider-specific value when none of these apply. 
      Semantic conventions for individual GenAI providers SHOULD document how their status
```
(please check my english)

Google provider 2 more options: `requires_action` and `budget_exceeded` https://ai.google.dev/api/interactions-api#Resource:Interaction

None of the values defined here are applicable for these.

lmolkova (approved):
LGTM, just a final minor comment. Thanks!

model/gen-ai/registry.yaml
JacksonWeber: Added

PR #354, opened 2026-07-01
Clarify workflow span criteria and add reference scenarios for langchain and openai agents

lmolkova:
- Clarify that invoke_workflow span is reported when framework API implies composite operation
- add examples covering OpenAI Agents Runner.run (explicitly documented as workflow) and similar ADK API
- add recommendation to document per-framework criteria on when to emit workflow vs agent

Fun fact: 

Code like this in OpenAI Agents (and similar on with ADK)

```python
agent_b = Agent(
    name="Agent B",
    instructions="You are Agent B. You just received a transferred request. Answer it politely."
)

agent_a = Agent(
    name="Agent A",
    instructions="You are Agent A. Always transfer the request to Agent B immediately.",
    handoffs=[agent_b]
)

result = await Runner.run(agent_a, "Help me compare metrics and traces.", run_config=RunConfig(workflow_name="sequential-agents"))
```

Produces traces like this:

image

Where agent A and agent B are siblings.

In the future, OpenAI and ADK can document additional heuristics (e.g. if runner.run is called with a single agent, no agents-as-tools, no sub-agents, no handoff agents, etc).

## Motivation

It's not clear today who and when should produce workflow spans. It's also not clear what should be used as a workflow name, resulting in static types like `StateGraph` being used that talk about the implementation details and topology rather than anything about application and its workflows.

## Checklist

See [CONTRIBUTING.md]


RKest (approved):

docs/gen-ai/gen-ai-agent-spans.md
```diff
@@ -691,14 +691,22 @@ Represents an operation that executes a coordinated process composed of multiple
...
+For example, when instrumenting user-facing API that supports composable operations.
+It SHOULD NOT be reported for standalone agent invocations.
+
+Framework-specific semantic conventions SHOULD define which operations to report
+as workflow spans and MAY specify heuristics to distinguish workflows from standalone
+agent invocations.
+
+Examples of workflow invocations in different frameworks include:
+
+- **ADK**: [`Runner.run(...)`](https://adk.dev/agents/) with multi-agent or graph workflow.
```
RKest: Consider adding wording to omit the `invoke_workflow` span, in case it happens to be an implementation detail.

Motivating example is using [AgentTool](https://github.com/google/adk-python/blob/ade8577745bda23588bc0ac71f93a9d3343ec80a/src/google/adk/tools/agent_tool.py#L107) in ADK, which uses a `Runner` internally, which means using `AgentTool` produces the following trace.

```
invoke_workflow                            # Entry point for the user
└── invoke_agent root
    ├── chat gpt4-o
    └── execute_tool subagent
        └── invoke_workflow                # ADK implementation detail
            └── invoke_agent subagent
```

I think the nested `invoke_workflow` here only adds noise given:
- With an exception of an exotic [ADK plugin](https://adk.dev/plugins/) setup) only one agent will even be part of this workflow
- Even if another agent is part of the subagent "workflow" here, this span won't create another trace. This is opposed to if we omitted the root `invoke_workflow` span based on a heuristic the setup "looks like a single agent setup".

My rough draft: "The `invoke_workflow` span SHOULD NOT be reported when the workflow invocation is an internal implementation detail of another operation (e.g., an agent or tool that spins up a runner/workflow under the hood to delegate to a sub-agent) rather than a user-facing entry point. In such cases instrumentations SHOULD suppress the nested `invoke_workflow` span so the trace reflects the caller-facing structure instead of framework internals."

docs/gen-ai/gen-ai-agent-spans.md
```diff
@@ -691,14 +691,22 @@ Represents an operation that executes a coordinated process composed of multiple
...
-This span SHOULD be reported by the instrumentations when they can
-reliably determine that invocation is a workflow (i.e. groups several agent
-invocations) and SHOULD NOT be reported by instrumentations that
-can't distinguish it `invoke_workflow` from `invoke_agent`.
-eg: Some frameworks like ADK have workflow agents that orchestrate other agents
-and report `invoke_agent` spans, so `invoke_workflow` SHOULD NOT be reported by such instrumentations.
-Conversely, frameworks like CrewAI have a distinct concept of crew (similar to workflow)
-that is separate from individual agents, so they SHOULD report `invoke_workflow` spans.
+The workflow span SHOULD be reported for operations executing composable processes
+(e.g., graphs, orchestrators) that can coordinate multiple agents or GenAI calls.
```
RKest: Open question is what should the framework provided orchestrators (e.g. ADK's `Runner`) be named? Either `invoke_workflow`, `invoke_workflow {per_framework_noun}` (e.g. `invoke_workflow runner` in ADK) or `invoke_workflow {first_invoked_agent_name|adk_app_name}`?

One pro of `invoke_workflow {first_invoked_agent_name|adk_app_name}`, is it makes it simple to catch at a glance, by simply looking at the root span, which agentic application is invoked.

Otherwise, `invoke_workflow` or `invoke_workflow {per_framework_noun}` seem cleaner, and probably easier to word in the standard.

RKest (approved):

docs/gen-ai/gen-ai-agent-spans.md
lmolkova: great point, added 

>    If a workflow operation can be invoked both by external callers and internally
      by the instrumented framework, only the external invocation SHOULD be reported as a workflow span.

docs/gen-ai/gen-ai-agent-spans.md
lmolkova: updated workflow name definition in the registry to include

>      For example, it can be the name of the first chain in LangChain,
>      the name of the crew in CrewAI, or the entry point agent in ADK or
>      OpenAI Agents when no explicit workflow name is provided.

>      It is NOT RECOMMENDED to use instrumentation-time constants or names of
>      types representing the workflow, such as "StateGraph". When no meaningful
>      workflow name is available for a framework, this attribute SHOULD NOT be set.

changelog.d/354.clarification.md
```diff
@@ -0,0 +1 @@
+Clarify the reporting criteria for `gen_ai.invoke_workflow` spans, recommend per-framework guidance, add reference scenarios
```
trask: Two notes on the fragment, plus some PR-description feedback I'm parking here since there's no better anchor.

The fragment is a three-item list where the third item — the reference scenarios — isn't a consumer-visible convention change, so I'd drop it and keep this to what actually changed in the conventions. The file is also missing a trailing newline (`.editorconfig` sets `insert_final_newline = true`).

```suggestion
Clarify when instrumentations should report `gen_ai.invoke_workflow` spans, add per-framework examples, and tighten the guidance on what `gen_ai.workflow.name` should contain.
```

On the description: the "Motivation section filled in above" box is unchecked and there's no motivation section. The reasoning in your "Fun fact" section *is* the motivation and deserves to be promoted into the body, especially since this reverses the previous explicit guidance that ADK instrumentations should not report `invoke_workflow`.

The description also only covers OpenAI Agents and ADK, but this PR adds a LangGraph workflow scenario and brand-new langchain `execute_tool` coverage — that's why `reference/README.md` and `reference/reports/execute-tool-span.md` change. Worth calling out so those diffs aren't a surprise.

model/gen-ai/spans.yaml
```diff
@@ -575,14 +575,24 @@ spans:
...
-      eg: Some frameworks like ADK have workflow agents that orchestrate other agents
-      and report `invoke_agent` spans, so `invoke_workflow` SHOULD NOT be reported by such instrumentations.
-      Conversely, frameworks like CrewAI have a distinct concept of crew (similar to workflow)
-      that is separate from individual agents, so they SHOULD report `invoke_workflow` spans.
+      The workflow span SHOULD be reported for operations that trigger the execution
+      of composable processes (e.g., graphs, orchestrators) coordinating multiple
+      agents or GenAI calls. It SHOULD NOT be reported for standalone agent invocations.
+
+      If a workflow operation can be invoked both by external callers and internally
+      by the instrumented framework, only the external invocation SHOULD be reported as a workflow span.
```
trask: I read the thread with @RKest that produced this sentence, so I understand the case it's for: ADK's `AgentTool` spins up a `Runner` internally, giving you a nested `invoke_workflow` underneath `execute_tool`. My concern is that the sentence as written doesn't carry that meaning to someone who hasn't read the thread.

It's framed around *who calls the operation* — "external callers" vs "internally by the instrumented framework" — but the property that actually separates the motivating case is that the nested invocation is an implementation detail of an enclosing GenAI operation. An instrumentation can detect that (there's an active GenAI span in context when the inner `Runner` starts). Detecting whether its caller was "external" is much less obvious. The sentence also doesn't say what becomes of the internal invocation: suppressed entirely, or reported as something else?

@RKest's draft in that thread is closer to implementable and I'd lean toward it, or something like it:

> The `invoke_workflow` span SHOULD NOT be reported when the workflow invocation is an internal implementation detail of another operation (e.g., an agent or tool that spins up a runner/workflow under the hood to delegate to a sub-agent) rather than a user-facing entry point.

I'm deliberately not putting a suggestion block here, because I don't want to guess at whether you also intend this to cover genuinely nested user-authored workflows (a LangGraph sub-graph invoked as a node of a parent graph). Those look identical in the trace but arguably should keep their span. Whichever way you decide, it's worth saying explicitly.

model/gen-ai/spans.yaml
```diff
@@ -575,14 +575,24 @@ spans:
...
+
+      Framework-specific semantic conventions SHOULD define which operations to report
+      as workflows and MAY specify heuristics to distinguish them from standalone
+      agent invocations and internal operations.
+
+      Examples of workflow invocations in different frameworks include:
+
+      - **ADK**: [`Runner.run(...)`](https://adk.dev/agents/) with multi-agent or graph workflow.
+      - **crewai**: [`Crew.kickoff()`](https://docs.crewai.com/concepts/crews)
+      - **langchain / langgraph**: [`*Graph*.invoke(...)`](https://reference.langchain.com/python/langgraph/graphs)
```
trask: Two things in this list.

The ADK link goes to the Agents index. That page does talk about workflows, but nothing on it describes `Runner.run`, which is what the bullet actually names. ADK has a page titled "Workflows: multi-agent, multi-node applications" at <https://adk.dev/workflows/> that matches this bullet much more closely.

The framework names are also inconsistent with each other and with the rest of the model — `spans.yaml:588,590` and `registry.yaml:858` already use `ADK`, `CrewAI`, and `LangChain`. Keeping `ADK` bare (not "Google ADK") stays consistent with those.

```suggestion
      - **ADK**: [`Runner.run(...)`](https://adk.dev/workflows/) with multi-agent or graph workflow.
      - **CrewAI**: [`Crew.kickoff()`](https://docs.crewai.com/concepts/crews)
      - **LangChain / LangGraph**: [`*Graph*.invoke(...)`](https://reference.langchain.com/python/langgraph/graphs)
```

model/gen-ai/spans.yaml
```diff
@@ -575,14 +575,24 @@ spans:
...
+      - **Microsoft Agent Framework**: [`Workflow*.run(...)`](https://learn.microsoft.com/agent-framework/workflows/)
+      - **OpenAI Agents**: [`Runner.run(starting_agent=...)`](https://openai.github.io/openai-agents-python/ref/run/#agents.run.Runner.run)
```
trask: This bullet is unqualified, which contradicts "It SHOULD NOT be reported for standalone agent invocations" three paragraphs above. `Runner.run(some_agent, ...)` with no handoffs, no sub-agents, and no agents-as-tools *is* a standalone agent invocation, but this line says every `Runner.run` is a workflow. The ADK bullet got a qualifier; this one didn't, so instrumentation authors get two different answers for the most common case.

```suggestion
      - **OpenAI Agents**: [`Runner.run(starting_agent=...)`](https://openai.github.io/openai-agents-python/ref/run/#agents.run.Runner.run) with handoffs, sub-agents, or agents-as-tools.
```

If your intent is instead "every `Runner.run` is a workflow because the SDK itself models it as one" — which is how I read the PR description, and the SDK's own docs do describe it as "Run a workflow starting at the given agent" — then it's the general rule above that needs to soften, not this bullet. Something like: a workflow span MAY be reported when the framework's own entry point is defined as a workflow, even when a particular invocation happens to contain a single agent. Right now the two statements can't both hold.

model/gen-ai/registry.yaml
```diff
@@ -846,12 +846,21 @@ attributes:
...
+
+      It is NOT RECOMMENDED to use instrumentation-time constants or names of
+      types representing the workflow, such as "StateGraph". When no meaningful
+      workflow name is available for a framework, this attribute SHOULD NOT be set.
+      
+      This attribute MUST have low cardinality; if there is no low-cardinality workflow name
+      available for a given framework, it MUST NOT be captured by default.
+
+      Semantic conventions for individual Generative AI frameworks SHOULD document 
+      what `gen_ai.workflow.name` means in the context of that framework.
```
trask: Two separate things.

**The two omission rules overlap.** "When no meaningful workflow name is available ... this attribute SHOULD NOT be set" and "if there is no low-cardinality workflow name available ... it MUST NOT be captured by default" aren't the same condition — "meaningful" and "low-cardinality" are independent — but they're adjacent and at different strengths, and the example you give (`StateGraph`) is squarely low-cardinality yet discouraged. So a reader whose only available name is `StateGraph` hits both rules and can't tell which governs. Collapsing them would help:

```suggestion
      This attribute MUST have low cardinality. It is NOT RECOMMENDED to use
      instrumentation-time constants or names of types representing the workflow,
      such as "StateGraph". When no meaningful, low-cardinality workflow name is
      available for a given framework, this attribute MUST NOT be captured by default.

      Semantic conventions for individual Generative AI frameworks SHOULD document
      what `gen_ai.workflow.name` means in the context of that framework.
```

Note this deliberately resolves the `SHOULD NOT` / `MUST NOT` mismatch in favor of `MUST NOT`, which is a strengthening — take it only if that's what you meant.

**Trailing whitespace on lines 859 and 863**, which `.editorconfig` disallows (`trim_trailing_whitespace = true`). Line 863 matters beyond tidiness: this is a `note: |` literal block, so that trailing space is copied verbatim into the generated docs — you can see `SHOULD document ` with a trailing space in the diffs of `docs/gen-ai/gen-ai-agent-spans.md`, `docs/gen-ai/gen-ai-metrics.md`, and `docs/registry/attributes/gen-ai.md`. (Copilot flagged the same class of thing on the whitespace-only blank lines in `spans.yaml`.)

reference/scenarios/langchain/scenario.py
```diff
@@ -175,12 +213,61 @@ def _capture_create(*args, **kwargs):
...
+    }
+    with _reference_tracer.start_as_current_span(
+        f"invoke_workflow {workflow_name}", attributes=workflow_span_attributes
+    ) as workflow_span:
+        workflow_span.set_attribute("gen_ai.workflow.name", workflow_name)
+        workflow_span.set_attribute(
+            "gen_ai.input.messages", json.dumps([{"role": "user", "parts": [{"type": "text", "content": input_text}]}])
+        )
+
+        # Note: Other spans (invoke_agent) are omitted for brevity in this scenario.
```
trask: This comment is inaccurate: it says agent spans are "omitted for brevity", but there are no agents in this graph to omit. `node_a` calls a tool directly and `node_b` formats a string. Worth rewording to describe what the graph actually does.

Separately — and I want to be clear this is a preference, not a defect — I'd push for a more representative graph. Capturability is genuinely fine here: `run_name`, input, and output are all readable at the `graph.ainvoke()` boundary, so this does demonstrate the attributes are collectable, which is the question the rubric actually asks. But the note this PR adds describes workflows as processes "coordinating multiple agents or GenAI calls", and this graph coordinates neither, so as *the* reference example for LangGraph it's thin: a workflow span whose only child is an `execute_tool` span.

`run_plan_and_execute_reference` already shows the pattern for standing up a mock-backed `ChatOpenAI` and emitting the `chat` child span from the captured request/response, and the mock server returns a normal completion for a plain chat call, so having one node do real inference would be cheap. Your call whether that's worth it.

reference/scenarios/openai-agents/scenario.py
```diff
@@ -158,12 +158,64 @@ async def _capture_create(*args, **kwargs):
...
+    agent_a = Agent(
+        name="agent-a",
+        instructions="You are agent A. Handoff to agent-b immediately to answer the user's weather question.",
+        model=model,
+        handoffs=[handoff(agent_b)],
+    )
+    input_text = "What's the weather in Seattle?"
+
+    print("  [workflow_run] agent run as workflow (reference implementation)")
+    workflow_name = "Agent workflow"
```
trask: `"Agent workflow"` is the OpenAI Agents SDK's own default for `RunConfig.workflow_name`, and the SDK docstring for that field says it "Should be a logical name for the run, like 'Code generation workflow' or 'Customer support agent'". Passing the SDK's generic placeholder back in makes for a weak example right next to a note telling instrumentations not to use generic names. The name from your own PR description makes the point much better, and since `data.json` and the reports only track attribute *names*, changing the value won't cause any generated drift:

```suggestion
    workflow_name = "sequential-agents"
```

Also, the module docstring at the top still says the scenario "Exercises: agent run with tool calling against a mock OpenAI server". You updated the langchain docstring for the new workflow run; this one is now stale.

trask (approved):

aabmass (approved):

PR #359, opened 2026-07-03
Add evaluator provenance attributes to gen_ai.evaluation.*

### LINKED ISSUE #79: # Semantic Conventions for GenAI Evaluation organization into Experiments and Test Cases
# RFC: Semantic Conventions for Evaluation Experiments and Test Cases

**Tracking Issue:** [#3398](https://github.com/open-telemetry/semantic-conventions-genai/issues/79)
**Status:** Draft
**Component:** `test`, `gen-ai`

---

## Problem

Building quality agents is hard. Evaluation is rarely binary pass/fail; test cases are subjective, scores are nuanced, and developers need tight feedback loops to debug agent quality and iterate. The `gen_ai.evaluation.result` event ([PR open-telemetry/semantic-conventions#2563](https://github.com/open-telemetry/semantic-conventions/pull/2563)) captures individual evaluation scores on spans, but there's no standard way to:

- Group evaluations into experiment runs
- Identify test cases with machine-readable IDs
- Correlate all spans belonging to a single evaluation execution
- Track evaluations online and offline to close the feedback loop

Without this, observability platforms can't aggregate scores by experiment, compare runs side-by-side, or build experiment tracking UIs. Debugging multi-agent evaluation workflows also becomes significantly harder without standardized conventions. Each framework invents its own schema, fragmenting the ecosystem.

## Evolution of This Proposal

The original issue proposed 13 new `gen_ai.experiment.*` / `gen_ai.test_case.*` attributes at the resource level. After [maintainer feedback](https://github.com/open-telemetry/semantic-conventions-genai/issues/79) pointing to existing `test.*` and `feature_flag.*` conventions, the proposal was revised:

```mermaid
graph LR
    A["Original: 13 new gen_ai.* attrs<br/>Resource-level<br/>GenAI-only"] -->|community feedback| B["Revised: 2 new test.* attrs<br/>Span-level<br/>Generic + reusable"]
```

### What Changed

| Original Proposal | Revised | Why |
|---|---|---|
| `gen_ai.experiment.id` | `test.suite.run.id` (new) | Reuse `test.*` namespace |
| `gen_ai.experiment.name` | `test.suite.name` (exists) | Already in registry |
| `gen_ai.test_case.id` | `test.case.id` (new) | Reuse `test.*` namespace |
| `gen_ai.test_case.name` | `test.case.name` (exists) | Already in registry |
| `gen_ai.experiment.version/tags` | Dropped | Keep lean |
| `gen_ai.test_case.metadata` (JSON) | Dropped | No unstructured blobs |
| `gen_ai.test_case.tags` | Dropped | Keep lean, add later if needed |
| `gen_ai.test_case.input/expected_output` | Deferred | PII concerns, needs separate design |
| `gen_ai.dataset.*` | Deferred | Separate follow-up issue |
| Resource-level placement | Span-level (parent span) | Process may run multiple experiments |

**Net result:** 2 new attributes + 4 existing reused. No new events, spans, or model groups.

---

### New Attributes

Added to `model/test/registry.yaml`:

| Attribute | Type | Stability | Description |
|---|---|---|---|
| `test.suite.run.id` | `string` | `development` | Unique identifier for a specific test suite run. Scope: globally unique (e.g., UUID, timestamp-based ID). |
| `test.case.id` | `string` | `development` | Machine-readable unique identifier for a test case. Scope: unique within a `test.suite.run.id`. Distinct from the human-readable `test.case.name`. |

These are generic, useful for GenAI evaluation, CI/CD test reporting, and ML evaluation alike.

### Why Span-Level, Not Resource-Level

```mermaid
graph TD
    subgraph "❌ Resource-level (original)"
        R["Resource<br/>experiment.id = exp_001<br/>⚠️ immutable for SDK lifetime"]
        R --> S1["invoke_agent"]
        R --> S2["invoke_agent"]
    end

    subgraph "✅ Span-level (revised)"
        P1["test_suite_run span<br/>test.suite.run.id = run_001"]
        P2["test_suite_run span<br/>test.suite.run.id = run_002"]
        P1 --> C1["invoke_agent"]
        P1 --> C2["invoke_agent"]
        P2 --> C3["invoke_agent"]
    end
```

- Resource attributes are immutable per SDK instance, so you can't run multiple experiments sequentially
- Span-level enables natural parent-child correlation via trace context
- Consistent with how `gen_ai.conversation.id` already works on spans

---

## Trace Structure

The `test_suite_run` span shown below is an application-created span with no normative span kind or naming requirements. It is simply the span the application creates to represent the experiment run. The only convention is that it carries `test.suite.*` attributes.

```
Span: test_suite_run
  test.suite.name: "GPT-4 vs Claude Accuracy"
  test.suite.run.id: "run_20250205_001"
  test.suite.run.status: "success"
  │
  ├── Span: invoke_agent [SpanKind: CLIENT]
  │     test.case.id: "case_001"
  │     test.case.name: "greeting_response"
  │     test.case.result.status: "pass"
  │     gen_ai.operation.name: "chat"
  │     gen_ai.request.model: "gpt-4"
  │     │
  │     ├── Event: gen_ai.evaluation.result
  │     │     gen_ai.evaluation.name: "accuracy"
  │     │     gen_ai.evaluation.score.value: 0.95
  │     │
  │     └── Event: gen_ai.evaluation.result
  │           gen_ai.evaluation.name: "tone"
  │           gen_ai.evaluation.score.value: 0.88
  │
  ├── Span: invoke_agent [SpanKind: CLIENT]
  │     test.case.id: "case_002"
  │     test.case.name: "summarization_quality"
  │     test.case.result.status: "pass"
  │     gen_ai.operation.name: "chat"
  │     gen_ai.request.model: "gpt-4"
  │     │
  │     └── Event: gen_ai.evaluation.result
  │           gen_ai.evaluation.name: "accuracy"
  │           gen_ai.evaluation.score.value: 0.87
  │
  └── Span: invoke_agent [SpanKind: CLIENT]
        test.case.id: "case_003"
        test.case.name: "code_generation"
        test.case.result.status: "pass"
        │
        └── Event: gen_ai.evaluation.result
              gen_ai.evaluation.name: "correctness"
              gen_ai.evaluation.score.value: 1.0
```

### Attribute Summary

```mermaid
graph TD
    P["<b>Parent Span: test_suite_run</b><br/>test.suite.name<br/>test.suite.run.id ← NEW<br/>test.suite.run.status"]
    C["<b>Child Spans: invoke_agent</b><br/>test.case.id ← NEW<br/>test.case.name<br/>test.case.result.status<br/>gen_ai.operation.name<br/>gen_ai.request.model"]
    E["<b>Events: gen_ai.evaluation.result</b><br/>gen_ai.evaluation.name<br/>gen_ai.evaluation.score.value<br/>gen_ai.evaluation.score.label"]

    P --> C
    C -.- E
```

---

## What This Enables

```
Experiments
├── GPT-4 vs Claude Accuracy (Feb 5) - 50 cases, 92% pass
│   ├── accuracy: avg 0.95
│   ├── relevance: avg 0.91
│   └── [View details →]
├── RAG Retrieval Test (Feb 4) - 120 cases, 87% pass
└── Tool Calling Test (Feb 3) - 30 cases, 95% pass
```

```
Compare: GPT-4 vs Claude-3
┌────────────────┬─────────┬─────────┐
│ Metric         │ GPT-4   │ Claude  │
├────────────────┼─────────┼─────────┤
│ Pass rate      │ 85%     │ 92% ↑   │
│ Avg accuracy   │ 0.82    │ 0.91 ↑  │
│ Avg latency    │ 2.3s    │ 1.8s ↓  │
└────────────────┴─────────┴─────────┘
```

---

## Instrumentation Flow

The app sets `test.*` attributes on spans as it iterates through the dataset. Evaluators inherit this context from the trace. No attribute copying needed.

### Pattern 1: Inline Evaluation

Evaluator runs in the same process, attaches events directly to the span that already carries `test.case.id`.

```
┌─ eval-runner process ─────────────────────────────────────────┐
│                                                               │
│  Span: test_suite_run                                         │
│    test.suite.run.id: "run_001"                               │
│    │                                                          │
│    └── Span: invoke_agent          ← app sets test.case.id    │
│          test.case.id: "case_001"                             │
│          test.case.name: "greeting"                           │
│          │                                                    │
│          │  1. app calls agent, gets response                 │
│          │  2. evaluator scores response                      │
│          │  3. evaluator attaches event to SAME span          │
│          │                                                    │
│          └── Event: gen_ai.evaluation.result  ← inherits ctx  │
│                gen_ai.evaluation.name: "accuracy"             │
│                gen_ai.evaluation.score.value: 0.95            │
│                                                               │
└───────────────────────────────────────────────────────────────┘
```

### Pattern 2: Post-Hoc Evaluation

Evaluator runs asynchronously (different process or later in time). It creates child spans or uses span links to reference the original trace. Note: in this pattern, the evaluator must independently know the `test.case.id` (e.g., from a database or exported span data). Trace context propagation does not apply across process boundaries in async scenarios.

```
┌─ app process ──────────────────────┐
│                                    │
│  Span: test_suite_run              │
│    test.suite.run.id: "run_001"    │
│    │                               │
│    └── Span: invoke_agent          │
│          test.case.id: "case_001"  │
│          span_id: "abc123"  ───────┼──┐
│                                    │  │
└────────────────────────────────────┘  │
                                        │  span link
┌─ evaluator process ───────────────┐   │
│                                   │   │
│  Span: evaluate                   │   │
│    links: ["abc123"]  ←───────────┼───┘
│    test.case.id: "case_001"       │
│    │                              │
│    └── Event: gen_ai.eval.result  │
│          gen_ai.evaluation.name:  │
│            "accuracy"             │
│          gen_ai.evaluation.       │
│            score.value: 0.95      │
│                                   │
└───────────────────────────────────┘
```

### Pattern 3: Session/Trace-Level Evaluation

Evaluator scores an entire conversation. It creates a span under the `test_suite_run` parent, inheriting `test.suite.run.id` from the trace.

```
Span: test_suite_run
  test.suite.run.id: "run_001"
  │
  ├── Span: invoke_agent (case_001)     ← agent call
  │     test.case.id: "case_001"
  │
  ├── Span: invoke_agent (case_002)     ← agent call
  │     test.case.id: "case_002"
  │
  └── Span: evaluate_session            ← evaluator scores whole session
        test.suite.run.id: "run_001"    ← inherited from parent
        │
        └── Event: gen_ai.evaluation.result
              gen_ai.evaluation.name: "overall_coherence"
              gen_ai.evaluation.score.value: 0.91
```

The key point: because experiment context is on spans, evaluators just need to be part of the same trace (as children or via links). The `test.*` attributes propagate naturally through the trace tree.

---

## CI/CD Integration

These attributes are not GenAI-specific. The same `test.*` conventions apply to CI/CD pipelines running evaluation suites as part of deployment gates or regression checks.

```
┌─ CI Pipeline ─────────────────────────────────────────────┐
│                                                           │
│  Step: Run evaluation suite                               │
│                                                           │
│  Span: test_suite_run                                     │
│    test.suite.name: "pre-deploy-eval"                     │
│    test.suite.run.id: "ci-run-4521"                       │
│    │                                                      │
│    ├── Span: invoke_agent                                 │
│    │     test.case.id: "case_001"                         │
│    │     test.case.result.status: "pass"                  │
│    │     └── Event: gen_ai.evaluation.result              │
│    │                                                      │
│    ├── Span: invoke_agent                                 │
│    │     test.case.id: "case_002"                         │
│    │     test.case.result.status: "fail"                  │
│    │     └── Event: gen_ai.evaluation.result              │
│    │                                                      │
│    test.suite.run.status: "failure"                        │
│                                                           │
│  Step: Gate decision                                      │
│    Query: test.suite.run.status == "failure" → block      │
│                                                           │
└───────────────────────────────────────────────────────────┘
```

Because `test.suite.run.id` correlates all spans in a run, CI systems can:

- Query evaluation results by pipeline run
- Compare scores across deployments (run N vs run N-1)
- Gate deployments on evaluation thresholds
- Track regression trends over time using `test.suite.name` as the stable key

This works the same way for ML model evaluation pipelines, not just GenAI agents.

---

## FAQ

**Does this wrap multi-turn conversations into a single span?**

No. `test.*` attributes are additive metadata on existing spans. For multi-turn conversations, each turn remains its own span; `test.case.id` provides correlation without structural changes. Group turns by `gen_ai.conversation.id` as usual.

```
Span: test_suite_run
  test.suite.run.id: "run_001"
  │
  └── Span: test_case (multi-turn conversation)
        test.case.id: "case_001"
        gen_ai.conversation.id: "conv_abc"
        │
        ├── Span: invoke_agent (turn 1)
        ├── Span: invoke_agent (turn 2)
        └── Span: invoke_agent (turn 3)
              └── Event: gen_ai.evaluation.result
```

**What about backends that don't support event attributes?**

The `test.*` attributes (`test.suite.run.id`, `test.case.id`, etc.) are all span attributes, so they work everywhere. The evaluation scores use the already-merged `gen_ai.evaluation.result` event convention. If a backend doesn't support event attributes well, instrumentations can fall back to recording evaluation data as span attributes instead, since the spec allows both. The core experiment/test case context is always on spans regardless.

**How do evaluators inherit experiment context from the app?**

The app sets `test.suite.run.id` and `test.case.id` on spans as it iterates through the dataset. Evaluators don't copy these attributes; they attach `gen_ai.evaluation.result` events to the same spans (or child spans) that already carry the test context. The attributes propagate naturally through the trace tree.

This works at session, trace, and span level:

```
┌─ App sets test context ───────────────────────────────┐
│                                                       │
│  Span: test_suite_run                                 │
│    test.suite.run.id: "run_001"                       │
│    test.suite.name: "accuracy_eval"                   │
│    │                                                  │
│    └── Span: invoke_agent                             │
│          test.case.id: "case_001"                     │
│          test.case.name: "greeting"                   │
│          gen_ai.conversation.id: "conv_abc"           │
│          │                                            │
│          │  ← evaluator attaches to this span         │
│          │     no attribute copying needed             │
│          │                                            │
│          └── Event: gen_ai.evaluation.result          │
│                gen_ai.evaluation.name: "relevance"    │
│                gen_ai.evaluation.score.value: 0.95    │
│                                                       │
└───────────────────────────────────────────────────────┘
```

For async/post-hoc evaluation, the evaluator creates spans linked to the original via span links, referencing the same `test.case.id`. See the "Instrumentation Flow" section above for detailed patterns.

**Why not use baggage for this?**

Baggage is for propagation context, not analytical attributes. Values in baggage are not automatically recorded as span attributes, so backends can't query or aggregate on them. `test.*` span attributes are directly queryable, indexable, and visible in trace UIs without extra processing.

Mohnish-Srivats:
## Description

This PR proposes adding provenance attributes to the gen_ai.evaluation.result event to track exactly who or what generated an evaluation score, solving the ambiguity when multiple judges evaluate the same agent trajectory. Fixes #79 

The issue itself was created for a different purpose, but we scoped this solve it for provenance which were discussed in the issue comments.

Our proposal comment on #79 : https://github.com/open-telemetry/semantic-conventions-genai/issues/79#issuecomment-4606903052

Acceptance comment on #79 : https://github.com/open-telemetry/semantic-conventions-genai/issues/79#issuecomment-4684088600

Agreement to narrow the scope on #79 : https://github.com/open-telemetry/semantic-conventions-genai/issues/79#issuecomment-4835206352

## Motivation

Currently, if an LLM-as-a-judge and a deterministic regular expression both evaluate the exact same agent trace, they emit indistinguishable gen_ai.evaluation.result events.

**User journey**: This supports the quality evaluation and analytics journeys. Without provenance, downstream dashboards have no standard way to filter scores by evaluator, compare LLM vs Human accuracy, or track the specific version of a prompt used for grading.

**Prior art**: Major evaluation frameworks like DeepEval, LangSmith, and Phoenix all track evaluator identity internally to differentiate between different test variants (e.g., GPT-4 judge vs human labeler)

## Prototype

To prove these attributes are easily implementable in real-world scenarios, I used the  repo ([os-agent-observability-evals](https://github.com/Mohnish-Srivats/os-agent-observability-evals)).
The repo has sample agentic runs and scripts to do different types of evaluations.

The prototype intercepts LangGraph outputs and scores them with DeepEval, mapping the evaluator identity to the new gen_ai.evaluation.evaluator.* attributes in the emitted trace spans.

PR demonstrating the changes and writing the output to a sample JSON file : https://github.com/anirudha/os-agent-observability-evals/pull/2

## Checklist

See [CONTRIBUTING.md]


reference/scenarios/deepeval/scenario.py
```diff
@@ -56,6 +56,11 @@ def _stub_generate(_prompt, schema=None):
             attributes = {
                 "gen_ai.evaluation.name": metric.name,
                 "gen_ai.evaluation.score.value": score,
+                "gen_ai.evaluation.evaluator.id": "deepeval-geval-o1-mini",
+                "gen_ai.evaluation.evaluator.type": "llm_judge",
+                "gen_ai.evaluation.evaluator.version": "1.0",
+                "gen_ai.evaluation.scope": "single_output",
+                "gen_ai.evaluation.reference_set.id": "example-golden-dataset",
             }
```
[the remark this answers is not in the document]
Mohnish-Srivats: Ran the "uv run run-scenario deepeval " command to have the changes reflected in the data.json.

reference/reports/gen-ai-evaluation-result-event.md
```diff
@@ -12,14 +12,19 @@
 
 | Attribute | Supporting Libraries |
 | --- | --- |
+| gen_ai.evaluation.reference_set.id | (none) |
 | gen_ai.evaluation.score.label | [azure-ai-evaluation], [dspy] |
 | gen_ai.evaluation.score.value | [azure-ai-evaluation], [deepeval], [dspy] |
```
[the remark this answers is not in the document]
Mohnish-Srivats: Ran the "uv run run-scenario deepeval " command to have the changes reflected in the data.json.

reference/reports/gen-ai-evaluation-result-event.md
```diff
@@ -12,14 +12,19 @@
...
 
 ## Recommended
 
 | Attribute | Supporting Libraries |
 | --- | --- |
+| gen_ai.evaluation.evaluator.id | (none) |
+| gen_ai.evaluation.evaluator.type | (none) |
+| gen_ai.evaluation.evaluator.version | (none) |
 | gen_ai.evaluation.explanation | [azure-ai-evaluation], [deepeval] |
+| gen_ai.evaluation.scope | (none) |
```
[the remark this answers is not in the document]
Mohnish-Srivats: The docs are now updated with dependency as deepeval after executing the "make generate-all" command to update the docs

changelog.d/79.enhancement.md
```diff
@@ -0,0 +1 @@
+Add evaluator provenance attributes (`evaluator.id`, `evaluator.version`, `evaluator.type`, `reference_set.id`, and `scope`) to the `gen_ai.evaluation.result` event.
```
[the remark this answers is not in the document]
Mohnish-Srivats: Change the name to 79.enhancement.md

model/gen-ai/registry.yaml
```diff
@@ -807,6 +807,53 @@ attributes:
...
+          value: "deterministic"
+          brief: "Evaluation performed by a deterministic script or rule-based system."
+        - id: human
+          value: "human"
+          brief: "Evaluation performed by a human reviewer."
+        - id: custom
+          value: "custom"
+          brief: "Custom or unspecified evaluator type."
+    brief: The type of evaluator.
+    stability: development
```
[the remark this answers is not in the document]
Mohnish-Srivats: Added the stability for enums.

model/gen-ai/registry.yaml
```diff
@@ -807,6 +807,53 @@ attributes:
...
+          value: "single_output"
+          brief: "The evaluation is scoped to a single output."
+        - id: full_run
+          value: "full_run"
+          brief: "The evaluation is scoped to a full agent run or workflow."
+        - id: human_interaction
+          value: "human_interaction"
+          brief: "The evaluation is scoped to a human interaction step."
+    brief: The unit the result is about.
+    stability: development
```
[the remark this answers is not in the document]
Mohnish-Srivats: Added the stability for enums.

model/gen-ai/events.yaml
```diff
@@ -35,7 +35,17 @@ events:
           The unique identifier assigned to the specific
           completion being evaluated. This attribute helps correlate the evaluation
           event with the corresponding operation when span id is not available.
-
+      - ref: gen_ai.evaluation.evaluator.id
+        requirement_level: recommended
+      - ref: gen_ai.evaluation.evaluator.version
+        requirement_level: recommended
+      - ref: gen_ai.evaluation.evaluator.type
```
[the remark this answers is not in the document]
Mohnish-Srivats: made the changes to both azure-ai-eval and dspy

docs/gen-ai/gen-ai-events.md
```diff
@@ -338,19 +338,28 @@ This event captures the result of evaluating GenAI output for quality, accuracy,
...
 | [`gen_ai.evaluation.score.label`](/docs/registry/attributes/gen-ai.md) | ![Development](https://img.shields.io/badge/-development-blue) | `Conditionally Required` If applicable. | string | Human readable label for evaluation. [2] | `relevant`; `not_relevant`; `correct`; `incorrect`; `pass`; `fail` |
 | [`gen_ai.evaluation.score.value`](/docs/registry/attributes/gen-ai.md) | ![Development](https://img.shields.io/badge/-development-blue) | `Conditionally Required` If applicable. | double | The evaluation score returned by the evaluator. | `4.0` |
+| [`gen_ai.evaluation.evaluator.id`](/docs/registry/attributes/gen-ai.md) | ![Development](https://img.shields.io/badge/-development-blue) | `Recommended` | string | Stable identifier for the evaluator definition, not the invocation. Results with different evaluator ids must not be treated as comparable. | `judge-v1`; `accuracy-model-abc` |
```
kylehounslow: Is `evaluator.id` intended to be a hash? If not, this can be unsafe when the definition changes but the id doesn't. This is something I observed when consuming the event in my own benchmarks:  

```
                  evaluator.id      evaluator.version
run A (rubric v1) strands-eval:opus-4-8  2026-06-19
run B (rubric v2) strands-eval:opus-4-8  2026-07-01   ← same id, changed rubric

```
If the consuming query groups by `id`-only, the results will get merged:  
```sql
GROUP BY evaluator.id
```
to keep comparability valid I had to key on `(id, version)` and bump `version` by hand on every rubric change:
```sql
GROUP BY evaluator.id, evaluator.version
```

For clarity, we could state comparability on `(id, version)`, and note `version` MUST change when any score-affecting part of the definition changes (`llm_judge`: model, rubric, scoring config).

docs/gen-ai/gen-ai-events.md
```diff
@@ -362,6 +371,27 @@ event with the corresponding operation when span id is not available.
...
+| `deterministic` | Evaluation performed by a deterministic script or rule-based system. | ![Development](https://img.shields.io/badge/-development-blue) |
+| `human` | Evaluation performed by a human reviewer. | ![Development](https://img.shields.io/badge/-development-blue) |
+| `llm_judge` | Evaluation performed by a Large Language Model. | ![Development](https://img.shields.io/badge/-development-blue) |
+
+---
+
+`gen_ai.evaluation.scope` has the following list of well-known values. If one of them applies, then the respective value MUST be used; otherwise, a custom value MAY be used.
+
+| Value | Description | Stability |
+| --- | --- | --- |
```
kylehounslow: >| `human_interaction` | The evaluation is scoped to a human interaction step. |. 

`single_output` and `full_run` are granularity levels, but `human_interaction` is a kind of step, not a granularity. It seems to sit on a different axis than the other two values. Is it meant as a scope, or is it capturing something `evaluator.type` doesn't?

>| `human` | Evaluation performed by a human reviewer. |

`scope` and `evaluator.type` both carry a "human" value (`human_interaction` vs `human`). My read is they're orthogonal: `evaluator.type` is who produced the score, `scope` is what unit the score is about. If that's right, a note saying so would help.

docs/gen-ai/gen-ai-events.md
Mohnish-Srivats: Thanks for pointing this @kylehounslow .  I think that `evaluator.id` should remain a human-readable identifier across time, but any dashboard queries must group by `(evaluator.id, evaluator.version)` to prevent mixing of any data.

I've updated the spec in `registry.yaml`:
1. Added a note to `evaluator.id` stating that valid comparability requires grouping by both `id` and `version`.
2. Added a rule on the note to `evaluator.version` stating it **MUST** change when any score-affecting parameter (model, rubric, config) is modified. 

Does this wording look good to you, or would you like it tightened up further?

docs/gen-ai/gen-ai-events.md
Mohnish-Srivats: @kylehounslow great call again. Yes, it is meant as a `scope` because it captures something completely independent from `evaluator.type`. 

`evaluator.type = human` means a human is the **Judge** producing the score. 
`scope = human_interaction` means the **Unit being Graded** is a specific step where the agent interacted with a user (e.g., asking for approval, or a thumbs-up button click). 

An `llm_judge` could evaluate a `human_interaction`, or a `human` judge could evaluate a `single_output`. I've added a note to both fields in the spec to explicitly call out that they are independent axes to prevent any confusion!

Mohnish-Srivats:
@lmolkova It was great meeting you at the SIG today!

I wanted to follow up on the suggestion regarding the DeepEval instrumentation prototype to make sure we are aligned on the scope.

In my current prototype ([run_evals.py](https://github.com/anirudha/os-agent-observability-evals/pull/2)), I am manually wrapping DeepEval to prove that the telemetry shape works downstream . After relevancy.measure(test_case) completes, I extract the score and manually emit an OTel event, injecting the new provenance attributes we just standardized (e.g., mapping AnswerRelevancyMetric to evaluator.type = llm_judge and evaluator.id = deepeval-relevancy-gpt4). This successfully validates the Semantic Conventions in a real dashboard.

When you suggested instrumenting DeepEval natively, are you envisioning a dedicated auto-instrumentation package (e.g., opentelemetry-instrumentation-deepeval)?

My main clarification: Is building that auto-instrumentation package a prerequisite for merging this Semantic Conventions PR, or is the current manual-wrapper prototype sufficient proof-of-concept to get the spec merged first?

Mohnish-Srivats:
@kylehounslow could you please review the changes and let me know if its ready for approval? If yes, could you please approve the PR if things look good on your side!!

kylehounslow (approved):

reference/scenarios/azure-ai-evaluation/scenario.py
```diff
@@ -43,6 +43,9 @@ def run_evaluation():
                     "gen_ai.evaluation.name": evaluation_name,
                     "gen_ai.evaluation.score.label": score_label,
                     "gen_ai.evaluation.score.value": score,
+                    "gen_ai.evaluation.evaluator.id": f"azure-{evaluation_name.lower()}-{model_config.get('model', 'unknown')}",
+                    "gen_ai.evaluation.evaluator.type": "llm_judge",
```
lmolkova: can you please show how/if this information can be obtained in instrumentation? Is there an evaluator type that instrumentation can inspect to know it's llm-as-a-judjge or some other property of evaluator?

reference/scenarios/azure-ai-evaluation/scenario.py
```diff
@@ -43,6 +43,9 @@ def run_evaluation():
...
+                    "gen_ai.evaluation.scope": "single_output",
```
lmolkova: similarly, is this something that generic instrumentation would know? Is it a property of dataset it receives?

lmolkova (changes requested):
I asked AI to do research on azure-ai-evals, deep-evals, dspy (the ones we're adding scenarios for)  to check which of these attributes we could capture with generic instrumentations


| Attribute | deepeval | dspy | azure-ai-evaluation |
|---|---|---|---|
| `evaluator.id` | **Constructed, not native.** Metric has `.name`; build id from name+model: `f"{metric.__name__}-{metric.evaluation_model}"` | **Constructed.** Metric is a callable → `metric.__name__` | **Constructed.** Evaluator is a class → `type(ev).__name__` + `model_config["model"]` |
| `evaluator.version` | **Not native** — no version concept. Scenario hardcodes `"1.0"` | **Not native.** Function has no version | **Not native** (not claimed in report). Would be hardcoded |
| `evaluator.type` | **Inferable.** Native metrics use an LLM (`using_native_model`/`evaluation_model` set) → `llm_judge`; custom `BaseMetric` w/o model → `deterministic` | **Weak.** Metric is just a callable; DSPy can't tell if it calls an LLM. Scenario assumes `deterministic` | **Inferable from class:** AI-assisted (`RelevanceEvaluator`…) → `llm_judge`; NLP/math (`F1ScoreEvaluator`, `BleuScoreEvaluator`, `RougeScoreEvaluator`) → `deterministic` |
| `reference_set.id` | **Available.** `EvaluationDataset`/`Golden` has `dataset.alias` → id | **Not native.** `devset` is a `list[Example]` with no id/alias | **Weak.** Runs over a jsonl dataset; no per-evaluator id, only a run/data path you'd pass in |
| `scope` | **Inferable from test-case type:** `LLMTestCase` → `single_output`; `ConversationalTestCase` → multi-turn (no clean enum match) | **`single_output`** — metric runs per `Example` (pred vs gold) | **Inferable from evaluator:** single-turn → `single_output`; agent evaluators (`IntentResolutionEvaluator`, `ToolCallAccuracyEvaluator`) → whole trajectory ≈ `full_run` |

Key takeaways:

- **`evaluator.version` is not intrinsic to any of the three libraries** — every scenario hardcodes it. That's the weakest capturability claim.
- **`evaluator.id` is always synthesized** from name+model; no library exposes a native stable id.
- **`scope`**: the libraries *do* know the unit, but via their own type system (test-case type / evaluator class), not a "single vs full run" flag. The enum maps imperfectly (e.g., deepeval conversational, azure agent-level).
- **Terminology mismatch**: none of these libs use the words "evaluator provenance," "reference set," or "scope." They say *metric* (deepeval/dspy) or *evaluator* (azure), *golden/dataset* (deepeval) or *devset* (dspy), and there's no "scope" concept — it's implied by the test-case/evaluator class.

-----

If the intent is to document conventions for eval-as-a-service, let's do a research on what popular offerings return in the results (or allow to configure). Here's a research on a random set of eval-as-a-service

| Attribute name | LangSmith mapping | Patronus AI | Langfuse | OpenAI Evals | Braintrust |
|---|---|---|---|---|---|
| `gen_ai.evaluation.evaluator.id` | [Evaluator resource `id`](https://docs.langchain.com/langsmith/manage-evaluators-sdk#retrieve-an-evaluator). ⚠️ Separate workspace resource — **not present on the feedback/score object**; linked only via run rules | [`evaluator_id` / `evaluator_profile_public_id`](https://docs.patronus.ai/docs/api_ref/evaluations/evaluate_v1_evaluate_post), returned **directly on each `evaluation_result`**; configure [custom Judge evaluators](https://docs.patronus.ai/docs/evaluators/judge-ui) | [Evaluator resource ID](https://langfuse.com/docs/evaluation/evaluation-methods/llm-as-a-judge#programmatic-setup-via-api). ⚠️ **Not present on the [Score object](https://langfuse.com/docs/evaluation/scores/data-model)** (only `name`/`configId`) | [`testing_criteria[].id`](https://developers.openai.com/api/docs/guides/evals#create-an-eval). ⚠️ Run results are aggregate pass/fail **per criterion**, not per-item | [Scorer Function `id`](https://www.braintrust.dev/docs/api-reference/functions/get-function), referenced as [`scorer_function_id`](https://www.braintrust.dev/docs/evaluate/score-online). ⚠️ Lives on the rule/scoring span, not on the numeric score |
| `gen_ai.evaluation.evaluator.version` | **Partial:** prompt [`commit_hash_or_tag`](https://docs.langchain.com/langsmith/manage-evaluators-sdk) only; other runtime settings unversioned; code evaluators have no version | **Partial:** [`criteria_revision`](https://docs.patronus.ai/docs/api_ref/evaluations/evaluate_v1_evaluate_post) plus date-versioned [`evaluator_id` aliases](https://docs.patronus.ai/docs/evaluators/reference_guide) | [Evaluators are versioned](https://langfuse.com/docs/evaluation/evaluation-methods/llm-as-a-judge#programmatic-setup-via-api) (new version per update). ⚠️ Version not recorded on the Score | Not Applicable; grader definition is inline in [`testing_criteria`](https://developers.openai.com/api/docs/guides/evals#create-an-eval) | **Unconfirmed:** Braintrust versions functions via `_xact_id`, but the [Functions API page](https://www.braintrust.dev/docs/api-reference/functions/get-function) doesn't explicitly expose a `version`/`_xact_id` field |
| `gen_ai.evaluation.evaluator.type` | Evaluator [`type`: `llm` or `code`](https://docs.langchain.com/langsmith/manage-evaluators-sdk#retrieve-an-evaluator) | [`evaluator_family` / `evaluation_type`](https://docs.patronus.ai/docs/api_ref/evaluations/evaluate_v1_evaluate_post); bucket via the [evaluator catalog](https://docs.patronus.ai/docs/evaluators/reference_guide) (e.g. Judge/GLIDER → `llm_judge`, NLP/exact-match → `deterministic`) | Evaluator kind: [LLM-as-a-Judge](https://langfuse.com/docs/evaluation/evaluation-methods/llm-as-a-judge) or [code evaluator](https://langfuse.com/docs/evaluation/evaluation-methods/code-evaluators) | Grader [`type`](https://developers.openai.com/api/docs/guides/graders): `score_model` (→ `llm_judge`), `string_check` / `text_similarity` / `python` (→ `deterministic`) | Scorer kind: [LLM judge, Autoeval, or custom code](https://www.braintrust.dev/docs/evaluate/write-scorers) |
| `gen_ai.evaluation.reference_set.id` | [Dataset `id`](https://docs.langchain.com/langsmith/example-data-format) when references come from a dataset | [`dataset_id`](https://docs.patronus.ai/docs/api_ref/evaluations/evaluate_v1_evaluate_post) is the set id. ⚠️ **Self-reported metadata** (existence not guaranteed); the [`gold_answer`](https://docs.patronus.ai/docs/datasets/using) field is per-sample reference *content*, not the set id | [Dataset resource ID](https://langfuse.com/docs/evaluation/experiments/datasets); reference values mapped as [evaluator inputs](https://langfuse.com/docs/evaluation/evaluation-methods/llm-as-a-judge#map-variables--preview-evaluation-prompt) | **Partial:** file-backed [`data_source` file id](https://developers.openai.com/api/docs/guides/evals); no dedicated reference-set id | Dataset id when scorer [`expected`](https://www.braintrust.dev/docs/evaluate/write-scorers) comes from a dataset |
| `gen_ai.evaluation.scope` | Inferred from [evaluation target](https://docs.langchain.com/langsmith/evaluation-concepts): leaf run → `single_output`; root workflow → `full_run`; [thread-level](https://docs.langchain.com/langsmith/online-evaluations-multi-turn) has **no matching enum value** | One [evaluation request/sample](https://docs.patronus.ai/docs/api_ref/evaluations/evaluate_v1_evaluate_post) → implicit `single_output`; no explicit field | [Observation → `single_output`; Trace → `full_run`](https://langfuse.com/docs/evaluation/evaluation-methods/llm-as-a-judge#understanding-each-evaluation-target); [Session / DatasetRun](https://langfuse.com/docs/evaluation/scores/data-model) have **no matching enum value** | One [output item](https://developers.openai.com/api/docs/guides/evals) → implicit `single_output`; no explicit field | Rule [Span → `single_output`; Trace → `full_run`](https://www.braintrust.dev/docs/evaluate/score-online) |

> ⚠️ **Provenance co-located with the result?** Only **Patronus** returns evaluator id/version/family/dataset directly on each result. **LangSmith, Langfuse, and Braintrust** keep evaluator identity/version in a *separate* resource that isn't on the score object — so populating `evaluator.id`/`evaluator.version` on the per-result event requires a join the emitter may not have at result time.
>
> ⚠️ **OpenAI Evals is deprecated:** read-only from **2026-10-31**, shutdown **2026-11-30** ([graders](https://developers.openai.com/api/docs/guides/graders) and [evals](https://developers.openai.com/api/docs/guides/evals) pages).

Sharing this not as a real research result, but as an example of what I'm looking for:

- are things being proposed apply to popular systems
- are they capturable, how and in which cases
-  does terminology align

what I see is that the following things seem to align:
- evaluator.id 
- evaluator type | kind
- reference set is usually called dataset, otherwise seems to align

what doesn't:
- version - only langfuse has clear version
- scope - the weakest - not reliable to infer, called differently (target)

Mohnish-Srivats:
@lmolkova I now clearly understood the expectations here. I will read through your comments, and try to get it sorted in our next SIG meeting.

Thanks to @AgentGymLeader for clarifying in the comments.

mukundheda:
We built this pattern independently (an app-level tag, `augmentloop.grade.source`, on `gen_ai.evaluation.result`) and it maps onto your `evaluator.type` enum: our `math` (a deterministic checker compares against a provably-correct answer) is your `deterministic`; our `ai_judge` is your `llm_judge`. We never let `ai_judge` grades into anything that feeds a headline "correct" number, filtered by source rather than by convention.

One gap: our third source, `reality`, has no slot. It's a grade that arrives asynchronously, after the decision's span has closed, from an actual outcome rather than a check or an opinion: the clip was actually kept, the appointment actually landed. Not `deterministic` (nothing computes it at eval time), not `llm_judge` (no model ever scores it). We handle it by re-emitting the event with a span link back to the original decision span, plus the same `gen_ai.response.id` as correlation fallback, so a late verdict stays one hop from the decision it judges. Might be worth a fifth value, or a line in the spec noting that resolution timing is orthogonal to evaluator type: `reality` could pair with any of the current four.

On @lmolkova's capturability table: it matches what we found. None of our three sources were auto-detected. The application tags grade source at the point it emits the event, because that's the only place that reliably knows which path produced the number, whether a checker, an async outcome, or a judge call. We didn't infer it from library internals. Given the table's own finding, that `evaluator.id`/`type` are constructed everywhere and native nowhere, that reads less like a workaround and more like the actual answer: tagging at emission is the right layer. Cost was small: one mandatory string field, set by the code that already knows its own kind. No inference logic, no per-library adapter.

Writeup, if useful:

- [`docs/conventions.md` section 3](https://github.com/mukundheda/augmentloop-signoz-hackathon/blob/main/docs/conventions.md#3-our-two-mandatory-extension-fields-adr-0002) (the two mandatory extension fields) and [section 6](https://github.com/mukundheda/augmentloop-signoz-hackathon/blob/main/docs/conventions.md#6-span-link-role-1-deferred-grades-link-back-mandatory) (how the deferred `reality` grade links back)
- [ADR 0001](https://github.com/mukundheda/augmentloop-signoz-hackathon/blob/main/docs/adr/0001-machine-checked-grades-only-in-the-headline-metric.md): why the headline metric counts machine-checked grades only
- [ADR 0002](https://github.com/mukundheda/augmentloop-signoz-hackathon/blob/main/docs/adr/0002-extend-the-standard-eval-event-do-not-deviate.md): extend the standard event, don't deviate from it

Mohnish-Srivats:
@mukundheda thanks for your suggestions, and I really appreciate your time to write this up. I have some different thoughts which we can discuss in detail!!

**Reality/Grader type**
I think we might be confusing agent evaluations and business outcomes here.

The core purpose of gen_ai/agent evaluations is to extract behavioral rules from agent prompts ,system instructions etc.., and then use these specifications to automatically evaluate traces for compliance, policy checks, groundedness, etc, with a rule based flow, ai_judge etc..

In your example, we are grading whether the agent correctly suggested an appointment time from an available slot rather than a booked slot, or whether it gave a totally irrelevant answer. What the user ultimately clicks (or whether the appointment lands) is totally out of scope for the evaluation of the agent itself. That is a business outcome, which is something the agent cannot control. Even a perfectly compliant and behaviorally correct agent response might not result in a conversion if the user simply changes their mind. So the outcome cannot be contolled, even if the agent did everything right. All we care is about, whether the agent could give the right response and follow all instructions.

Because of this, I believe tracking "reality" or asynchronous business outcomes belongs in standard product analytics or a separate User Feedback semantic convention, rather than being shoehorned into gen_ai.evaluation.result as a new evaluator type. Keeping this boundary clean ensures our evaluation metrics strictly measure agent behavior and policy compliance.

Mohnish-Srivats:
Hi @lmolkova 

I've instrumented as much as I could here to automatically capture the provenance fields!  I wanted to showcase that we actually do have a technical way to construct evaluator.type and scope dynamically at the library level if we choose to go that route. For fields like version, its a gap and there isn't a way to construct it. However, we would still need version. For example the system prompt to an llm_judge could be updated which is a new version of that llm_judge.

In the Azure scenario, I used some built in modules to dynamically figure out the type and scope from the lib:
`
is_llm_judge = hasattr(evaluator, "_model_config")
attributes["gen_ai.evaluation.evaluator.type"] = "llm_judge" if is_llm_judge else "deterministic"


sig = inspect.signature(evaluator.__call__)
if "conversation" in sig.parameters or "messages" in sig.parameters:
    attributes["gen_ai.evaluation.scope"] = "full_run"
else:
    attributes["gen_ai.evaluation.scope"] = "single_output"
`
This auto-detects if it's an LLM judge by checking for internal model configs and auto-detect's scope by checking the evaluator's function signature

I applied a similar introspection pattern for DeepEval (getattr(metric, "model", None)) and DSPy to automatically enrich the telemetry without requiring any manual developer tagging.

**Limitations to consider**: 

Brittleness: It relies on internal, undocumented library properties (like _model_config), meaning a minor update to Azure/DeepEval property like model_config could silently break the telemetry collection.

Incomplete Coverage: Library introspection can only guess llm_judge or deterministic. It's impossible for an auto-instrumentation script to detect if a human or a custom business process generated the grade. But if they use a library for evaluations, this behavior is expected.

I've pushed up these changes so we have the auto-inference reference implementation available. I'm happy to keep it this way , or strip it out and treat these fields as explicit capture gaps (app-level tagging) based on what you think is best for the spec!

AgentGymLeader:
@Mohnish-Srivats Good progress — evaluator.id now falls out of the real GEval and GPTModel, and dropping evaluator.version instead of inventing one is the right call. The azure signature check is a nice touch.

One thing I'd look at again: reference_set.id.

    dataset = EvaluationDataset()
    dataset.alias = "example-golden-dataset"
    ...
    if getattr(dataset, "alias", None):
        attributes["..."] = dataset.alias

That dataset is constructed empty and never reaches metric.measure(test_case) — it appears four times in the file and none of them touch the evaluation. It's the same literal from before, read back from an object that isn't part of the run. It now looks derived, which makes it harder to spot than when it was written inline.

I don't think that's a mistake to fix by finding a better source — I think it's your answer. Having to build a detached EvaluationDataset to produce the value is good evidence that DeepEval's evaluation path doesn't carry dataset identity. That's worth saying on the attribute itself.

Two smaller things:

evaluator.id does f"...{metric.model.model}" unguarded, and the evaluator.type check right below it does getattr(metric, "model", None) — so those two lines disagree about whether model can be absent.

type(test_case).__name__ == "LLMTestCase" — LLMTestCase is imported above, so isinstance works and survives subclasses.

Also, dspy lost evaluator.type: "deterministic". That one was true — contains_mock_response really is a plain function.

mukundheda:
@Mohnish-Srivats Fair point on keeping the boundary clean, and I'm not asking to fold business outcomes into the agent behavior check. `evaluator.type` answers who produced the grade rather than what got graded, and a reality grade never claims to score compliance, it just needs a place that isn't `custom` because `custom` right now carries two opposite meanings, this came from an observed outcome and nobody classified this.

If reality belongs in a separate outcome or user feedback convention entirely that works for me too, but `evaluator.type` still can't say this value didn't come from a check or a judge without also saying we never bothered to find out, and that gap holds regardless of where reality itself ends up living.

Mohnish-Srivats:
@AgentGymLeader thanks for your suggestions. I have made the changes and pushed it as a commit

Mohnish-Srivats:
@mukundheda as you rightly it should be part of a separate span attribute and not on the eval provenance. They type should only say who produced the grade.

Mohnish-Srivats:
Hello @lmolkova , I would like your feedback on the changes I made on the instrumentation, and I have made sure not to hard code any values . The table you gave was very much helpful. Its now ready for your review and approval when you have a moment!"

reference/scenarios/dspy/scenario.py
```diff
@@ -64,7 +64,12 @@ def run_evaluation():
                 "gen_ai.evaluation.name": contains_mock_response.__name__,
                 "gen_ai.evaluation.score.label": score_label,
                 "gen_ai.evaluation.score.value": score,
+                "gen_ai.evaluation.evaluator.id": f"dspy-{contains_mock_response.__name__.lower()}",
             }
+
+            if type(evaluate).__name__ == "Evaluate":
+                attributes["gen_ai.evaluation.scope"] = "single_output"
+            attributes["gen_ai.evaluation.evaluator.type"] = "deterministic"
```
lmolkova: Assuming user gives a custom metric, there is no obvious way to tell if it's deterministic or not - e.g. https://dspy.ai/getting-started/metrics/

reference/scenarios/dspy/scenario.py
```diff
@@ -64,7 +64,12 @@ def run_evaluation():
                 "gen_ai.evaluation.name": contains_mock_response.__name__,
                 "gen_ai.evaluation.score.label": score_label,
                 "gen_ai.evaluation.score.value": score,
+                "gen_ai.evaluation.evaluator.id": f"dspy-{contains_mock_response.__name__.lower()}",
             }
+
+            if type(evaluate).__name__ == "Evaluate":
+                attributes["gen_ai.evaluation.scope"] = "single_output"
```
lmolkova: this seems like a weird heuristic to use why method being Evaluate means single_output ?

reference/scenarios/deepeval/scenario.py
```diff
@@ -53,10 +53,24 @@ def _stub_generate(_prompt, schema=None):
     with _reference_tracer.start_as_current_span("reference.evaluation", kind=SpanKind.INTERNAL) as span:
         try:
             score = float(metric.measure(test_case))
+
+            evaluator_model = getattr(metric, "model", None)
+            evaluator_model_name = evaluator_model.model if evaluator_model else "unknown"
             attributes = {
                 "gen_ai.evaluation.name": metric.name,
                 "gen_ai.evaluation.score.value": score,
+                "gen_ai.evaluation.evaluator.id": f"deepeval-{type(metric).__name__.lower()}-{evaluator_model_name}",
```
lmolkova: what is the point in creating synthetic id from parts that are / can be reported individually  (model, library, metric name) ?

reference/scenarios/deepeval/scenario.py
```diff
@@ -53,10 +53,24 @@ def _stub_generate(_prompt, schema=None):
...
             }
+
+            if getattr(metric, "model", None) is not None or getattr(metric, "evaluation_model", None) is not None:
+                attributes["gen_ai.evaluation.evaluator.type"] = "llm_judge"
```
lmolkova, line 63: assuming we record model name on eval result, would we still need to report llm_judge as an attribute? presence of model is indicative on its own.

reference/scenarios/deepeval/scenario.py
```diff
@@ -53,10 +53,24 @@ def _stub_generate(_prompt, schema=None):
...
+            else:
+                attributes["gen_ai.evaluation.evaluator.type"] = "deterministic"
+
+            if isinstance(test_case, LLMTestCase):
+                attributes["gen_ai.evaluation.scope"] = "single_output"
+            elif isinstance(test_case, ConversationalTestCase):
+                attributes["gen_ai.evaluation.scope"] = "full_run"
```
lmolkova: it seems deepeval considers then synonyms

> Throughout this documentation, you should assume the term 'test case' refers to an LLMTestCase instead of MLLMImage or ConversationalTestCase.

and they both are single turn test cases as opposed to multi-turn test cases

https://deepeval.com/docs/evaluation-test-cases

reference/scenarios/azure-ai-evaluation/scenario.py
```diff
@@ -34,16 +35,27 @@ def run_evaluation():
             result = evaluator(query=query, response=response)
             score = float(result["relevance"])
             score_label = str(result["relevance_result"])
+            attributes = {
+                "gen_ai.evaluation.explanation": str(result["relevance_reason"]),
+                "gen_ai.evaluation.name": evaluation_name,
+                "gen_ai.evaluation.score.label": score_label,
+                "gen_ai.evaluation.score.value": score,
+                "gen_ai.evaluation.evaluator.id": f"azure-{evaluation_name.lower()}-{model_config.get('model', 'unknown')}",
```
lmolkova: same here: what benefit does this attribute bring on top of recording evaluation name, lib name and model as individual properties?

reference/reports/gen-ai-evaluation-result-event.md
```diff
@@ -12,14 +12,19 @@
 
 | Attribute | Supporting Libraries |
 | --- | --- |
+| gen_ai.evaluation.reference_set.id | (none) |
```
lmolkova, line 15: it seems we never record it in reference instrumentations, when would it be recorded?

reference/reports/gen-ai-evaluation-result-event.md
```diff
@@ -12,14 +12,19 @@
...
 | gen_ai.evaluation.score.label | [azure-ai-evaluation], [dspy] |
 | gen_ai.evaluation.score.value | [azure-ai-evaluation], [deepeval], [dspy] |
 
 ## Recommended
 
 | Attribute | Supporting Libraries |
 | --- | --- |
+| gen_ai.evaluation.evaluator.id | [azure-ai-evaluation], [deepeval], [dspy] |
+| gen_ai.evaluation.evaluator.type | [azure-ai-evaluation], [deepeval], [dspy] |
+| gen_ai.evaluation.evaluator.version | (none) |
```
lmolkova, line 25: it seems it's not reported - how would it be recorded?

maReins:
@lmolkova  Thanks for the detailed review! I went back through all these comments together, and I think they exposed one common problem in the previous revision: the reference scenarios were trying too hard to populate every proposed provenance field, which led us to infer or synthesize values the underlying evaluator did not actually expose.

I've pushed a revision that addresses that consistently:

* **Removed `gen_ai.evaluation.scope` from this PR.** The DSPy and DeepEval mappings were heuristic, and scope is separable from evaluator provenance.
* **Stopped synthesizing `evaluator.id`.** DeepEval and DSPy now leave it unset because they do not expose a stable evaluator identifier in these scenarios. For Azure, we now use the actual framework-provided `RelevanceEvaluator.id` (`azureai://built-in/evaluators/relevance`) rather than constructing an ID from the library, evaluation name, and model.
* **Retained `evaluator.type`, but tightened its semantics.** It now describes the mechanism that determines the evaluation result, rather than simply whether a model is present. This is important because model presence is not sufficient to distinguish evaluator mechanisms. The reference scenarios now set the type only where the mechanism is actually known: Azure `RelevanceEvaluator` and DeepEval `GEval` are `llm_judge`; the locally defined DSPy string-check metric is `deterministic`.
* **Clarified `evaluator.version` and `reference_set.id`.** Instrumentation should use stable values supplied by the evaluation framework/application when available and omit them otherwise, rather than manufacture provenance. `reference_set.id` is conditionally required when a reference set was used and its stable identifier is available.

The reference report now reflects the actual capture support rather than synthetic coverage: Azure provides a real `evaluator.id`, all three scenarios demonstrate `evaluator.type`, and unsupported provenance fields remain unset. All three reference scenarios pass the conformance runner, and the policy checks pass locally.

I think this addresses the common concern behind the review comments while preserving the provenance distinctions the proposal is intended to standardize.

Would you be willing to take another look at the revised PR? Let us know if there's anything else you consider blocking for the provenance attributes themselves.

reference/reports/gen-ai-evaluation-result-event.md
maReins: You're right, none of these reference scenarios currently expose a stable reference-set identifier on the evaluation path, so this is an intentional capture gap rather than something we should synthesize.

The revised PR leaves it unset in those cases. The attribute would be populated when the evaluation framework or application supplies a stable identifier for the reference/dataset actually used by that evaluation.

reference/reports/gen-ai-evaluation-result-event.md
maReins: Agreed, none of the current reference libraries expose a stable evaluator version that generic instrumentation can capture here. We previously hardcoded/inferred values in the scenarios, but have removed that.

So for these reference scenarios this is an intentional capture gap. evaluator.version would only be emitted when the framework or application provides a real stable version value; otherwise it remains unset.

maReins:
@lmolkova  I've replied on the two remaining reference-report threads. For both evaluator.version and reference_set.id, the (none) entries are intentional capture gaps in these reference scenarios rather than missing implementation. The revised scenarios only emit provenance that the framework/application can actually supply; they no longer hardcode or synthesize unsupported values.

I think that resolves the remaining questions, but happy to discuss at the next SIG as well.

docs/gen-ai/gen-ai-events.md
```diff
@@ -411,18 +411,32 @@ This event captures the result of evaluating GenAI output for quality, accuracy,
...
 | [`error.type`](https://github.com/open-telemetry/semantic-conventions/blob/v1.44.0/docs/registry/attributes/error.md) | ![Stable](https://img.shields.io/badge/-stable-lightgreen) | `Conditionally Required` If the operation ended in an error. | string | Describes a class of error the operation ended with. [1] | `timeout`; `java.net.UnknownHostException`; `server_certificate_invalid`; `500` |
-| [`gen_ai.evaluation.score.label`](/docs/registry/attributes/gen-ai.md) | ![Development](https://img.shields.io/badge/-development-blue) | `Conditionally Required` If applicable. | string | Human readable label for evaluation. [2] | `relevant`; `not_relevant`; `correct`; `incorrect`; `pass`; `fail` |
+| [`gen_ai.evaluation.reference_set.id`](/docs/registry/attributes/gen-ai.md) | ![Development](https://img.shields.io/badge/-development-blue) | `Conditionally Required` [2] | string | Stable identifier for the reference set against which the evaluation result was produced. [3] | `dataset-123`; `golden-set-v1` |
```
singankit, line 414: Is `reference set` is ground truth dataset ? And when is the the condition when it is required ?

maReins: Hi @singankit ! Good question: yes, we're thinking of a reference set is the ground-truth/golden/reference dataset used by the evaluation, when the evaluation relies on that kind of set.

The attribute is conditionally required when a reference set was actually used to produce the evaluation result and a stable identifier for that set is available from the framework or application. If the evaluation does not use a reference set, or no stable identifier is available, it should be taken out rather than synthesized.

I'm going to clarify the requirement text to make that condition explicit.

Mohnish-Srivats:
Hello @lmolkova , I have addressed most of the comments a few weeks back, and I would really appreciate your feedback in this PR.If there is something you feel thats missing I would like to address that.

We have framed the instrumentation according to the guidelines that @trask has shared previously in the SIG meeting, and please let me know if some fields still need an update.
Thanks!!

Issue #56, opened 2025-10-15
need more guidance on what gen AI instrumentations should do when a request fails


DylanRussell:
### Area(s)

area:gen-ai

### What's missing?

Should inputs, outputs, system instructions be added to the log/span, and should the upload hook be called in this scenario ?

AFAICT the current guidance is just to add the `error.type` attribute

### Tip

<sub>[React](https://github.blog/news-insights/product-news/add-reactions-to-pull-requests-issues-and-comments/) with 👍 to help prioritize this issue. Please use comments to provide useful context, avoiding `+1` or `me too`, to help us triage it. Learn more [here](https://opentelemetry.io/community/end-user/issue-participation/).</sub>

lmolkova:
Failure should not affect input messages and instructions. 

I can see a few scenarios when it comes to output messages: 

| Streaming  | Partial response (e.g. half of the stream) or no response | How to record output | 
| ---- | ---- | ---- | 
| No |  There is a full output (success) or nothing (failure) |  no output, record `finish_reasons` = error, `error.type` + span status | 
| Yes | Partial output is available | record partial  - e.g. truncated string.  `error.type` attribute + span status  would explain that output is truncated. `finish_reasons` should also be error |  
| Yes | No partial output is available |  no output, record `error.type` + span status | 

I agree we need to document hook behavior (call always but without output / with partial one).

WDYT?

aabmass:
It feels a little weird to me to set finish_reason to error in certain cases. When the LLM inference isn't even started (e.g. quota exhausted, invalid argument, permission denied) I think it should remain unset. A simple rule would be to only record finish reason if one was returned from the API?

lmolkova:
> It feels a little weird to me to set finish_reason to error in certain cases. When the LLM inference isn't even started (e.g. quota exhausted, invalid argument, permission denied) I think it should remain unset. A simple rule would be to only record finish reason if one was returned from the API?

If inference wasn't even started we would not set attribute at all.
I was thinking more of a case when stream(s) has started, but didn't finish. Let's make it more fun - there were 2 candidates / choices and one has streamed to the end and another one didn't. Until complex attributes were allowed [we couldn't have array attribute containing `null` elements](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.50.0/specification/common/README.md#attribute) like `["stop", None]`. It's probably ok now when it comes to spec, but might still be a problem with implementations.

[...]

PR #363, opened 2026-07-03
gen-ai: deprecate per-message finish reason

### LINKED ISSUE #362: [gen-ai] Avoid duplicating finish reasons in output messages
## Problem

`gen_ai.response.finish_reasons` already represents finish reasons at the response envelope level, while `gen_ai.output.messages[].finish_reason` duplicates the same metadata inside replayable output message content. This creates two places to report the same value and can lead to inconsistencies.

## Proposal

Make `gen_ai.response.finish_reasons` the authoritative source for per-generation finish reasons. Treat `gen_ai.output.messages[].finish_reason` as optional deprecated compatibility metadata, and update examples/reference scenarios to prefer output messages that contain replayable message content without duplicated response envelope metadata.

## Scope

- Keep `gen_ai.response.finish_reasons` populated wherever known.
- Remove finish reasons from preferred `gen_ai.output.messages` examples.
- Mark `OutputMessage.finish_reason` optional/deprecated in the non-normative model and generated schema.
- Preserve the existing enum for compatibility.

yyuuttaaoo:
## Description

Deprecate `gen_ai.output.messages[].finish_reason` in favor of `gen_ai.response.finish_reasons`, so finish reasons are not duplicated inside replayable output messages.

This also clarifies the existing positional semantics of `gen_ai.response.finish_reasons`: values correspond to provider generations in order, and an expected position with no terminal reason is reported as `error` after generation has started. If inference never starts, the attribute remains unset.

Fixes #362

## Motivation

Finish reasons are response metadata used for incident triage, evaluation, and dashboards. Keeping the same value in both the response envelope and each output message creates two sources that can diverge.

Multiple choices are not introduced by this PR; `gen_ai.response.finish_reasons` has always been an array corresponding to generations. The new guidance makes missing positions deterministic instead of shifting later values left.

`error` is intentionally a normalized abnormal-termination result. It may be reported by a provider or inferred by instrumentation; this attribute does not encode provenance.

## Changes

- Deprecate the per-message field while retaining its meaning for compatibility.
- Clarify positional ordering, missing positions, and output-message filtering.
- Preserve finish reasons in OpenAI Responses and LiteLLM streaming scenarios.
- Replace finish-reason filters with positional `error` placeholders in affected scenarios.
- Add an OpenAI multi-choice reference path that emits `["stop", "error"]` from the request/response boundary.
- Regenerate the JSON schema and Weaver documentation.
- Add separate deprecation and clarification changelog fragments.

## Validation

- Weaver registry and Markdown generation against the pinned semantic-conventions v1.44.0 model.
- `make generate-json-schemas`.
- Ruff and Python compilation for all touched scenarios.
- OpenAI reference scenario against the pinned conformance mock server.
- Focused assertions for `["stop", "error"]`, an index gap, and Responses status mapping.

## Checklist

- [x] Motivation documented
- [x] Reference scenarios updated
- [x] Towncrier fragments added


changelog.d/363.deprecation.md
```diff
@@ -0,0 +1 @@
+Deprecate `gen_ai.output.messages[].finish_reason` in favor of the authoritative `gen_ai.response.finish_reasons` attribute.
```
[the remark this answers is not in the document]
yyuuttaaoo: Fixed in 50e62a2 by renaming the Towncrier fragment to `363.deprecation.md`.

model/gen-ai/registry.yaml
```diff
@@ -192,6 +192,9 @@ attributes:
   - key: gen_ai.response.finish_reasons
     type: string[]
     brief: Array of reasons the model stopped generating tokens, corresponding to each generation received.
+    note: |
+      This is the authoritative attribute for finish reasons. Values correspond
+      to generations in the same order as the returned choices/candidates.
```
```suggestion
```

is this necessary?

model/gen-ai/registry.yaml
```diff
@@ -759,6 +762,12 @@ attributes:
       (choice/candidate) and vice versa - one choice cannot be split across
       multiple messages or one message cannot contain parts from multiple choices.
 
+      This attribute SHOULD describe replayable output message content and
+      SHOULD NOT duplicate response envelope metadata, including finish
+      reasons, response identifiers, response model, token usage, cost, or
+      timing metadata. Finish reasons SHOULD be recorded in
+      `gen_ai.response.finish_reasons`.
```
```suggestion
```

this seems to be unnecessary if finish reason is removed from the model

model/gen-ai/gen-ai-output-messages.json
```diff
@@ -290,16 +290,20 @@
...
                             "type": "string"
+                        },
+                        {
+                            "type": "null"
                         }
                     ],
-                    "description": "Reason for finishing the generation.",
+                    "default": null,
+                    "deprecated": true,
+                    "description": "Deprecated. Use `gen_ai.response.finish_reasons` as the authoritative source for finish reasons.",
```
lmolkova: this means that we need a better story for finish_reasons to be guaranteed and be in the same order as candidates, see https://github.com/open-telemetry/semantic-conventions-genai/issues/56 for some context.

E.g. user asked for 3 choices / candidates, first one finished with "stop", second one got truncated with "length", third never finished streaming and failed. 

Today we don't really prescribe what finish reason should be in case of failure. In practice, you'd see it not being populated, set to None or error. I.e. with 3 output assistant messages, users would sometimes see ["stop", "max_length"]. 

It's not awesome even today, but will get even worse if we make attribute a single source of this information. 

So please update the finish_reasons attribute to provide definitive guidance on what to do in case of failures / response being cancelled before stream has ended and any other case when finish reason is not available. My preference is to use "error" whenever we expected but never received finish reason.

model/gen-ai/registry.yaml
yyuuttaaoo: Accepted and updated in 63c02da. I removed the unnecessary "authoritative" wording and replaced it with concrete ordering and missing-finish-reason guidance on `gen_ai.response.finish_reasons`.

model/gen-ai/gen-ai-output-messages.json
yyuuttaaoo: Accepted in 63c02da. `gen_ai.response.finish_reasons` now says values correspond to returned choices/candidates in order, and that if a finish reason was expected but not received because generation failed, was cancelled, or the stream ended before the final event, instrumentation SHOULD report `error` for that position instead of omitting it.

model/gen-ai/registry.yaml
yyuuttaaoo: Accepted in 63c02da. I removed this output-message note and kept the guidance focused on the `gen_ai.response.finish_reasons` attribute itself.

templates/registry/markdown/attribute_macros.j2
```diff
@@ -23,13 +23,12 @@
...
-   (their YAML file lives under ./model/) and is null for attributes imported
-   from a dependency. Local attributes link to /docs/registry/attributes/<ns>.md;
-   imported attributes link to the pinned upstream registry page. #}
+{# v2 link rendering: this repo only generates attribute pages for namespaces
+   owned in this repository. Attributes imported from the pinned upstream
+   registry must link to the pinned upstream docs page. #}
 {% macro name_with_link(attribute, attribute_registry_base_url, upstream_docs_base) -%}
 {%- set ns_id = name(attribute) | trim | split_id | first -%}
-[`{{ name(attribute) | trim }}`]({% if attribute.provenance %}{{ attribute_registry_base_url }}/{{ ns_id | kebab_case }}.md{% else %}{{ upstream_docs_base }}/docs/registry/attributes/{{ ns_id | kebab_case }}.md{% endif %})
+[`{{ name(attribute) | trim }}`]({% if ns_id in ["aws", "gen_ai", "mcp", "openai"] %}{{ attribute_registry_base_url }}/{{ ns_id | kebab_case }}.md{% else %}{{ upstream_docs_base }}/docs/registry/attributes/{{ ns_id | kebab_case }}.md{% endif %})
```
lmolkova: this should not be necessary, fixed in https://github.com/open-telemetry/semantic-conventions-genai/pull/405

lmolkova (approved):

changelog.d/363.deprecation.md
trask: Parking my PR-level notes here since there's no better anchor.

**Terminology is now inconsistent across the PR.** After the review round removed the "authoritative" framing from `model/gen-ai/registry.yaml`, that word survives in three places the reader still sees: this fragment, the `finish_reason` description in `docs/gen-ai/non-normative/models.py`, and the generated `model/gen-ai/gen-ai-output-messages.json`. The registry itself now only describes ordering and missing-position handling. Worth making these agree — and the description's Prototype bullet ("Clarifies in `model/gen-ai/registry.yaml` that `gen_ai.response.finish_reasons` is authoritative and that output messages should avoid duplicating response-envelope metadata") no longer describes the diff and should be updated.

**A second fragment is needed.** This one covers the deprecation, but the PR also adds new normative guidance to `gen_ai.response.finish_reasons` — positional correspondence, and reporting `error` for a position the provider didn't report. `changelog.d/README.md` asks for one fragment per consumer-visible change, and #217 set the precedent of a `breaking` + `clarification` pair in one PR:

`changelog.d/363.clarification.md`

```
Clarify that `gen_ai.response.finish_reasons` values correspond positionally to the returned choices/candidates, and that instrumentations report `error` for a position where the provider did not report a finish reason.
```

**One note on validation, since it affects what CI can tell you here.** CI does run all 15 scenarios and diffs each `data.json`, so scenario breakage is covered. But `data.json` records a *union of attribute names per span type per library* — `reference/scenarios/openai/data.json` lists `gen_ai.response.finish_reasons` once under `spans.inference` for the whole library. So a single inference span losing that attribute is invisible to CI as long as any other inference span in the same library still sets it. That's the blind spot behind the two comments I left on `reference/scenarios/openai/scenario.py` and `reference/scenarios/litellm/scenario.py`; both need a human read rather than a green check.

model/gen-ai/registry.yaml
```diff
@@ -201,9 +201,20 @@ attributes:
...
+    note: |
+      Values correspond to generations in the same order as the returned
+      choices/candidates.
+
+      When a request is expected to produce multiple choices/candidates, each
+      position SHOULD contain the finish reason for the corresponding
+      choice/candidate. If a finish reason was expected but not received, for
+      example because generation failed, was cancelled, or a stream ended before
+      the final event, instrumentations SHOULD report `error` for that
+      position instead of omitting it.
```
trask: This paragraph introduces normative behavior that nothing in the repo demonstrates, and that a good deal of existing reference code contradicts.

**No reference coverage.** Per AGENTS.md ("No convention without a reference scenario"), the `error`-for-missing-position rule needs a scenario that emits it. I checked every finish-reason emission across `reference/scenarios/` and none produces an array containing `error`, and none exercises a response where one candidate reports a finish reason and another doesn't. The rule ships unexercised.

**Five scenarios currently do the opposite.** These all filter out missing positions, which shifts every later value left — precisely what the new SHOULD forbids:

- `reference/scenarios/mistralai/scenario.py:63,91,151` — `[c.finish_reason for c in resp.choices if c.finish_reason]` (touched by this PR)
- `reference/scenarios/azure-ai-inference/scenario.py:43,144` (touched by this PR)
- `reference/scenarios/dspy/scenario.py:67,115,205` (touched by this PR)
- `reference/scenarios/autogen/scenario.py:207`
- `reference/scenarios/openai-agents/scenario.py:142`

None of them violates the rule *observably* today, since the mocks always return a non-null finish reason — but they encode the shifting behavior, and three are already open in this PR. Worth aligning at least those three, or noting why the pattern is acceptable.

(`reference/scenarios/autogen/scenario.py:323` is the one place that already does the right thing — `[c.finish_reason for c in resp.choices]`, no filter.)

model/gen-ai/registry.yaml
```diff
@@ -201,9 +201,20 @@ attributes:
...
     examples:
       - [stop]
       - [stop, length]
+      - [stop, length, error]
```
trask, line 228: Adding `error` to the examples surfaces a collision that's reproducible in this repo: **`error` is already a value providers return.**

- **Mistral** — `ChatCompletionChoiceFinishReason` is `stop | length | model_length | error | tool_calls`. Lowercase `error`, byte-identical to the value proposed here. ([SDK docs](https://github.com/mistralai/client-python/blob/main/docs/models/chatcompletionchoicefinishreason.md))
- **Cohere** — `ERROR` in v2 Chat, plus `ERROR_TOXIC` and `ERROR_LIMIT` in v1. Different casing, so this one only collides after case-folding — but `reference/scenarios/dspy/scenario.py:65,113,203` already lowercases finish reasons (`str(choice.finish_reason).lower()`), so that's not hypothetical either.

Both libraries have scenarios here, and both pass the provider value straight through:

- `reference/scenarios/mistralai/scenario.py:63,91,151` — `[c.finish_reason for c in resp.choices if c.finish_reason]`
- `reference/scenarios/cohere/scenario.py:46,73,127` — `[resp.finish_reason]`

So a Mistral generation that fails server-side emits `["error"]` today, and that is indistinguishable from what this note asks instrumentations to synthesize for a position the provider *didn't* report. A consumer can't tell "Mistral reported an error for this choice" from "the instrumentation found no finish reason at this position" — and those mean quite different things. OpenAI, Anthropic, Gemini/Vertex and Bedrock have no error-like value, so they're unaffected.

To be clear about what is *not* changing: `gen_ai.response.finish_reasons` is a bare `string[]` with no defined members, and the deprecated per-message field never constrained values either — its schema is `anyOf: [FinishReason, string]`, so that enum was always advisory. This PR doesn't remove a normalization guarantee. The narrower problem is that it's the first time a synthetic value is written into an array that otherwise holds raw provider strings, and the value chosen already exists in two providers' vocabularies.

This also looks like the concrete form of a question left open upstream. In #56 lmolkova wrote "I'm fine aligning on null/None, but I want to better understand concerns with `error` too", after @aabmass raised that `error` feels wrong as a finish reason. lmolkova's follow-up scoped it ("If inference wasn't even started we would not set attribute at all"), which this note's wording matches — but the null-vs-`error` choice itself doesn't appear to have been closed out, and this PR makes `error` normative. Worth confirming that's a settled decision rather than one resolved by side effect here.

Two ways out, depending on how settled you consider it:

- If `error` is settled, a sentence in the note saying the value is instrumentation-supplied rather than provider-reported at least documents the ambiguity — though given Mistral it doesn't let a consumer actually resolve it at query time.
- If it isn't, `null` sidesteps the collision entirely: a null at position *i* unambiguously means "not reported" and can't be confused with anything a provider returns. That does require the attribute to permit nulls in the array.

Defining a full normalized value set would also resolve it, but that needs explicit provider mappings and reference updates, and is probably its own PR.

model/gen-ai/registry.yaml
```diff
@@ -786,8 +797,7 @@ attributes:
                 "type": "text",
                 "content": "The weather in Paris is currently rainy with a temperature of 57°F."
               }
-            ],
-            "finish_reason": "stop"
+            ]
           }
         ]
```
trask, line 960: Now that `finish_reason` is gone from the message objects, position is the only thing linking `gen_ai.output.messages` to `gen_ai.response.finish_reasons`. But the note just above says instrumentations MAY filter or truncate output messages, and a filter that drops a message from the middle breaks that alignment — with no per-message field left to re-associate the two.

Is that OK? It may well be intentional, since finish reasons are response-level and arguably aren't meant to line up with a user-filtered message list. If so, it might be worth saying so explicitly, because the 1:1 message-to-choice rule stated just above invites the assumption.

docs/gen-ai/non-normative/models.py
```diff
@@ -348,8 +348,10 @@ class OutputMessage(ChatMessage):
     specific response (choice, candidate).
     """
 
-    finish_reason: Union[FinishReason, str] = Field(
-        description="Reason for finishing the generation."
+    finish_reason: Optional[Union[FinishReason, str]] = Field(
+        default=None,
+        description="Deprecated. Use `gen_ai.response.finish_reasons` as the authoritative source for finish reasons.",
+        deprecated=True,
     )
```
trask, line 358: The new description replaces the field's meaning with a pointer, so the property no longer says what it holds. Instrumentations MUST follow this schema, and it's also how previously-emitted telemetry gets interpreted. Keep both:

```suggestion
    finish_reason: Optional[Union[FinishReason, str]] = Field(
        default=None,
        description=(
            "Deprecated: report finish reasons in `gen_ai.response.finish_reasons` instead. "
            "Reason for finishing the generation."
        ),
        deprecated=True,
    )
```

I applied this and ran `make generate-json-schemas`: `model/gen-ai/gen-ai-output-messages.json` regenerates with exactly one line changed (the `description`), so that file needs regenerating and committing alongside. Also confirmed the schema as committed in this PR already regenerates byte-for-byte from the current `models.py`, so nothing else is out of sync.

(This also drops the last "authoritative" wording from the model, which no longer matches `registry.yaml`.)

reference/scenarios/openai/scenario.py
```diff
@@ -55,9 +55,9 @@ def responses_output_messages(response):
             if text:
                 parts.append({"type": "text", "content": text})
         if parts:
-            output_messages.append({"role": role or "assistant", "parts": parts, "finish_reason": "stop"})
+            output_messages.append({"role": role or "assistant", "parts": parts})
     if pending_parts:
-        output_messages.append({"role": "assistant", "parts": pending_parts, "finish_reason": "compaction"})
+        output_messages.append({"role": "assistant", "parts": pending_parts})
```
trask, line 87: This helper feeds two functions that set no finish reason anywhere else:

- `run_responses_compaction_reference` (line 174)
- `run_responses_continuation_reference` (line 251)

Neither sets `gen_ai.response.finish_reasons` on its span or its event, so after this change both emit no finish reason at all, where previously the value rode inside the output message. #362's own scope says "Keep `gen_ai.response.finish_reasons` populated wherever known", so this looks unintended. CI won't flag it — `openai/data.json` aggregates attribute names across all inference spans in the library, and the chat-completions spans still set it.

The fix I'd suggest is *not* to copy the `["stop"]` literal from `run_responses_with_prompt_template_reference` (line 647). `.github/instructions/reference-scenarios.instructions.md` is explicit that "response ids, finish reasons, and token counts should come from the current response or streamed result object" — that existing literal is weak by the repo's own rule, and propagating it would make things worse.

The Responses API exposes this properly via `response.status` (`completed` / `incomplete`) and `response.incomplete_details.reason`. `RESPONSES_RESPONSE` in `reference/src/semconv_genai/mock_server/openai.py` doesn't currently set `status`, so this needs a small mock addition first, after which both functions can derive the value from the response object.

If you'd rather keep the mock as-is, the honest alternative is to leave these two spans without a finish reason and call it a capture gap — but then it's worth saying so, since it reads as an accidental drop right now.

One thing you can safely ignore: the `"finish_reason": "compaction"` you removed from the `pending_parts` branch was unreachable with the current mock. The compaction item is appended *after* the message item, so it always merges into the existing assistant message and that branch never runs. `compaction` remains demonstrated end-to-end by `reference/scenarios/anthropic/scenario.py`, which reads `resp.stop_reason` into `gen_ai.response.finish_reasons`.

reference/scenarios/litellm/scenario.py
```diff
@@ -138,20 +136,16 @@ def run_chat_streaming():
             stream=True,
         )
         text = ""
-        finish_reason = None
         for chunk in resp:
             if chunk.choices[0].delta.content:
                 text += chunk.choices[0].delta.content
```
trask: Straight loss here: `run_chat_streaming` used to accumulate the streaming finish reason, and now it's discarded without being moved anywhere — nothing else in this function sets `gen_ai.response.finish_reasons`, so the span ends up with no finish reason at all.

Compare `run_chat_streaming_reference` in `reference/scenarios/openai/scenario.py` and `run_chat_streaming` in `reference/scenarios/mistralai/scenario.py`: both keep the accumulator and promote it to the response attribute. The mock's final chunk carries `finish_reason: "stop"` (`reference/src/semconv_genai/mock_server/openai.py`), and the deleted code already read `chunk.choices[0].finish_reason`, so this is capturable from the stream rather than hardcoded:

```suggestion
        text = ""
        finish_reasons = []
        for chunk in resp:
            if chunk.choices[0].delta.content:
                text += chunk.choices[0].delta.content
            if chunk.choices[0].finish_reason is not None:
                finish_reasons.append(chunk.choices[0].finish_reason)
        if finish_reasons:
            span.set_attribute("gen_ai.response.finish_reasons", finish_reasons)
```

trask (commented):

yyuuttaaoo (commented):
Addressed the latest review feedback in 7119778.

changelog.d/363.deprecation.md
yyuuttaaoo: Good catches. Added the clarification fragment, removed the remaining `authoritative` wording, updated the PR description, regenerated docs/schema, and fixed the per-span capture gaps below.

model/gen-ai/registry.yaml
yyuuttaaoo: Agreed. Added an OpenAI `n=2` reference path where the mock returns one indexed choice, so the instrumentation derives `["stop", "error"]` from the request/response boundary. I also replaced the filtering patterns in the current Mistral, Azure AI Inference, AutoGen, OpenAI Agents, and CrewAI scenarios, and made streaming accumulation positional.

model/gen-ai/registry.yaml
yyuuttaaoo: I’m keeping `error` intentionally. This attribute represents the normalized termination outcome, not its provenance: provider-reported and instrumentation-inferred `error` both mean that generation ended abnormally. I added that sentence explicitly. The inference is only made once a generation/position exists or is expected; if inference never starts, the attribute remains unset.

model/gen-ai/registry.yaml
yyuuttaaoo: Agreed. Clarified that `gen_ai.response.finish_reasons` remains aligned with provider generations, not with a filtered or truncated `gen_ai.output.messages` value.

docs/gen-ai/non-normative/models.py
yyuuttaaoo: Agreed. Kept both the deprecation pointer and the field meaning, then regenerated the JSON schema.

reference/scenarios/openai/scenario.py
yyuuttaaoo: Agreed. Current main’s Responses mock now exposes `response.status`. Added a shared status-to-finish-reason mapping and used it for the compaction, continuation, prompt-template, and fetch spans/events, so the value comes from the current response rather than a literal.

reference/scenarios/litellm/scenario.py
yyuuttaaoo: Agreed. Restored streaming finish-reason capture. It now tracks choice indexes and emits `error` for any started position that never receives a terminal finish reason.

reference/scenarios/openai/scenario.py
yyuuttaaoo: Follow-up: the Anthropic reference currently uses opentelemetry-util-genai 1.0b0, whose OutputMessage constructor still requires finish_reason. I kept that compatibility-only value in the Anthropic scenario (while the response attribute remains populated) and called it out in code; the preferred dict/schema examples remain de-duplicated.

model/gen-ai/registry.yaml
```diff
@@ -210,9 +210,22 @@ attributes:
...
     brief: Array of reasons the model stopped generating tokens, corresponding to each generation received.
+    note: |
+      Values correspond to generations in the same order as the returned
+      choices/candidates.
+
+      Each position SHOULD contain the finish reason for the corresponding
+      choice/candidate. If a finish reason was expected but not received, for
+      example because generation failed, was cancelled, or a stream ended before
+      the final event, instrumentations SHOULD report `error` for that position
+      instead of omitting it.
```
trask: The note ties positions to the returned choices/candidates, but the OpenAI reference sizes the array from `gen_ai.request.choice.count`: with `n=2` the mock returns only choice 0, so it emits `["stop", "error"]`. This wording also permits `["stop"]` there. Say whether a known requested choice count sets the array length when the provider returns fewer generations.

model/gen-ai/registry.yaml
```diff
@@ -849,7 +862,9 @@ attributes:
       multiple messages or one message cannot contain parts from multiple choices.
 
       Instrumentations MAY provide a way for users to filter or truncate
-      output messages.
+      output messages. `gen_ai.response.finish_reasons` remains aligned with the
+      generations returned by the provider, not with a filtered or truncated
+      `gen_ai.output.messages` value.
```
trask: The schema marks `finish_reason` deprecated, but the generated attribute notes do not, so a reader only finds this migration guidance by opening the schema.

```suggestion
      output messages. `gen_ai.response.finish_reasons` remains aligned with the
      generations returned by the provider, not with a filtered or truncated
      `gen_ai.output.messages` value.

      The per-message `finish_reason` field is deprecated; report finish reasons
      in `gen_ai.response.finish_reasons` instead.
```

JWinermaSplunk (approved):
LGTM! If I am correct, let's open up a follow up issue about the Anthropic scenario. There are also some small inconsistencies between checking if finish_reasons is defined before setting in spans, events, etc in the scenarios, so I wouldn't mind if we standardized all to check if they are defined before setting the attribute. Thanks!

yyuuttaaoo:
Thanks! Created #471 to track the Anthropic/util-genai cleanup.

The conditional checks are intentional: record `error` when a generation was expected but its finish reason is missing, and omit the attribute when no generation exists or the response is non-terminal.

yyuuttaaoo:
Hi @lmolkova and @JWinermaSplunk — I’ve merged the latest `main`, regenerated the generated docs and reference data, and resolved the merge conflict. The follow-up `eb4414e` refreshes the reference findings for the new conformance runner. The PR is mergeable again, and the fresh CI is fully green (41/41 checks). When you have a chance, could you please take another look? Thank you!

Issue #364, opened 2026-07-03
Add delta-based input message capture for repeated GenAI conversations


yyuuttaaoo:
## Problem

`gen_ai.input.messages` is useful for reconstructing a GenAI model request, but in multi-turn chat and agent loops it often repeats a large stable prompt or conversation prefix on every span. That repeated prefix can dominate telemetry storage even when the effective input only changes by one or a few appended messages.

This is especially visible in workloads with prompt caching or high conversation-prefix reuse: the provider may only process a small delta, while telemetry still records the full message array for every request. In those workloads, delta capture can substantially reduce content storage, and in some repeated-prefix cases the reduction can approach the same order as the prompt cache hit ratio.

## Proposal

Add an opt-in content attribute:

`gen_ai.input.messages_delta`

The attribute would use the same schema as `gen_ai.input.messages`, but represent only new input messages appended since the previous GenAI model request in the same conversation.

Suggested guidance:

- Use with `gen_ai.conversation.id` to correlate the ordered conversation.
- For the first request in a conversation, implementations may record the full initial input in `gen_ai.input.messages_delta`.
- When the effective input changes in a non-append-only way, such as compaction, truncation, prompt rewrite, or user-visible history editing, record `gen_ai.input.messages` as a full checkpoint instead.
- When `gen_ai.output.messages` is also captured, consumers can reconstruct each later request by starting from the latest `gen_ai.input.messages` checkpoint and applying subsequent `gen_ai.output.messages` and `gen_ai.input.messages_delta` values in operation order.

## Tradeoffs

This attribute has the same sensitive-data and PII risks as `gen_ai.input.messages`, so it should remain opt-in. Reconstruction also depends on having sufficient ordering and correlation data, and on content capture being enabled for the relevant spans.

jaden3824:
Useful proposal. I think the guidance should explicitly separate two quantities that are easy to conflate:

1. content retained in telemetry; and
2. provider-reported model input/cache usage.

gen_ai.input.messages_delta can reduce the first substantially while leaving the second unchanged if the full conversation is reconstructed before inference. I would also record a checkpoint/reset reason for compaction, rewrite, or truncation, and keep provider usage nullable when it is unavailable rather than inferring zero.

[...]

PR #365, opened 2026-07-03
Add GenAI input messages delta attribute

yyuuttaaoo:
## Description

Adds `gen_ai.input.messages_delta` as an opt-in GenAI content attribute using
the same schema as `gen_ai.input.messages`.

This attribute represents input messages appended since earlier telemetry in the
same `gen_ai.conversation.id`. It is intended for emitters that can reliably
identify appended messages and correlate them with prior turns, while falling
back to `gen_ai.input.messages` when they cannot.

Refs #364

## Motivation

`gen_ai.input.messages` is useful for reconstructing model requests, but
multi-turn chat and ReAct-style agent loops often send the accumulated
conversation history on every model call. This can repeatedly record a large
stable prompt or conversation prefix even when each turn only appends a small
user message, tool result, or observation.

**User journey:** this supports incident triage, cost/storage investigation,
quality evaluation, and security/audit workflows where users need enough content
telemetry to reconstruct agent behavior without storing the same conversation
prefix on every request.

**Prior art:** agent runtimes, hooks, and application-level instrumentation often
already observe turn-level input deltas. Content offload/upload patterns also
exist for reducing repeated large payloads; `messages_delta` addresses the
complementary case where the repeated structure is the growing conversation
history itself.

## Prototype

Reference scenarios were added for multi-turn behavior in:

- `reference/scenarios/openai/scenario.py`
- `reference/scenarios/anthropic/scenario.py`
- `reference/scenarios/vertexai/scenario.py`

Each scenario emits `gen_ai.input.messages_delta` with
`gen_ai.conversation.id`, refreshes its `data.json`, and updates the reference
reports.

Validation used during this PR included:

- `make generate-registry`
- `make generate-docs`
- `make check-policies`
- `uv run --locked --default-index https://pypi.org/simple update-reports` from
  `reference/`
- `git diff --check`

## Checklist

See [CONTRIBUTING.md]


model/gen-ai/registry.yaml
```diff
@@ -747,6 +747,60 @@ attributes:
           }
         ]
     stability: development
+  - key: gen_ai.input.messages_delta
+    type: any
+    brief: >
+      New input messages appended since the previous GenAI model request in
+      the same conversation.
+    annotations:
+      type:
```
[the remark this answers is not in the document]
yyuuttaaoo: Addressed in 476243f: added multi-turn reference scenarios for OpenAI Chat Completions, Anthropic Messages, and Vertex AI. Each scenario emits `gen_ai.input.messages_delta` together with `gen_ai.conversation.id`, refreshes its `data.json`, and updates the reference reports. The examples intentionally use full-context/stateless request APIs where applicable, so the storage benefit is demonstrated as a telemetry delta even when the provider request still carries the accumulated history.

model/gen-ai/registry.yaml
```diff
@@ -747,6 +747,60 @@ attributes:
...
+        json_schema: model/gen-ai/gen-ai-input-messages.json
+    note: |
+      This attribute has the same schema as `gen_ai.input.messages`, but only
+      contains the new messages that were not already captured by earlier
+      telemetry in the same `gen_ai.conversation.id`.
+
+      Instrumentations SHOULD prefer `gen_ai.input.messages_delta` over
+      `gen_ai.input.messages` when the effective input context is append-only.
+      For the first request in a conversation, this attribute SHOULD contain the
+      full initial input if it is used.
```
lmolkova: instrumentation is stateless, how would it know what was sent in the previous request?

Also pretty much nothing records `gen_ai.conversation.id`

yyuuttaaoo: Thanks, this is a fair concern for low-level provider SDK instrumentation. If it only observes independent provider requests, it should not try to infer deltas.

My intent is broader: GenAI SemConv also applies to application, framework, agent-runtime, and hook-based instrumentation. Those emitters may already have session context or receive only the appended turn. OpenClaw, Claude Code, and Codex-style hooks are examples where turn-level deltas are directly observable.

So I think this attribute is useful, but should be scoped: emit `gen_ai.input.messages_delta` only when the emitter can reliably identify the appended messages and correlate them to prior turns, typically via `gen_ai.conversation.id` or equivalent context. Otherwise prefer `gen_ai.input.messages`.

aabmass: @yyuuttaaoo I'm curious if you've seen the uploading hook which allows you to use an fsspec path to upload inputs/outputs? The reason I bring it up is it has [support for hashing system instructions](https://github.com/open-telemetry/opentelemetry-python-genai/commit/3a9348eba054987eba5bdc6e1f90cdb19cfd0325) so the statefulness is pushed to the "filesystem" and I think we could expand this approach to input/output parts.

yyuuttaaoo: @aabmass Yes, I’m familiar with that approach. We actually adopted a similar content-addressed upload pattern for multimodal payload offloading in our distribution, inspired by the upstream uploader: [uploading multimodal messages](https://github.com/alibaba/loongsuite-python/pull/94/changes#diff-00911ad5c6b6419ddc6cefecd66bf06ce30857ec3dc154af4418ca5a66b14d8c).

I agree it works well for large stable artifacts such as system instructions, tool definitions, documents, or multimodal blobs.

For input/output messages, I think it is complementary but not a full replacement for deltas. In ReAct-style loops, each request often repeats a growing conversation prefix. If we expand refs down to message/part level, we may still repeat a growing list of refs each turn, and potentially create many small uploaded objects.

I’ll add [a separate comment](https://github.com/open-telemetry/semantic-conventions-genai/pull/365#issuecomment-4917017583) with the lightweight delta algorithm I have in mind.

yyuuttaaoo:
For more context, the target pattern here is common ReAct-style agent execution.

A later model request often contains the accumulated conversation history:

```text
request[N] =
  input[1] + output[1] +
  input[2] + output[2] +
  ...
  input[N-1] + output[N-1] +
  new_input[N]
```

The repeated prefix can be large, while the truly new input is often small: a user turn, a tool result, or another observation.
This can be implemented with very small per-conversation state. The emitter only needs to keep:

```text
chain_hash
chain_length
```

Both input messages and output messages are appended to the same rolling hash chain:

```python
H0 = sha256("")
Hn = sha256(Hn-1 + stable_serialize(message_n))
```

On each request:
1. Use chain_length as the expected prefix length.
2. Hash the request’s prefix of that length.
3. If the prefix hash matches chain_hash, emit only:

```python
gen_ai.input.messages_delta = request_messages[chain_length:]
```

4. Advance the chain with those delta messages.
5. When the response is available, also advance the chain with gen_ai.output.messages, because those outputs normally become part of the next request’s history.
6. If the prefix hash does not match, emit full gen_ai.input.messages and reset the chain. This handles compaction, truncation, inserted context, edited history, session restore, or any non-append-only change.

So this does not require storing the full conversation in the instrumentation. It only requires a small hash and count per conversation, while still detecting when delta capture is unsafe and falling back to a full input snapshot.

We have prototyped this in our collector here: [local input messages delta](https://github.com/alibaba/loongcollector/pull/2579/changes#diff-a730c5a35b0fa4b59dd5a77006fc9ea1d2e8253d7eb3e2ec71ed27387731b497R365)

model/gen-ai/registry.yaml
yyuuttaaoo: I also softened the PR text in 1abfa79 to match the opt-in direction from the discussion: `gen_ai.input.messages_delta` is now described as something emitters MAY use when they can reliably identify appended messages and correlate them with prior turns. It no longer says instrumentations should prefer deltas over full `gen_ai.input.messages`, and it keeps `gen_ai.conversation.id` as the recommended correlation key when the delta attribute is emitted.

model/gen-ai/registry.yaml
```diff
@@ -747,6 +747,50 @@ attributes:
...
+      See [Recording content on attributes](/docs/gen-ai/gen-ai-spans.md#recording-content-on-attributes)
+      section for more details.
+    examples:
+      - |
+        [
+          {
+            "role": "tool",
+            "parts": [
+              {
+                "type": "tool_call_response",
```
alexmojaki, line 977: this example looks like a mistake to me, unless the previous turn's `gen_ai.output.messages` are also considered to be implicitly appended to the current turn's input messages before adding the delta. is that the intention?

model/gen-ai/registry.yaml
```diff
@@ -747,6 +747,50 @@ attributes:
...
+      the same conversation.
+    annotations:
+      type:
+        json_schema: model/gen-ai/gen-ai-input-messages.json
+    note: |
+      This attribute has the same schema as `gen_ai.input.messages`, but contains
+      only messages appended since earlier telemetry in the same
+      `gen_ai.conversation.id`.
+
+      Emitters MAY use this attribute when they can reliably identify appended
```
alexmojaki, line 938: This attribute requires extra work for telemetry backends/consumers, it won't always be supported. Users should probably not get deltas by default even when they've enabled recording messages, it should be an additional layer of opting in.

model/gen-ai/registry.yaml
```diff
@@ -747,6 +747,50 @@ attributes:
...
+      messages and correlate them with prior turns. When emitting this attribute,
```
alexmojaki: I think a span link to the previous turn would be better than `gen_ai.conversation.id`, and could be an independently useful proposal.

yyuuttaaoo: I agree span links could be useful for representing previous-turn dependency, and this may be worth a separate proposal. For this PR, I would keep the existing `gen_ai.conversation.id` as the correlation key.

model/gen-ai/registry.yaml
yyuuttaaoo: Yes, that is the intention. The reconstruction state advances in operation order: apply the recorded input snapshot or delta, then append the corresponding `gen_ai.output.messages`. In this example, the previous assistant tool call is therefore already in the reconstructed state before the tool-response delta is applied. I agree the current text does not make this clear and will clarify it.

model/gen-ai/registry.yaml
yyuuttaaoo: Good point. `gen_ai.input.messages_delta` is already Opt-In, but I agree that enabling general message content capture should not implicitly select delta encoding. I will clarify that instrumentations should treat delta encoding as a separate opt-in. Manual instrumentation opts in by emitting the attribute explicitly.

model/gen-ai/registry.yaml
alexmojaki: > even when spans are sampled separately

i don't know what point you're making here. how are you handling the previous turn being sampled out? doesn't the whole system fall apart?

yyuuttaaoo: Sorry, my previous wording was incorrect. I only meant that this PR reuses the existing conversation identifier. I have updated the comment.

alexmojaki: the point about sampling remains. if some spans are excluded by sampling then this system doesn't work without adjustment. and i'm not sure if instrumentations can actually tell whether that happened.

model/gen-ai/registry.yaml
yyuuttaaoo: Yes. The reconstruction guidance now explicitly says to apply recorded `gen_ai.output.messages` and `gen_ai.input.messages_delta` values in operation order. The previous assistant tool call is therefore appended to the reconstructed conversation state before the next tool-response delta is applied.

model/gen-ai/registry.yaml
yyuuttaaoo: Addressed in 9ff2523. The convention now says that instrumentations SHOULD require a separate opt-in for delta encoding and that enabling input message capture SHOULD NOT implicitly enable it.

model/gen-ai/registry.yaml
yyuuttaaoo: Agreed. Instrumentation generally cannot know whether a downstream sampling or processing stage discarded another operation, since telemetry creation and sampling/export are typically decoupled. This delta scheme therefore assumes complete retention of the related GenAI operation telemetry, as in LLM trace and agent-analysis scenarios. It is not safe for independently sampled traditional spans. In 9ff2523 I clarified that emitters SHOULD use full `gen_ai.input.messages` whenever sampling or other telemetry processing may independently discard an operation needed for reconstruction.

yyuuttaaoo:
> /dashboard route:reviewers

/dashboard route:reviewers

reference/scenarios/openai/scenario.py
aabmass, line 1: This feature seems more interesting in the context of the [OpenAI Responses](https://developers.openai.com/api/docs/guides/migrate-to-responses?lang=javascript&update-multiturn=responses#3-update-multi-turn-conversations) and Gemini Interactions APIs, where the API supports delta via saved state on the server.

If we model against these APIs, you can also handle cases like branched conversations because the `previous_response_id` is captured. Apologies if this was already discussed on this PR but I don't see it in scenarios.

yyuuttaaoo: Good point. I saw the recent addition of `gen_ai.request.previous_response.id`, and agree that server-managed continuation is a better fit. In f618e07, I added OpenAI Responses and Google Interactions examples using it, while retaining `gen_ai.conversation.id` for conversation-level correlation. I kept the Chat Completions example for the widely used stateless case.

yyuuttaaoo:
Hi @lmolkova, @aabmass, and @alexmojaki — I’ve merged the latest `main`, resolved the conflicts, regenerated the docs, and rerun the Google GenAI reference scenario. When you have a chance, could you please take another look? Thank you!

model/gen-ai/registry.yaml
alexmojaki: > Yes. The reconstruction guidance now explicitly says to apply recorded `gen_ai.output.messages` and `gen_ai.input.messages_delta` values in operation order.

where does it say that?

yyuuttaaoo: Good catch — that reconstruction order was stated in the hand-written span guidance, but not in the attribute definition itself. I made it explicit in 6259b06: process correlated operations in order; for each operation, replace the state with `gen_ai.input.messages` when present (otherwise append `gen_ai.input.messages_delta`), then append that operation’s `gen_ai.output.messages` before processing the next operation. Thanks for pointing this out.

model/gen-ai/registry.yaml
alexmojaki, line 1: please add a full example to https://github.com/open-telemetry/semantic-conventions-genai/blob/main/docs/gen-ai/non-normative/examples-llm-calls.md

PR #368, opened 2026-07-06
Add an opaque governance record reference for decision operations

AgentGymLeader:
This PR replaces the earlier draft's attachment plan and does not depend on #427.

SDKs already expose guardrail evaluations as spans. A draft PR against python-genai (https://github.com/open-telemetry/opentelemetry-python-genai/pull/621) maps the OpenAI Agents SDK's guardrail span data onto the span shape proposed in #427. That shape carries the verdict and little else, so the outcome and details of the evaluation stay with the component that ran it. A trace consumer needs one opaque attribute to join the span to whatever that component keeps. Nothing else here is standardized.

**`gen_ai.external_ref`** (string, development): an opaque, producer-defined identifier of an out-of-band record associated with the operation. Consumers compare it for equality only. The record's format, lifecycle, retrieval, and validation are outside this specification. meshailabs asked on this thread for the equality-only rule to be stated in the definition; it is. Producers are expected to keep the referent stable while the reference is in use; how is up to them.

Defined once in `model/gen-ai/registry.yaml` and listed as recommended, when such a record exists for the operation, on the eleven spans already on `main` (inference, embeddings, retrieval, fetch_response, memory, create_agent, invoke_agent client and internal, execute_tool, invoke_workflow, plan). The reference scenario shows it on `gen_ai.invoke_agent.internal` for both the allowed run and the tripwire run. The python-genai PR above does not emit this attribute; it only shows that evaluation spans exist in a real SDK. If #427 merges, its run-guardrail span gets the same line.

The join pattern is generic and has the same shape as `gen_ai.guardrail.security.external_finding_id` in #427. What is specific is where it is defined and listed: gen_ai spans. It carries no permission-check semantics; that topic went to open-telemetry/semantic-conventions#4022 after #461. An external RFC (OpenSecureAIAlliance/RFCs#13, cited here on 2026-08-13) describes the same split: the pointer is defined, the content is not.

The reference scenario is hand-authored. A second scenario from a real control-plane or guardrail component would be welcome.

Refs #239.


lmolkova (changes requested):
Is there a real library or component that's possible to instrument?

Manual-only instrumentations are not in scope of this project

AgentGymLeader:
@lmolkova The OpenAI Agents SDK is one: its guardrail runtime is a library-owned decision point. Reworked the scenario to instrument it, with a plain-function input guardrail (`run_in_parallel=False`), and the decision attributes are read from the SDK's own guardrail runtime objects (`InputGuardrailResult` / the tripwire exception). The allowed run executes a function tool; the tripwire run records a terminal block with zero model calls and no child execute span.

`decision.id` is omitted in the scenario — the SDK exposes no natural runtime identifier for the guardrail evaluation, so that one stays producer-supplied and opt-in.

reference/scenarios/agent-governance/scenario.py
```diff
@@ -0,0 +1,265 @@
...
+        attributes=agent_attributes,
+    ) as agent_span:
+        agent_span.set_attribute("gen_ai.input.messages", input_messages)
+        try:
+            await Runner.run(agent, prompt)
+        except InputGuardrailTripwireTriggered as exc:
+            # The tripped guardrail is read straight off the SDK's own exception payload
+            # (exc.guardrail_result.guardrail), the same InputGuardrail instance registered
+            # on the agent above -- no third-party product name, no synthetic identifier.
+            agent_span.set_attribute("gen_ai.agent.decision.outcome", "block")
```
lmolkova: PTAL at the https://github.com/open-telemetry/semantic-conventions-genai/pull/262 - it addresses a general guardrail tracing problem.

AgentGymLeader:
@lmolkova Read #262 — I'll drop the overlap from this PR: `decision.outcome` goes away, and so does stamping decisions on the agent span.

What stays is one thing: a payload-free reference from a decision operation to the producer-owned record that governed it (policy, approval, execution context). Not telemetry about the check that ran — a join to the record behind it. The guardrail span is the natural first attachment, so I'll rework the PR into that shape.

On real instrumentations: openai-agents exposes guardrail results on the run result, and ADK's tool callbacks look like a second candidate. Out-of-process decisions (proxy, approval gateway) would carry the same ref on their own tier's spans.

reference/scenarios/agent-governance/scenario.py
AgentGymLeader: Done in 313c02d: `decision.outcome` and the agent-span attachment are gone. The PR is down to `governance.ref` and marked draft until there's a guardrail span to attach it to.

meshailabs:
Producer-side input, since this is the layer we work on: a control plane in front of agent actions, which is the component that would emit this ref. Two notes on the attribute definition.

First, worth writing the opacity rule into the definition itself, the same way the `.method` tokens landed in #291: the ref is an opaque, producer-scoped token, compared for equality only, and consumers must not parse it or infer meaning from its format. Without that line in the registry, someone eventually standardizes on parsing a popular producer's format, and the payload-free boundary erodes.

Second, for the audit case this attribute exists for, the useful guarantee is that the ref identifies the record as it stood when it governed the decision. Policies get edited and approvals get superseded. If the ref resolves to the current version of a record, a later edit silently changes what an old trace appears to say. A definition note that the ref should be stable to the record state at decision time (a version id, a snapshot id, whatever the producer uses) keeps the join trustworthy for a reviewer. That needs no mandated format and no hash, just the stability expectation stated.

The sequencing also reads right to us: draft until #262, with the ref on the decision operation's span rather than copied onto agent spans.

meshailabs:
The opaque reference now has its first downstream consumer drafting requirements for what sits behind the pointer: OpenSecureAIAlliance/RFCs#13 proposes machine-verifiable approval-to-execution scope binding for the SAFE incident-evidence compact, and frames itself explicitly against this PR's design ("the pointer exists and its contents are undefined," which it treats as the right division of labor). Linking it here so the two discussions stay aware of each other rather than forking. It reads as validation of the decision this PR made: the transport stays payload-free while record semantics evolve where they belong, and nothing in that issue asks this PR to change.

AgentGymLeader:
@meshailabs Both go into the definition text. Opaque, producer-scoped, equality only, don't parse — that's the line that keeps the boundary from eroding, and #291's `.method` is the precedent.

The stability point I had underweighted: a ref that resolves to the current version makes an old trace say what it didn't say at the time. Stating the expectation without mandating a format is the right amount.

Neither is written yet — the registry entry follows the attachment span, which is still open (#262 closed, #427 carries it now).

Thanks for the RFC link; agreed it asks nothing of this PR.

AgentGymLeader:
@lmolkova addressed in 90599be: it instruments the OpenAI Agents SDK's own guardrail runtime (`InputGuardrail` / `Runner.run`), not a hand-rolled gate. The scenario README now says that the `record_ref` key is the scenario's, not a native SDK field. PTAL.

@meshailabs added the stability note you asked for: the reference keeps identifying the record version that was in effect for the operation.

PR #372, opened 2026-07-10
Add new sem conv request attribute `gen_ai.request.previous_response.id`

DylanRussell:
## Description

Add `gen_ai.request.previous_response.id` attribute to reference a previous response or interaction ID when passing continuation context in multi-turn or follow-up operations.

## Motivation

There doesn't exist an attribute that captures this. Both open AI and gemini have support for this:

https://developers.openai.com/api/docs/guides/conversation-state#passing-context-from-the-previous-response

https://ai.google.dev/gemini-api/docs/interactions-overview

## Prototype

Added scenarios

## Checklist


model/gen-ai/registry.yaml
```diff
@@ -179,6 +179,15 @@ attributes:
       which input parameter maps to this attribute.
     examples: ['low', 'medium', 'high']
     stability: development
+  - key: gen_ai.request.previous_response_id
```
```suggestion
  - key: gen_ai.request.previous_response.id
```

to match `gen_ai.response.id` (and since id is a property of previous_response)

lmolkova (commented):
LGTM with a small naming suggestion

model/gen-ai/registry.yaml
DylanRussell: sounds good, done

lmolkova (approved):
LGTM, but please update title and description to reflect changes

DylanRussell:
Done !

model/gen-ai/spans.yaml
```diff
@@ -107,6 +107,9 @@ attribute_groups:
       - ref: gen_ai.output.type
         requirement_level:
           conditionally_required: When applicable and if the request includes an output format.
+      - ref: gen_ai.request.previous_response.id
```
JWinermaSplunk: did we want to add an `invoke_agent` reference scenario? Seems like adding it to `attributes.gen_ai.invoke_agent.common` makes it applicable to both client and internal agent spans, or is there a reason to not add to the agent scenarios?

reference/scenarios/openai/scenario.py
```diff
@@ -247,6 +249,95 @@ def run_responses_compaction_reference(client):
...
+        {
+            "type": "message",
+            "role": "user",
+            "content": "Initial prompt.",
+        }
+    ]
+    initial_response = client.responses.create(
+        model=request_model,
+        input=initial_conversation,
+        store=False,
```
JWinermaSplunk: did we want to set `store=False` here? Would the Responses API retain the previous response and its id if set to false?

DylanRussell: yeah good catch if it's false I believe it makes it stateless // not use response id

model/gen-ai/spans.yaml
DylanRussell: I don't think previous response ID is used by agent frameworks. AFAIK it's only used in the responses/interactions API

JWinermaSplunk (approved):
The one thing I might ask is should we be adding the attribute to `attributes.gen_ai.invoke_agent.common` without a reference scenario, but otherwise looks good!

docs/gen-ai/gen-ai-agent-spans.md
```diff
@@ -504,18 +511,19 @@ Examples: LangChain agents, CrewAI agents.
 | [`gen_ai.request.max_tokens`](/docs/registry/attributes/gen-ai.md) | ![Development](https://img.shields.io/badge/-development-blue) | `Recommended` | int | The maximum number of tokens the model generates for a request. | `100` |
 | [`gen_ai.request.model`](/docs/registry/attributes/gen-ai.md) | ![Development](https://img.shields.io/badge/-development-blue) | `Recommended` If applicable. | string | The name of the GenAI model configured for the agent. [7] | `gpt-4` |
 | [`gen_ai.request.presence_penalty`](/docs/registry/attributes/gen-ai.md) | ![Development](https://img.shields.io/badge/-development-blue) | `Recommended` | double | The presence penalty setting for the GenAI request. | `0.1` |
+| [`gen_ai.request.previous_response.id`](/docs/registry/attributes/gen-ai.md) | ![Development](https://img.shields.io/badge/-development-blue) | `Recommended` [8] | string | The unique identifier of a previous response or interaction used to provide context for the current operation. [9] | `resp_0123456789aBCdef`; `interaction-123` |
```
lmolkova: to @JWinermaSplunk point - we should not add it to internal agent spans - it makes sense on client agents and inference, but there is no(?) use case for internal

DylanRussell: Ack I've removed it from common and put it on client agents. Google's interactions API can be used with agents and supports it so i think it makes sense to leave it there..

PR #374, opened 2026-07-12
Fix usage metrics to provide meaningful aggregation and break down by modality, reasoning, cache usage

### LINKED ISSUE #23: More detailed token usage span attributes and metrics

lmolkova:
Fixes #23, #76, #96. Replaces #197.

Restructures GenAI inference token telemetry so consumers can break usage down by direction, modality, cache status, and phase, and aggregate it correctly.

## Metrics

Two metric families are defined:

Per-modality counters (monotonic; broken down by `gen_ai.token.modality`):
- `gen_ai.client.inference.usage.input_tokens`
- `gen_ai.client.inference.usage.output_tokens`
- `gen_ai.client.inference.usage.cache_read.input_tokens`
- `gen_ai.client.inference.usage.cache_write.input_tokens`
- `gen_ai.client.inference.usage.reasoning.output_tokens`

Per-operation histograms (distributions across operations; not broken down by modality):
- `gen_ai.client.inference.operation.input_tokens`
- `gen_ai.client.inference.operation.output_tokens`

OpenAI refines the applicable instruments with OpenAI-specific response attributes.

Removes the operation-agnostic `gen_ai.client.token.usage` histogram and the `gen_ai.token.type` attribute it required.

See [Design choices behind token metrics](docs/gen-ai/non-normative/token-metrics-design.md) for full rationale on why histograms cannot be broken down by modality and the naming choices.

## Motivation

- **Reasoning tokens** had no attribute or metric.
- **Cache tokens** (read/write) had no metric.
- **Incorrect histogram aggregation**: the old `gen_ai.client.token.usage` mixed input and output in a single instrument keyed by `gen_ai.token.type`, so a sum was only meaningful with a token-type filter. Separate per-direction instruments aggregate correctly by default.

## Follow-up

Embeddings token usage (input-only) is out of scope here and will get a dedicated `gen_ai.client.embeddings.usage.*` instrument in a follow-up. The removed histogram covered embeddings via `operation.name`, so embeddings metrics are temporarily uncovered; embeddings span attributes are unchanged.

## Checklist

See [CONTRIBUTING.md]


model/gen-ai/spans.yaml
```diff
@@ -750,7 +778,7 @@ span_refinements:
         note: >
           Anthropic reports this separately from `input_tokens`.
           This value MUST be added to the Anthropic `input_tokens` to compute `gen_ai.usage.input_tokens`.
-      - ref: gen_ai.usage.cache_creation.input_tokens
+      - ref: gen_ai.usage.cache_write.input_tokens
```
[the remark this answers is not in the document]
lmolkova: cache_creation_input_tokens is anthropic property, not an attribute name

model/gen-ai/spans.yaml
alexmojaki, line 1: This doesn't look like it overlaps with the metrics changes at all, even conceptually or in terms of naming symmetry. Should this be a separate PR?

docs/gen-ai/gen-ai-inference-usage-metrics.md
```diff
@@ -0,0 +1,1145 @@
...
+
+The conventions described in this document are specific to Generative AI client
+applications and describe how inference token usage is reported.
+
+Individual systems may include additional system-specific attributes.
+It is recommended to check system-specific documentation, if available.
+
+Two families of token usage instruments are defined:
+
+- Per-operation **histograms** (`gen_ai.client.inference.usage.operation.*`)
```
alexmojaki: What is meant by "Per-operation" here? Both sets of metrics seem to have the `gen_ai.operation.name` attribute. I thought it maybe meant that `operation` in the metric name was a template field rather than literal, e.g. that the metric might be named `gen_ai.client.inference.usage.chat.input_tokens`, but that doesn't seem right.

Why is the only difference between the names of `gen_ai.client.inference.usage.operation.input_tokens` and `gen_ai.client.inference.usage.input_tokens` the `.operation` part? I don't see what that's trying to communicate or how it relates to the actual difference between the metrics.

docs/gen-ai/gen-ai-inference-usage-metrics.md
```diff
@@ -0,0 +1,1145 @@
...
+  describe the distribution of tokens per inference operation. Input, output,
+  and total are separate instruments so that each can be aggregated without an
+  attribute filter. They carry minimal attributes and are not broken down by
+  modality, cache, or phase.
+- Per-modality **counters** (`gen_ai.client.inference.usage.*`) let consumers
+  aggregate token spend (for example to attribute cost) across all inference
```
alexmojaki: Calculating costs from metrics will typically be inaccurate due to different prices for different request sizes (unless attributes are added for that), the semconv shouldn't recommend trying without mentioning the caveat.

docs/gen-ai/gen-ai-inference-usage-metrics.md
```diff
@@ -0,0 +1,1145 @@
...
+It is recommended to check system-specific documentation, if available.
+
+Two families of token usage instruments are defined:
+
+- Per-operation **histograms** (`gen_ai.client.inference.usage.operation.*`)
+  describe the distribution of tokens per inference operation. Input, output,
+  and total are separate instruments so that each can be aggregated without an
+  attribute filter. They carry minimal attributes and are not broken down by
+  modality, cache, or phase.
+- Per-modality **counters** (`gen_ai.client.inference.usage.*`) let consumers
```
alexmojaki: Why counters?

docs/gen-ai/gen-ai-inference-usage-metrics.md
```diff
@@ -0,0 +1,1145 @@
...
+  aggregate token spend (for example to attribute cost) across all inference
+  operations. Each token category is a separate monotonic counter broken down
+  by [`gen_ai.token.modality`](/docs/registry/attributes/gen-ai.md), so a plain
+  sum of a single metric is always meaningful. The `cache_read`,
+  `cache_write`, `tool`, and `reasoning` counters report subsets of the
```
alexmojaki: Why separate metrics for cache etc but not for modalities?

docs/gen-ai/gen-ai-inference-usage-metrics.md
```diff
@@ -0,0 +1,1145 @@
...
+
+Individual systems may include additional system-specific attributes.
+It is recommended to check system-specific documentation, if available.
+
+Two families of token usage instruments are defined:
+
+- Per-operation **histograms** (`gen_ai.client.inference.usage.operation.*`)
+  describe the distribution of tokens per inference operation. Input, output,
+  and total are separate instruments so that each can be aggregated without an
+  attribute filter. They carry minimal attributes and are not broken down by
```
alexmojaki: Why minimal attributes?

lmolkova: AI slop, sorry

model/gen-ai/spans.yaml
lmolkova: sorry for the delay @alexmojaki. absolutely, ptal https://github.com/open-telemetry/semantic-conventions-genai/pull/440

I'll get back to this one soon

docs/gen-ai/gen-ai-inference-usage-metrics.md
```diff
@@ -0,0 +1,936 @@
...
+| `audio` | Audio tokens. | ![Development](https://img.shields.io/badge/-development-blue) |
+| `image` | Image tokens. | ![Development](https://img.shields.io/badge/-development-blue) |
+| `text` | Text tokens. | ![Development](https://img.shields.io/badge/-development-blue) |
+| `unknown` | The modality is not known. | ![Development](https://img.shields.io/badge/-development-blue) |
+
+<!-- prettier-ignore-end -->
+<!-- END AUTOGENERATED TEXT -->
+<!-- endweaver -->
+
+[DocumentStatus]: https://opentelemetry.io/docs/specs/otel/document-status
```
trask, line 936: `[ExplicitBucketBoundaries]` is used in both histogram sections but never defined in this file, so it renders as literal text. `gen-ai-metrics.md`, where those sections came from, defines it next to `[DocumentStatus]`.

```suggestion
[DocumentStatus]: https://opentelemetry.io/docs/specs/otel/document-status
[ExplicitBucketBoundaries]: https://github.com/open-telemetry/opentelemetry-specification/blob/v1.55.0/specification/metrics/api.md#instrument-advisory-parameters
```

reference/reports/gen-ai-client-inference-usage-operation-input-tokens-metric.md
```diff
@@ -0,0 +1,22 @@
+# Client Inference Operation Input Tokens Metric
```
trask: These new reports have no `> **[Semantic Convention](...)**` line; every other page in `reference/reports/` has one. `SEMCONV_DOC_LINKS` in `reference/src/semconv_genai/report.py` still maps the removed `gen_ai.client.token.usage` to a `gen-ai-metrics.md` anchor this PR deletes, and has no entry for either new metric. (Same for the output-tokens report.)

reference/src/semconv_genai/semconv_model.py
```diff
@@ -54,7 +54,8 @@
 # are tracked here because agent-framework and anthropic emit them; this list
 # records what the reference scenarios emit, not what instrumentations should emit.
 _METRICS = {
-    "gen_ai.client.token.usage": "Client Token Usage",
+    "gen_ai.client.inference.usage.operation.input_tokens": "Client Inference Operation Input Tokens",
```
trask: The comment above still names `gen_ai.client.token.usage`, which this PR removes.

changelog.d/374.enhancement.md
```diff
@@ -0,0 +1 @@
+Add inference token usage counters that let consumers aggregate token spend across operations, broken down by the new `gen_ai.token.modality` attribute and by token category (cached, cache-written, reasoning).
```
trask, line 1: The PR description still lists `gen_ai.client.inference.usage.tool.input_tokens` under the counters, but nothing in the tree defines it.

trask (approved):

docs/gen-ai/gen-ai-inference-usage-metrics.md
alexmojaki: I think I see the intent of this now but I'm still worried that these two names are too easy to mix up. I propose:

```
gen_ai.client.inference.usage.operation.* -> gen_ai.client.inference.usage.*
gen_ai.client.inference.usage.* -> gen_ai.client.inference.usage.detailed.*
```

lmolkova: thanks, I really like it! Pretty consistent, descriptive and not awkward.

I updated the PR to use this naming.

changelog.d/374.enhancement.md
alexmojaki, line 1: i think the metric names should be mentioned (not necessarily in full)

alexmojaki (approved):

docs/gen-ai/non-normative/usage-metrics-design.md
```diff
@@ -0,0 +1,105 @@
+# Design choices behind usage metrics
```
lmolkova, line 1: todo: investigate 'short' and 'long' tokens

trask:
I think this might be a bit more consistent with existing semantic convention metric naming:

## Histograms

- `gen_ai.client.inference.operation.input_tokens`
- `gen_ai.client.inference.operation.output_tokens`

## Counters
- `gen_ai.client.inference.input_tokens`
- `gen_ai.client.inference.output_tokens`
- `gen_ai.client.inference.cache_read.input_tokens`
- `gen_ai.client.inference.cache_write.input_tokens`
- `gen_ai.client.inference.reasoning.output_tokens`

lmolkova:
@trask 

assuming we follow through with #249 and rename `gen_ai.client.operation.duration` to `gen_ai.client.inference.duration`:

| Option 1: `operation` on histograms (prev) | Option 4: `detailed` on counters (current) |
|---|---|
| `gen_ai.client.inference.duration` | `gen_ai.client.inference.duration` |
| `gen_ai.client.inference.time_to_first_chunk` | `gen_ai.client.inference.time_to_first_chunk` |
| `gen_ai.client.inference.time_per_output_chunk` | `gen_ai.client.inference.time_per_output_chunk` |
| `gen_ai.client.inference.operation.input_tokens` | `gen_ai.client.inference.input_tokens` |
| `gen_ai.client.inference.operation.output_tokens` | `gen_ai.client.inference.output_tokens` |
| `gen_ai.client.inference.input_tokens` | `gen_ai.client.inference.detailed.input_tokens` |
| `gen_ai.client.inference.output_tokens` | `gen_ai.client.inference.detailed.output_tokens` |
| `gen_ai.client.inference.cache_read.input_tokens` | `gen_ai.client.inference.detailed.cache_read.input_tokens` |
| `gen_ai.client.inference.cache_write.input_tokens` | `gen_ai.client.inference.detailed.cache_write.input_tokens` |
| `gen_ai.client.inference.reasoning.output_tokens` | `gen_ai.client.inference.detailed.reasoning.output_tokens` |


`operation` would stand out:  
 - `gen_ai.client.inference.duration` histogram
 - `gen_ai.client.inference.operation.input_tokens` histogram
 - `gen_ai.client.inference.input_tokens` counter

We'd then need to keep using operation in `gen_ai.client.inference.operation.duration` to make things look ok:

Histograms:

- `gen_ai.client.inference.operation.duration`
- `gen_ai.client.inference.operation.time_to_first_chunk`
- `gen_ai.client.inference.operation.time_per_output_chunk`
- `gen_ai.client.inference.operation.input_tokens`
- `gen_ai.client.inference.operation.output_tokens`

Counters:

- `gen_ai.client.inference.input_tokens`
- `gen_ai.client.inference.output_tokens`
- `gen_ai.client.inference.cache_read.input_tokens`
- `gen_ai.client.inference.cache_write.input_tokens`
- `gen_ai.client.inference.reasoning.output_tokens`

Mandark-droid:
On the Option 1 / Option 4 naming question, one observation from building both sides of this and then charting the result, because it is the sort of thing that only shows up once a dashboard exists.

## The two options differ in *which* wildcard is safe

Metric backends and dashboards select by prefix constantly. Under either option one family gets a clean selector and the other does not:

| | clean selector | selector that mixes instrument types |
|---|---|---|
| **Option 1 / trask** | `gen_ai.client.inference.operation.*` → histograms only | `gen_ai.client.inference.*` → histograms **and** counters |
| **Option 4** | `gen_ai.client.inference.usage.detailed.*` → counters only | `gen_ai.client.inference.usage.*` → histograms **and** counters |

The failure mode is quiet rather than loud: a panel wildcarding the ambiguous prefix picks up both instrument types, and summing a histogram's count alongside a counter produces a number that looks plausible and is wrong. Neither option prevents that; they differ only in which family you can select safely.

So the question reduces to: which family do people actually wildcard? In our experience it is the counters — the token-family panels are the ones you want as `sum by (modality)` across several instruments, whereas histograms get named individually because each needs its own quantile. On that specific ground **Option 4's `detailed.*` is the more useful of the two clean selectors**, which is the opposite of what I expected before writing it out.

That is a narrow argument and I would not weight it heavily against consistency with existing metric naming, which is a real point in Option 1's favour.

## A smaller one: the instrument type is not inferable from any of the names

`usage`, `detailed` and `operation` are all English words that could plausibly attach to either a distribution or a total. A consumer reading a metric name still cannot tell whether to reach for `sum()` or `histogram_quantile()` without consulting the registry. That is not solvable by naming alone and I am not proposing a fix — but if a `[!NOTE]` in the spec spelled out which of these are counters and which are histograms, it would save the lookup, and that is the mistake most likely to survive into a dashboard.

## Deprecation path for `gen_ai.client.token.usage`

The PR removes the operation-agnostic histogram and `gen_ai.token.type`. Is the intent a clean removal, or a period where both spellings are emitted?

Asking because the span-side rename in #440 (`cache_creation` → `cache_write`) was straightforward to absorb precisely because `OTEL_SEMCONV_STABILITY_OPT_IN=gen_ai/dup` gave a defined way to emit both. There is no equivalent story for a metric whose *instrument type* changes: a producer cannot dual-emit a histogram and a counter under one name, so whatever the window is, it has to be two instrument names for its duration. Worth stating explicitly in the PR, since every producer will hit the same question.

For context on where this comes from: we implemented the merged span-side breakdown from #440 and are holding off on the metric side until the naming here settles, which is why the question is live for us.

docs/gen-ai/non-normative/usage-metrics-design.md
lmolkova: Quick research:

- OpenAI has [short- and long-context pricing](https://developers.openai.com/api/docs/models/gpt-5.4), but no corresponding API property. The pricing band is derivable from the model and total input-token count.
- Anthropic exposes actual pricing dimensions: [`cache_control.ttl` and `usage.cache_creation`](https://platform.claude.com/docs/en/build-with-claude/prompt-caching#mixing-different-ttls), [`usage.service_tier`](https://platform.claude.com/docs/en/api/service-tiers), and [`usage.inference_geo`](https://platform.claude.com/docs/en/manage-claude/data-residency). We will probably need to record these as additional attributes on the detailed counters. service_tier and geo can go on histograms too since there is one of each per request.

Created issue to follow up - #503

lmolkova:
@Mandark-droid 

Let me check that I extracted the main points from your comment. You prefer Option 4 because it gives dashboards a safe prefix for selecting the counters, although none of the proposed names identifies the instrument type. 
Separately, you are asking for the migration path from `gen_ai.client.token.usage` and `gen_ai.token.type`. 

Is that accurate?

There is no backward-compatibility mechanism. All GenAI semantic conventions are currently in development and may change, so this is a clean replacement. We're going to kick off stabilization effort, after which breaking changes would require at least a major version bump and would likely be rare.

Mandark-droid:
Half accurate, and the half that is off is worth correcting since you may be counting positions.

**The migration question: yes, exactly that, and your answer settles it.** Clean replacement is actually the easier outcome for a producer. A transition window would have been worse here than for an attribute rename, because you cannot dual-emit a histogram and a counter under one name — any window would have forced two instrument names to coexist anyway. Knowing it is a clean swap means we just change names on the next release.

**The preference: I would not record us as preferring Option 4.** What I meant is narrower. On the single axis of wildcard safety, Option 4's `detailed.*` is the more useful of the two clean selectors, because counters are what people actually wildcard. I also said that is a narrow argument I would not weight against consistency with existing metric naming, which is a real point for Option 1 and, I would guess, the more important one.

So our position is closer to: no strong preference between the two, plus one consideration that had not been raised. If it helps to have it as a single sentence — either option is fine to implement against, and I would let naming consistency decide rather than the wildcard point.

The instrument-type observation was separate and applies to both options equally, so it should not move the choice either way.

**One thing your answer raises.** If there is no backward-compatibility mechanism and everything can change until stabilization, then the cost of a rename lands entirely on producers noticing it happened. We did not notice #440 renaming `cache_creation` to `cache_write` for over two weeks, and spent that time emitting a name no conforming consumer reads. That is our failure to watch, not a spec problem, and we have since automated the check.

But it does suggest one cheap thing: renames are much easier to catch when the changelog entry names both the old and new key literally, so a producer can grep their own codebase for the old string. #440's entry does this well. Worth keeping as a habit through stabilization, because the failure mode is silent on the producer side — nothing errors, the attribute simply stops being read.

Is there a rough timeline for the stabilization effort? That is the point at which it becomes worth pinning to a schema version rather than tracking `main`.

lmolkova:
@Mandark-droid can you please express it in concise and human readable way?

Mandark-droid:
Sorry, that was too long.

On migration, yes, that's accurate, and a clean swap is fine for us. We'd just rename on our next release.

On naming though, I don't actually prefer Option 4. The wildcard argument was minor and I'd rather consistency with the existing metric names decide. We can implement either.

Any rough timeline for stabilisation? That's when we'd pin a schema version instead of tracking main.

lmolkova:
the plan is to announce release candidate by KubeCon NA 2026 (Nov)  - https://cloud-native.slack.com/archives/C06KR7ARS3X/p1788297280304949

AgentGymLeader:
Non-blocking observation: when testing with synthetic responses where both candidate and thought counts (`candidates_token_count` and `thoughts_token_count`) are 0, the Google reference drops the output usage attribute. For outputs of 0 and 10, the histogram recorded a count of 1 and a sum of 10.

This truthiness check predates this PR, so I am happy to open a separate PR with a small fix and regression test to keep zero distinct from missing usage.

lmolkova:
Notes from the SIG discussion:

- `detailed` is confusing; its primary purpose is separating counters from histograms.
- `usage` in the namespace might be unnecessary and has a minor conflict with [semconv instrument naming guidance](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/general/naming.md#instrument-naming).
- Counters are the primary use case (tracking spend and aggregation), while histograms are mostly for tracking outliers. Counter names should be prioritized.

Options discussed:

**Option A (drop `usage`)**:
- Histogram: `gen_ai.client.inference.input_tokens`
- Counter: `gen_ai.client.inference.detailed.input_tokens`
- Pros: shorter. Cons: still confusing, prioritizes histograms.

**Option B (`usage` in counters)**:
- Histogram: `gen_ai.client.inference.input_tokens`
- Counter: `gen_ai.client.inference.usage.input_tokens`
- Cons: still confusing, still prioritizes histograms.

**Option C (new word for `detailed`)**:
- Histogram: `gen_ai.client.inference.input_tokens`
- Counter: `gen_ai.client.inference.<qualifier>.input_tokens`
- Cons: still prioritizes histograms. We don't have a good word

**Option D**:
- Histogram: `gen_ai.client.inference.input_tokens_per_operation` / `_per_inference`
- Counter: `gen_ai.client.inference.input_tokens`
- Pros: histograms are clearly separated. Cons: `gen_ai.client.inference.input_tokens` next to `gen_ai.client.inference.duration` looks like a histogram.

**Current winner**:
- Counter: `gen_ai.client.inference.usage.input_tokens`
- Histogram: `gen_ai.client.inference.operation.input_tokens`
- `operation` clarifies per-operation distribution for histograms, while `usage` names the counters.

I'm going to update the PR with it. 

@alexmojaki mind giving it another look? thanks!

aabmass:
I am OK with pretty much any option since OTel doesn't have general guidance. Had some questions though

> * `operation` clarifies per-operation distribution for histograms, while `usage` names the counters.

Is using `operation` intentionally reminiscent of `gen_ai.*.operation.duration` metrics and `gen_ai.operation.name` attribute key? If we are planning to deprecate that metric, I suppose it's moot anyway.

> * Cons: still confusing, still prioritizes histograms.

What is bad about prioritizing the histograms?

> * Counter: `gen_ai.client.inference.<qualifier>.input_tokens`

I'm happy with like `usage` or maybe `total` (which is used in prometheus ecosystem). One more idea, the [Prometheus mapping](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/compatibility/prometheus_and_openmetrics.md) naturally disambiguates histograms from counters with suffixes. So the proposed metrics would look like this in prometheus:

- Counter
  - `gen_ai_client_inference_usage_input_tokens_total`
- Hist
  - `gen_ai_client_inference_operation_input_tokens_bucket`
  - `gen_ai_client_inference_operation_input_tokens_sum`
  - `gen_ai_client_inference_operation_input_tokens_count`

docs/gen-ai/non-normative/token-metrics-design.md
```diff
@@ -0,0 +1,106 @@
+# Design choices behind token metrics
```
aabmass, line 1: Would be nice to include a TLDR. The thing that made it click for me is "it is invalid to record more than one observation per operation that a histogram represents"

aabmass (approved):

trask:
> > * Cons: still confusing, still prioritizes histograms.
> 
> What is bad about prioritizing the histograms?

Option B is also nice I think

- Histogram: `gen_ai.client.inference.input_tokens`
- Counter: `gen_ai.client.inference.usage.input_tokens`

vs

Current winner

- Histogram: `gen_ai.client.inference.operation.input_tokens`
- Counter: `gen_ai.client.inference.usage.input_tokens`

docs/gen-ai/non-normative/token-metrics-design.md
lmolkova: added!

lmolkova:
> What is bad about prioritizing the histograms?

@aabmass they are only there to show outliers. The first stop is usage (how much I'm spending and what contributes to it - specific model, modality, reasoning, etc)


> Option B is also nice I think

I like the symmetry with having namespace for usage and histograms. It's not a strong preference, but I'd prefer to keep the current one.

@alexmojaki I'd love to get your take on the current shape of it and how we end up here - https://github.com/open-telemetry/semantic-conventions-genai/pull/374#issuecomment-5674493388. I'd like to merge this PR by EoD (pacific time) Monday unless you need extra time or have objections. LMK and thank you!

AgentGymLeader:
Filed the explicit-zero fix I mentioned earlier as a separate PR: #529. It only touches `_usage_attributes` in the google-genai scenario; #374 reads the attribute with `is not None` and is unaffected.

trask (approved):

reference/scenarios/openai/scenario.py
```diff
@@ -7,13 +7,103 @@
...
+    output_tokens = usage.get("gen_ai.usage.output_tokens")
+    if output_tokens is not None:
+        _operation_output_tokens.record(output_tokens, metric_attributes)
+        _add_by_modality(
+            _output_tokens, output_tokens, usage.get("gen_ai.usage.audio.output_tokens"), metric_attributes
+        )
+    for attribute, counter in (
+        ("gen_ai.usage.cache_read.input_tokens", _cache_read_input_tokens),
+        ("gen_ai.usage.cache_write.input_tokens", _cache_write_input_tokens),
+        ("gen_ai.usage.reasoning.output_tokens", _reasoning_output_tokens),
```
aabmass, line 102: Should we explicitly document how `0` values work for the new counters vs histograms?

Right now the scenarios in this PR do different things: `anthropic` and `google-genai` only call `counter.add()` when `> 0`, but `openai/scenario.py` checks `if value is not None:` here for `cache_read`, `cache_write`, and `reasoning`. Since OpenAI returns `cached_tokens=0` and `reasoning_tokens=0` by default, `openai` ends up calling `add(0)` (and setting `0` on the span) on every request.

aabmass: Happy to file this as a follow up if you prefer to merge right away @trask

AgentGymLeader:
From #529: when `usageMetadata` is returned but `candidatesTokenCount` is omitted (proto3 zero), is `gen_ai.usage.output_tokens` 0 or unset? A sentence in the notes would settle it for the reference scenarios.

reference/scenarios/openai/scenario.py
AgentGymLeader: Hit the same question in #529: the Gemini API omits a zero `candidatesTokenCount` (plain proto3 int32), so for google-genai neither `> 0` nor `is not None` ever records a 0 from a real response — the difference only shows up with providers like OpenAI that send explicit zeros. #529 is on hold until this is decided here.

lmolkova: good points, I think it's not straightforward for histograms, created https://github.com/open-telemetry/semantic-conventions-genai/issues/530 to track. #502 tracks attributes

PR #383, opened 2026-07-16
Sync stale core semconv version references with the versions.env pin

ANcpLua:
## What

`versions.env` is the single source of truth for the pinned core semantic-conventions version (`SEMCONV_VERSION`, currently `v1.43.0`, tracked by Renovate). Two hand-maintained references to that version were left behind by the last bump and this PR brings them back in sync:

- **`model/manifest.yaml`** — the filtered upstream dependency still declared `https://opentelemetry.io/schemas/1.41.0` (in both the comment and the `schema_url`), while the Makefile actually clones the tag pinned in `versions.env` into `.build/sc-upstream-filtered`. The comment now points at `versions.env` as the source of truth instead of hardcoding a tag, and the `schema_url` is bumped to `1.43.0`.
- **`AGENTS.md`** — the four "Further reading" links pointed at `blob/v1.42.0`; bumped to `blob/v1.43.0`. All four targets verified reachable at that tag (HTTP 200).

Also included: a whitespace normalization in `docs/gen-ai/gen-ai-metrics.md` that `make generate-all` produces on current `main` (the committed generated output had drifted from `model/gen-ai/metrics.yaml`).

## Why this happened

- #342 (2026-06-26) added `AGENTS.md` with `v1.42.0` links — current at the time.
- #366 (2026-07-04) bumped `SEMCONV_VERSION` to `v1.43.0`; Renovate's custom manager only tracks `versions.env`, and `make generate-all` only regenerates the generated docs — so the hand-written `AGENTS.md` links and the `manifest.yaml` dependency declaration were not updated.

No behavior change: the build already uses `v1.43.0`; this only makes the declared metadata and docs match reality. If it's useful, a follow-up could add Renovate regex managers for these two spots so they can't drift again.

`make check-policies` and `make generate-all` pass locally with no further diff.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -59,7 +59,7 @@ changes, not a skill to invoke.
 
 ## Further reading
 
-- [How to write conventions — best practices](https://github.com/open-telemetry/semantic-conventions/blob/v1.42.0/docs/how-to-write-conventions/README.md#best-practices)
-- [Naming](https://github.com/open-telemetry/semantic-conventions/blob/v1.42.0/docs/general/naming.md)
-- [Events](https://github.com/open-telemetry/semantic-conventions/blob/v1.42.0/docs/general/events.md)
-- [Recording errors](https://github.com/open-telemetry/semantic-conventions/blob/v1.42.0/docs/general/recording-errors.md)
+- [How to write conventions — best practices](https://github.com/open-telemetry/semantic-conventions/blob/v1.43.0/docs/how-to-write-conventions/README.md#best-practices)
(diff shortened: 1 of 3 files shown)


lmolkova (approved):

lmolkova:
@ANcpLua thanks! please sign EasyCLA and we should be good to go

ANcpLua:
EasyCLA sorted — the commit was authored with a stale email that GitHub resolves to an old account of mine, so the CLA check was matching the wrong identity. Amended author+committer to the right email; all checks are green now.

PR #394, opened 2026-07-23
Realtime / live voice model generation (generate_live_content span)

singankit:
## What this PR proposes

Instrumentation conventions for **realtime, bidirectional speech-to-speech (voice) model sessions** such as the OpenAI Realtime API and Google Gemini Live. A live session streams user audio continuously while the server streams incremental output, so a generation cannot be represented by the client-timed `inference` span. This PR models it explicitly:

- **New `gen_ai.generate_live_content.client` span** (+ `generate_live_content` operation) representing one server-side generation reconstructed from the provider's streaming events. It starts at the first output chunk (or an earlier voice-activity end anchor) and ends at generation completion/interruption.
- **Live-session lifecycle events** — `gen_ai.client.live_session.started` / `gen_ai.client.live_session.ended`. The session itself is long-lived and mostly idle, so it is modeled as events, not a span; generations in a session are correlated by `gen_ai.conversation.id`.
- **Optional `gen_ai.user_input.client` span** (+ `user_input` operation, `Opt-In`) for the user's utterance, bracketed by provider **voice-activity (VAD)** events. Only recordable when the provider exposes user VAD events (e.g. OpenAI Realtime `input_audio_buffer.speech_started`/`speech_stopped`); providers without them (Gemini Developer API) carry the user input on the generation span via `gen_ai.input.messages`.
- **Audio token usage attributes** referenced on the generation span:
  - `gen_ai.usage.audio.input_tokens`, `gen_ai.usage.audio.output_tokens` (each SHOULD be included in the corresponding `gen_ai.usage.*_tokens` total).
- **Audio `transcript` field** on audio message parts (`BlobPart`, `FilePart`, `UriPart`) in `docs/gen-ai/non-normative/models.py`, so the spoken-audio transcript rides on the audio part itself instead of a separate ambiguous text part. Regenerates the message JSON schemas.
- **Tool-call span shape** guidance for live generations — **sibling** pattern (OpenAI Realtime, Gemini Live 2.5) vs **child** pattern (Gemini Live 3.x).
- **Documentation**: trace-structure diagrams, an "other voice-agent vendors" survey (ElevenLabs, PipeCat), and an **Open questions** section for aspects intentionally left unresolved (session/turn modeling, grouping generations by user input, continuous-generation models, usage attribution).
- Changelog fragment.

## Motivation

Realtime / speech-to-speech APIs (OpenAI Realtime `gpt-realtime`, Google Gemini Live) are a distinct and growing class of GenAI workload whose streaming, session-oriented shape doesn't fit the request/response `inference` span. This PR introduces the minimal, provider-neutral surface to capture them — a dedicated generation span, session lifecycle events, an optional VAD-gated user-input span, and audio token/transcript signals — validated against multiple independent providers so the conventions generalize beyond a single SDK.

## Reference scenarios (capturability proof)

| Scenario | Provider | Demonstrates |
| --- | --- | --- |
| `openai-realtime` | OpenAI Realtime API (`gpt-realtime`, websockets) | `generate_live_content` + session events + **`user_input` span** (VAD) + **sibling** tool call + audio tokens/transcript |
| `gemini-live` | Google Gemini Live (`client.aio.live.connect`) | `generate_live_content` + session events + **sibling** tool call; honest capture gaps (no user VAD → no `user_input`; no session id → no `conversation.id`) |
| `gemini-live-3` | Google Gemini Live 3.x | **child** tool-call pattern (nested `execute_tool` within one generation) |

## Provider mapping: audio & transcript events

Both realtime providers stream audio and its transcript over a bidirectional WebSocket with different wire shapes; generic instrumentation normalizes both onto the same convention (one `transcript` field on the audio part + audio-token attributes on the generation span).

<details>
<summary>OpenAI Realtime vs Gemini Live — how audio &amp; transcript are delivered</summary>

| Signal | OpenAI Realtime | Gemini Live |
| --- | --- | --- |
| Transcript (incremental) | `response.output_audio_transcript.delta` → `delta` | `serverContent.outputTranscription.text` (nested) |
| Transcript (final, full) | `response.output_audio_transcript.done` → `transcript` | *(none — concatenate the deltas)* |
| Audio | `response.output_audio.delta` → `delta` (base64) | `serverContent.modelTurn.parts[].inlineData.data` (SDK decodes) |
| Turn end + usage | `response.done` → `response.usage` | `serverContent.turnComplete` + `usageMetadata` |
| Audio-token usage | `input/output_token_details.audio_tokens` | per-modality `promptTokensDetails`/`responseTokensDetails` (`AUDIO` entry) |
| Event identity | distinct top-level `type` per signal (snake_case) | one `serverContent` envelope; signal = which nested key is set (camelCase) |

</details>

## Relationship to other PRs

- The audio-token attributes are **shared with #393** (openai-audio / chat-completions audio) and intentionally duplicated so each PR is independently mergeable; whichever lands first, the other drops the duplicate.
- This is the realtime/live slice extracted from the voice-agents proposal (#390).

## Merge requirements

- [x] Changelog fragment (`changelog.d/`)
- [x] Reference scenarios demonstrating capturability (`openai-realtime`, `gemini-live`, `gemini-live-3`)
- [x] Generated docs/reports refreshed (`make generate-all`)
- [x] DCO sign-off


reference/scenarios/gemini-live-3/mock_key.pem
```diff
@@ -0,0 +1,27 @@
+-----BEGIN RSA PRIVATE KEY-----
```
lmolkova: needed? maybe there is a way to disable cert validation with API/env var without adding fake certs?

reference/scenarios/gemini-live/scenario.py
```diff
@@ -0,0 +1,195 @@
...
+                        "modality": "audio",
+                        "mime_type": "audio/pcm",
+                        "content": INPUT_AUDIO_B64,
+                    }
+                ],
+            }
+        ]
+    )
+    chat_attributes = {
+        "gen_ai.operation.name": "chat",
```
lmolkova: it's not really a chat, is it? I think we need a new span type for it

reference/scenarios/openai-realtime/scenario.py
```diff
@@ -0,0 +1,164 @@
...
+        "gen_ai.request.model": request_model,
+        # The realtime session produces speech output for this turn.
+        "gen_ai.output.type": "speech",
+    }
+    if host:
+        chat_attributes["server.address"] = host
+    if port is not None:
+        chat_attributes["server.port"] = port
+
+    with _reference_tracer.start_as_current_span(f"chat {request_model}", attributes=chat_attributes) as chat_span:
```
lmolkova: how would instrumentation know when to start and end this span? when session is established and when it's closed?

reference/scenarios/openai-realtime/scenario.py
```diff
@@ -0,0 +1,164 @@
...
+        chat_span.set_attribute("gen_ai.input.messages", input_messages)
```
lmolkova: I assume we'd need to accumulate different chunks in one blob part or would we emit a blob part per chunk?

reference/scenarios/openai-realtime/scenario.py
```diff
@@ -0,0 +1,164 @@
...
+        output_audio = "".join(audio_deltas)
+        output_messages = [
+            {
+                "role": "assistant",
+                "parts": [
+                    {
+                        "type": "blob",
+                        "modality": "audio",
+                        "mime_type": "audio/pcm",
+                        "content": output_audio,
```
lmolkova: same, should we concatenate individual chunks in one blob or keep them as independent blob parts?

I see you're concatenating, but would like to understand if it's a deliberate choice and other options were considered

reference/scenarios/openai-realtime/scenario.py
```diff
@@ -0,0 +1,164 @@
...
+                        "transcript": transcript,
```
lmolkova: this would only apply for audio blobs and only when transcribing. Should we have a separate part BlobWithTranscription or AudioPart that implies audio modality? It seems weird to have transcript on image or document parts.

reference/scenarios/gemini-live/scenario.py
```diff
@@ -0,0 +1,195 @@
...
+        "gen_ai.request.model": request_model,
+        # The live session produces speech output for this turn.
+        "gen_ai.output.type": "speech",
+    }
+    if host:
+        chat_attributes["server.address"] = host
+    if port is not None:
+        chat_attributes["server.port"] = port
+
+    with _reference_tracer.start_as_current_span(f"chat {request_model}", attributes=chat_attributes) as chat_span:
```
lmolkova: I assume instrumentation would start this span when session is established and end it when it closes similarly to openai case?

is there anything important to capture about session itself (here or in openai)?

lmolkova (commented):
Thanks for the PR!

I left a few questions, my main one is whether we should use inference | chat span to describe realtime. 

The API we cover has very different shape than chat / responses API, parameters seem to be different, we might want to call operations "translation" or "transcription" instead of chat and so on. I think we need a new span type for it and definitely some new operation names. We can reuse attributes and message types as much as makes sense, but  we should limit this new span to only the attributes that are applicable to it.

reference/scenarios/gemini-live/scenario.py
singankit: For realtime models sessions can be long (openai allows upto 1 hour). Within this session there can be multiple back and forth between model and user as conversation happens. This proposal is to model the turns as invoke_agent span like we do for text models. Whether there should be a span that models the long running sessions is still an open question.

reference/scenarios/openai-realtime/scenario.py
singankit: This is similar to streaming case in text models. We should handle it the same way would be my suggestion. And it should be for each turn and not entire sesssion.

reference/scenarios/openai-realtime/scenario.py
singankit: `BlobPart`, `UriPart` and `FilePart` already have `modality` field, can we not re-use it and have schema update to allow `transcript` only when modality is `audio` ?

reference/scenarios/openai-realtime/scenario.py
singankit: Response from realtime model is sent back in chunks, similar to streaming case for text models (conceptually). This sample is putting together all the chunks so that we have a complete audio response for the turn (not for s session.)

reference/scenarios/gemini-live-3/mock_key.pem
singankit: Thanks will look into it.

reference/scenarios/gemini-live/scenario.py
singankit: Gathered more information to help decide whether a new span vs. the existing chat span would work — full write-up here: https://github.com/singankit/semantic-conventions-genai/blob/users/singankit/voice-inference-realtime/docs/gen-ai/non-normative/realtime-vs-chat-span-design.md

singankit:
Following up on your review (#pullrequestreview-4801878547) on whether realtime should reuse the chat/inference span or get a new span type: I gathered more information to help decide new-vs-existing span. Full write-up (non-normative design doc, all API claims verified against the installed OpenAI and google-genai SDKs): https://github.com/singankit/semantic-conventions-genai/blob/users/singankit/voice-inference-realtime/docs/gen-ai/non-normative/realtime-vs-chat-span-design.md

model/gen-ai/spans.yaml
```diff
@@ -44,8 +44,14 @@ attribute_groups:
...
       - ref: gen_ai.usage.cache_creation.input_tokens
         requirement_level: recommended
+      - ref: gen_ai.usage.audio.input_tokens
+        requirement_level:
+          recommended: When the model reports audio input tokens separately.
       - ref: gen_ai.usage.output_tokens
         requirement_level: recommended
+      - ref: gen_ai.usage.audio.output_tokens
+        requirement_level:
+          recommended: When the model reports audio output tokens separately.
```
DylanRussell: https://github.com/open-telemetry/semantic-conventions-genai/pull/440 -- i think this PR adds these, so no need to do it here

model/gen-ai/spans.yaml
```diff
@@ -274,6 +277,155 @@ spans:
       - ref: gen_ai.request.model
         sampling_relevant: true
 
+  - type: gen_ai.generate_live_content.client
+    requirement_level: recommended
+    kind: client
+    name:
+      note: |
+        **Span name** SHOULD be `{gen_ai.operation.name} {gen_ai.request.model}`.
```
DylanRussell, line 278: won't this always just be the constant `generate_live_content` ? If so can we just specify that directly ? Maybe you are following the existing pattern but it is weird

model/gen-ai/spans.yaml
```diff
@@ -274,6 +277,155 @@ spans:
...
+        and MUST follow the overall [guidelines for span names](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.56.0/specification/trace/api.md#span).
+    brief: >
+      Describes one server-side model generation within a live, bidirectional speech-to-speech
+      session (such as OpenAI Realtime or Gemini Live).
+    note: |
+      Unlike `gen_ai.inference.client`, this span is not a single client-timed request/response.
+      The client streams user audio continuously and the server streams incremental events; the
+      generation is reconstructed from those events. The span SHOULD start at the first output
+      chunk of the generation (or at an earlier voice-activity end anchor when the provider
+      exposes one) and end at the generation completion or interruption event. If usage is
```
DylanRussell: "if usage is reported afet..."

What does "usage" mean ?

model/gen-ai/spans.yaml
```diff
@@ -274,6 +277,155 @@ spans:
...
+
+      The `gen_ai.operation.name` SHOULD be `generate_live_content`.
+
+      The live session itself is not modeled as a span. It is represented by the
+      `gen_ai.client.live_session.started` and `gen_ai.client.live_session.ended` events and by
+      `gen_ai.conversation.id` correlating the generations within it. Turn-level spans (a full
+      user-and-model exchange) are out of scope: turn boundaries cannot be detected reliably
+      across providers.
+
+      When a generation calls a tool, the relationship between the generation and the
```
DylanRussell: "When a generation" 

When a model generation ? What is a "generation"

changelog.d/448.enhancement.md
```diff
@@ -0,0 +1 @@
+Add a `generate_live_content` operation and the `gen_ai.generate_live_content.client` span to model per-generation server-side output in live, bidirectional speech-to-speech sessions (such as OpenAI Realtime and Gemini Live), together with `gen_ai.client.live_session.started`/`gen_ai.client.live_session.ended` events for session lifecycle, an optional VAD-gated `user_input` operation and `gen_ai.user_input.client` span for the user's utterance, and guidance on sibling versus child tool-call span shapes. Reference scenarios cover OpenAI Realtime and Gemini Live (sibling pattern) and Gemini Live 3.x (child pattern).
```
DylanRussell: what is PR #448 ?

model/gen-ai/gen-ai-output-messages.json
```diff
@@ -39,6 +40,19 @@
...
+                    "anyOf": [
+                        {
+                            "type": "string"
+                        },
+                        {
+                            "type": "null"
+                        }
+                    ],
+                    "default": null,
+                    "description": "The text transcript of the attached audio. Only applicable when `modality` is `audio`. For output audio this is the model-generated transcript of the spoken response; for input audio it is the transcription of the provided audio.",
```
DylanRussell: so the instrumentation library will needs to do some conversion of audio files to text in order to populate this ?

https://ai.google.dev/gemini-api/docs/live-api/get-started-sdk -- looks like for gemini live the input from the user could be text, audio or video, and the output from the server can be audio or text.. Have you given any thought on if and how to capture video/audio content  in telemetry?

model/gen-ai/spans.yaml
```diff
@@ -274,6 +277,155 @@ spans:
...
+      chunk of the generation (or at an earlier voice-activity end anchor when the provider
+      exposes one) and end at the generation completion or interruption event. If usage is
+      reported after the completion event, instrumentation SHOULD wait for it before ending the
+      span.
+
+      The `gen_ai.operation.name` SHOULD be `generate_live_content`.
+
+      The live session itself is not modeled as a span. It is represented by the
+      `gen_ai.client.live_session.started` and `gen_ai.client.live_session.ended` events and by
+      `gen_ai.conversation.id` correlating the generations within it. Turn-level spans (a full
```
DylanRussell: For gemini there is also the concept of a resumption token: https://ai.google.dev/gemini-api/docs/live-api/session-management#session-resumption (that whole page is probably worth reading BTW).. I think that would be a separate thing from a conversation id and maybe should be it's own attribute ?

docs/gen-ai/gen-ai-spans.md
```diff
@@ -398,16 +408,692 @@ and SHOULD be provided **at span creation time** (if provided at all):
...
+
+#### Trace structure
+
+A live session is bracketed by the `started` / `ended` events (not a span), and
+each server-side generation is a `gen_ai.generate_live_content.client` span
+correlated by `gen_ai.conversation.id`. The `gen_ai.conversation.id` is the
+provider session identifier when one is available; when the provider exposes no
+session identifier (for example, the Gemini Developer API `setupComplete`
+message has no fields), instrumentation may instead mint a client-created
+per-connection id — the live connection is the session boundary — and use it as
```
lmolkova: `gen_ai.conversation.id`  is orthogonal to live sessions. I can have multiple live sessions with the agent on a single conversation. 

I think we need a new attribute (e.g. `gen_ai.live_session.id`)

model/gen-ai/events.yaml
```diff
@@ -1,5 +1,57 @@
...
+      - ref: server.port
+        requirement_level:
+          conditionally_required: If `server.address` is set.
+  - name: gen_ai.client.live_session.ended
+    requirement_level: recommended
+    brief: >
+      Recorded when a live, bidirectional speech-to-speech session is closed.
+    note: >
+      Emitted when the session connection is closed, whether normally or because of an error.
+      Paired with the `gen_ai.client.live_session.started` event through
```
lmolkova: Pairing by `gen_ai.conversation.id` doesn't work for a provider you shipped a scenario for. It is conditionally required, and Gemini Live exposes no session id - the `gemini-live` scenario emits `started`/`ended` with only provider, model and `server.*`. Those two events cannot be paired with each other, and the generations cannot be attributed to the session.

Either the session needs an identifier that is always available, or the session should be a span (which gives correlation via context for free).

model/gen-ai/spans.yaml
```diff
@@ -270,6 +270,157 @@ spans:
...
+        Semantic conventions for individual GenAI systems and frameworks MAY specify different span name format
+        and MUST follow the overall [guidelines for span names](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.56.0/specification/trace/api.md#span).
+    brief: >
+      Describes one server-side model generation within a live, bidirectional speech-to-speech
+      session (such as OpenAI Realtime or Gemini Live).
+    note: |
+      Unlike `gen_ai.inference.client`, this span is not a single client-timed request/response.
+      A *generation* is one server-side model response within the live session (the model's
+      spoken or textual turn). The client streams user audio continuously and the server streams
+      incremental events; the generation is reconstructed from those events. The span SHOULD start
```
lmolkova, line 288: Both `openai-realtime` and `gemini-live` start the span when the client finishes sending input (`response.create` / after `audio_stream_end`), not at the first output chunk, and they compute `time_to_first_chunk` from a timestamp taken before span start. 

We probably should drop time to first chunk in live scenarios unless there is a good way to know when model started processing

model/gen-ai/spans.yaml
```diff
@@ -270,6 +270,157 @@ spans:
...
+          recommended: When available. A live generation may not carry a response id on all providers.
+      - ref: gen_ai.response.model
+        requirement_level: recommended
+        note: >
+          If available. The name of the GenAI model that provided the response. If the model is supplied by a vendor,
+          then the value must be the exact name of the model actually used. If the model is a
+          fine-tuned custom model, the value should have a more specific name than the base model that's been fine-tuned.
+      - ref: gen_ai.response.finish_reasons
+        requirement_level: recommended
+      - ref: gen_ai.response.time_to_first_chunk
```
lmolkova: `gen_ai.response.time_to_first_chunk` is defined as "measured from when the client issues the generation request". There is no client-issued request here, and with the start rule above (span starts at the first output chunk) the value is 0 by construction. Drop it, or define what the measurement is anchored to for a live generation.

model/gen-ai/spans.yaml
```diff
@@ -270,6 +270,157 @@ spans:
...
+      - ref: gen_ai.operation.name
+        sampling_relevant: true
+      - ref: server.address
+        sampling_relevant: true
+      - ref: server.port
+        sampling_relevant: true
+      - ref: gen_ai.request.model
+        sampling_relevant: true
+
+  - type: gen_ai.user_input.client
```
lmolkova: `kind: client` doesn't fit a span whose duration is the user speaking - no outbound call is bounded by it. `server.address`/`server.port`/`gen_ai.request.model` on it describe the session, not this operation; if they are only there to correlate back to the session, that is an argument for a session identifier instead.

reference/scenarios/openai-realtime/conformance.yaml
```diff
@@ -0,0 +1,34 @@
...
+    run: uv run --frozen --python 3.12 --project . python scenario.py
+    # The realtime / live voice operations are not yet recognised by the shared
+    # GenAI conformance classifier, so declare their span types here: an emitted
+    # INTERNAL span carrying the given operation name is the model's client span
+    # type. Remove once the classifier recognises these operations upstream.
+    spans:
+      - match:
+          attributes:
+            gen_ai.operation.name: generate_live_content
+          kind: internal
```
lmolkova: Both scenarios emit these as INTERNAL spans while the model declares `kind: client`, and this mapping papers over the mismatch rather than proving the convention. Fix the kind in one place or the other - as it stands nothing in the reference validates the span kind you are shipping.

model/gen-ai/spans.yaml
```diff
@@ -270,6 +270,157 @@ spans:
       - ref: gen_ai.request.model
         sampling_relevant: true
 
+  - type: gen_ai.generate_live_content.client
```
lmolkova: `generate_live_content` is Gemini's terminology and uses operation name which is long term should be idiomatic to the vendor.

Suggestions:
- gen_ai.realtime_inference.client
- gen_ai.live_inference.client

reference/scenarios/openai-realtime/scenario.py
```diff
@@ -0,0 +1,361 @@
...
+    when the user was speaking. The span wraps that interval.
+    """
+    attributes = _live_attributes(provider, request_model, session_id, host, port, "user_input")
+    with _reference_tracer.start_as_current_span("user_input", attributes=attributes) as span:
+        # The user speaks: append their audio and commit it. The server responds
+        # with the voice-activity events that delimit the utterance.
+        conn.send({"type": "input_audio_buffer.append", "audio": INPUT_AUDIO_B64})
+        conn.send({"type": "input_audio_buffer.commit"})
+        for event in conn:
+            if event.type == "input_audio_buffer.committed":
```
lmolkova: This doesn't demonstrate the thing the `user_input` span is defined by. The convention says start at `input_audio_buffer.speech_started` and end at `speech_stopped`; here the span wraps the app's own `append`/`commit` and ends at `input_audio_buffer.committed`. With manual commit there are no server VAD events at all, so a VAD-bracketed span remains unproven - and the boundaries actually used here are app calls, which generic instrumentation of a real streaming client would not see the same way.

docs/gen-ai/gen-ai-spans.md
```diff
@@ -398,16 +408,692 @@ and SHOULD be provided **at span creation time** (if provided at all):
...
+**[43]:** May be used when specific backend is unknown.
 
-**[42]:** Used when accessing the 'aiplatform.googleapis.com' endpoint.
+**[44]:** Used when accessing the 'aiplatform.googleapis.com' endpoint.
 
 <!-- prettier-ignore-end -->
 <!-- END AUTOGENERATED TEXT -->
 <!-- endweaver -->
 
+### Live content generation
```
lmolkova: please move it to a different file and link it from here

docs/registry/attributes/gen-ai.md
```diff
@@ -394,9 +395,14 @@ what `gen_ai.workflow.name` means in the context of that framework.
 | `text_completion` | Text completions operation such as [OpenAI Completions API (Legacy)](https://platform.openai.com/docs/api-reference/completions) | ![Development](https://img.shields.io/badge/-development-blue) |
 | `update_memory` | Update existing memory records | ![Development](https://img.shields.io/badge/-development-blue) |
 | `upsert_memory` | Create or update memory records without the caller choosing which | ![Development](https://img.shields.io/badge/-development-blue) |
+| `user_input` | Client-side capture of a user input (utterance) within a live, bidirectional speech-to-speech session, delimited by the provider's user voice-activity (start/stop) events. [48] | ![Development](https://img.shields.io/badge/-development-blue) |
```
lmolkova: live sessions are multi-modal, so we should either call it record | receive_user_speech or relax it to cover other modalities. Live can even be used for text

model/gen-ai/events.yaml
```diff
@@ -1,5 +1,57 @@
 file_format: definition/2
 events:
+  - name: gen_ai.client.live_session.started
+    requirement_level: recommended
+    brief: >
+      Recorded when a live, bidirectional speech-to-speech session (such as OpenAI Realtime
```
lmolkova: same, let's see if we need it to be speech specific or relax it

lmolkova (commented):
Thanks for putting this together!

A few high-level feedback points:

- Multi-modal scope: Realtime WebSocket connections handle text, audio, and video, so let's make sure the conventions don't treat voice as the only modality.

- Naming: Let's move away from provider-specific terms (generate_live_content) to vendor-neutral names like realtime_inference (or live_inference) and realtime_session.

- Keep definitions concise and normative: The current brief and note descriptions are very verbose and read like design rationale. Semantic conventions should be short, factual, and normative (MUST/SHOULD). Please move justifications/history to the PR description and keep the convention text focused on actionable rules for instrumentation authors and telemetry consumers.

docs/gen-ai/gen-ai-spans.md
singankit: Addressed in the latest push. Introduced a dedicated `gen_ai.live_session.id` attribute (in `registry.yaml`) that correlates the `gen_ai.generate_live_content.client` and `gen_ai.user_input.internal` spans and the live events, instead of overloading `gen_ai.conversation.id`. The docs prose and generated tables now reference it.

model/gen-ai/spans.yaml
singankit: Addressed. Removed `gen_ai.response.time_to_first_chunk` from the `gen_ai.generate_live_content.client` span. Since the span starts at input-end (when the user stops speaking) and the model may already be processing before that boundary, TTFC cannot be computed relative to a meaningful span-start, so it was dropped from the live scenarios and the design doc.

model/gen-ai/spans.yaml
singankit: Addressed. Removed `gen_ai.response.time_to_first_chunk` from the `gen_ai.generate_live_content.client` span. Since the span starts at input-end (when the user stops speaking) and the model may already be processing before that boundary, TTFC cannot be computed relative to a meaningful span-start, so it was dropped from the live scenarios and the design doc.

model/gen-ai/spans.yaml
singankit: Addressed. Renamed the span to `gen_ai.user_input.internal` with `kind: internal`: its duration is the user speaking (voice-activity start to stop) over an already-open session, with no outbound request awaited and no matching server span, so CLIENT did not fit. Also removed the session-describing attributes (`server.address`, `server.port`, `gen_ai.request.model`) since they describe the session rather than this operation; correlation to the session is now carried by `gen_ai.live_session.id`. A note on the span explains the INTERNAL rationale.

model/gen-ai/spans.yaml
singankit: Addressed. Renamed the operation and span to the vendor-neutral `realtime_inference` / `gen_ai.realtime_inference.client` (dropping the Gemini-specific `generate_live_content` term). Updated the `gen_ai.operation.name` enum member, the span type, all model cross-references, the three reference scenarios + conformance mappings, the reference report (now `realtime-inference-span.md`), and regenerated docs/reports. Scenarios pass and generate-all is clean.

docs/gen-ai/gen-ai-spans.md
singankit: Done. Moved the entire live/realtime section out of `gen-ai-spans.md` into a new dedicated doc, `docs/gen-ai/gen-ai-realtime-spans.md` (the ``Live content generation`` narrative, trace-structure diagrams, the `gen_ai.realtime_inference.client` and `gen_ai.user_input.internal` span reference blocks, open questions, and other-vendor notes). `gen-ai-spans.md` now keeps a short stub under `### Live content generation` that links to the new page, and its table of contents was trimmed accordingly. Updated the inbound link from `gen-ai-events.md` and added the page to the `docs/gen-ai/README.md` index. generate-all is idempotent and all anchors resolve.

reference/scenarios/openai-realtime/scenario.py
singankit: Good catch — the previous scenario wrapped the app's own `append`/`commit` and ended at `input_audio_buffer.committed`, which are client calls, not the server VAD boundaries the convention defines.

Reworked the openai-realtime scenario to server-VAD semantics: the client just streams audio (`input_audio_buffer.append`, no manual commit), and the mock now emits server-detected `input_audio_buffer.speech_started` / `speech_stopped` (then an auto `committed`). The `user_speech` span now **starts at the received `speech_started` event and ends at the received `speech_stopped` event** — the exact boundaries a generic instrumentation observing the realtime event stream would key off — via `start_span()` / `span.end()` on those events rather than wrapping app calls. This demonstrates the VAD-bracketed span the convention requires.

docs/registry/attributes/gen-ai.md
singankit: Renamed the operation `user_input` → `user_speech` and the span `gen_ai.user_input.internal` → `gen_ai.user_speech.internal`, and made the definition explicitly speech-specific rather than generically multi-modal.

Rationale (matching your suggestion to make it speech-specific): the span's sole justification is the speaking *interval* reconstructed from voice-activity events. Text and other non-speech input have no such interval, so they are not modeled by this span — they are carried on the `realtime_inference` span through `gen_ai.input.messages`. The brief/note now state this carve-out explicitly.

I went with the noun form `user_speech` (over `record_/receive_user_speech`) to match how operations name the thing in this registry (`chat`, `embeddings`, `retrieval`), but happy to switch to a verb form if you prefer. Updated across the model, the openai-realtime scenario/conformance, docs (`gen-ai-realtime-spans.md`), the report (`user-speech-span.md`), and the changelog.

model/gen-ai/events.yaml
singankit: Relaxed to modality-neutral `live, bidirectional streaming session` across the session/generation briefs (live_session.started/.ended, `gen_ai.live_session.id`, `realtime_inference`), keeping OpenAI Realtime / Gemini Live as examples. A session is just a long-lived streaming connection and is genuinely multimodal, so it shouldn't be speech-scoped. Only `user_speech` stays speech-specific (VAD-delimited speaking interval).

changelog.d/394.enhancement.md
```diff
@@ -0,0 +1 @@
+Add a `realtime_inference` operation and the `gen_ai.realtime_inference.client` span to model per-generation server-side output in live, bidirectional streaming sessions (such as OpenAI Realtime and Gemini Live), together with `gen_ai.client.live_session.started`/`gen_ai.client.live_session.ended` events for session lifecycle, a `gen_ai.live_session.id` attribute to correlate the generations, user inputs, and lifecycle events that belong to the same live session, an optional VAD-gated `user_speech` operation and `gen_ai.user_speech.internal` span for the user's utterance, audio token usage (`gen_ai.usage.audio.input_tokens`/`gen_ai.usage.audio.output_tokens`), an optional `transcript` field on audio message parts to carry the text transcript of spoken audio, and guidance on sibling versus child tool-call span shapes. Reference scenarios cover OpenAI Realtime and Gemini Live (sibling pattern) and Gemini Live 3.x (child pattern).
```
DylanRussell: nit: what is an "operation"- a log event ?

DylanRussell: What does "VAD-gated" mean ?

DylanRussell:
Raw audio (PCM chunks for gemini) and video (1FPS of jpeg/png for gemini) is going to be way too big to put in the telemetry.. A minute of audio/video is MBs of bytes.. I would explicitly state we should leave it out of telemetry entirely.. Maybe we can figure out how to integrate it with the completion hook later on

model/gen-ai/registry.yaml
```diff
@@ -443,6 +443,27 @@ attributes:
       SHOULD leave it unset otherwise.
     examples: [true]
     stability: development
+  - key: gen_ai.live_session.id
```
lmolkova: The identifiers mix `realtime_inference` with `live_session` for the same protocol family. Please pick either `realtime` or `live` consistently across our operation, span, attribute, and event names.

docs/gen-ai/non-normative/models.py
```diff
@@ -207,6 +207,12 @@ class BlobPart(BaseModel):
     content: bytes = Field(
         description="Raw bytes of the attached data. This field SHOULD be encoded as a base64 string when serialized to JSON."
     )
+    transcript: Optional[str] = Field(
```
lmolkova: An attached transcript alone cannot represent audio and transcription chunks with independent completion boundaries. Please support a separate `transcription` part or reuse `text` part.

- [OpenAI Realtime](https://developers.openai.com/api/docs/guides/realtime-conversations) and [Grok](https://docs.x.ai/developers/rest-api-reference/inference/voice): separate audio/transcript deltas and `.done` events.
- [Gemini Live](https://ai.google.dev/api/live): `inputTranscription`/`outputTranscription` arrive separately from audio, without guaranteed chunk alignment.
- Buffered content can share a part once both are complete and their association is known, but it's extra logic in instrumentation and requires artificial coupling

reference/scenarios/openai-realtime/scenario.py
```diff
@@ -0,0 +1,364 @@
+"""Reference implementation for OpenAI Realtime (voice-native) inference.
```
lmolkova, line 1: All three new scenarios are missing the required README with library scope and operation coverage. Please add one per scenario, including which operations the SDK can instrument. See the [reference requirements](https://github.com/open-telemetry/semantic-conventions-genai/blob/2b902ce14e7fa9d3d97cac49de02bae6917c02ae/.github/skills/reference/SKILL.md#scenario-readme-referencescenarioslibraryreadmemd).

reference/scenarios/openai-realtime/scenario.py
```diff
@@ -0,0 +1,364 @@
...
+        "gen_ai.tool.call.id": function_call["call_id"],
+        "gen_ai.tool.type": "function",
+    }
+    with _reference_tracer.start_as_current_span(
+        f"execute_tool {function_call['name']}", attributes=tool_attributes
+    ) as span:
+        span.set_attribute("gen_ai.tool.call.arguments", json.dumps(arguments))
+        result = {"location": arguments.get("location"), "temperature_f": 72, "conditions": "sunny"}
+        span.set_attribute("gen_ai.tool.call.result", json.dumps(result))
+    return result
```
lmolkova: These spans measure application code. The realtime interfaces used here require the application to execute custom tools and send results back: [OpenAI](https://developers.openai.com/api/docs/guides/realtime-conversations), [Gemini Live](https://ai.google.dev/gemini-api/docs/live-api/tools).

```suggestion
def _run_execute_tool(provider, function_call):
    arguments = json.loads(function_call["arguments"])
    return {"location": arguments.get("location"), "temperature_f": 72, "conditions": "sunny"}
```

- Remove `execute_tool` spans from all three scenarios and update their coverage reports.
- Capture received requests as `tool_call` output parts and sent results as `tool_call_response` input parts. Gemini Live 3 omits both.
- Use the response part's `response` field, not `result`, in OpenAI and Gemini Live.

reference/scenarios/openai-realtime/scenario.py
```diff
@@ -0,0 +1,364 @@
...
+        span.set_attribute("gen_ai.input.messages", input_messages)
+
+        response_id = None
+        transcript_deltas = []
+        audio_deltas = []
+        final_transcript = None
+        status = None
+        usage = None
+        function_call = None
+        for event in conn:
```
lmolkova: The predetermined speech → generation → tool sequence drops interleaved events: this loop ignores `input_audio_buffer.speech_started` during a response, so interrupted output loses the new user-speech span.

Please demonstrate continuous event dispatch with independent speech/generation state, driven by provider events rather than the scenario's expected sequence.

model/gen-ai/registry.yaml
```diff
@@ -608,6 +629,30 @@ attributes:
...
+            This operation describes a single server-side generation reconstructed from a
+            long-lived streaming session, not a single client request/response. It is distinct
+            from `chat` and `generate_content`: the request is streamed continuously and the
+            generation is bounded by provider events (first output chunk or a voice-activity
+            end anchor through generation completion or interruption) rather than by a client
+            call boundary.
+          stability: development
+        - id: user_speech
+          value: "user_speech"
+          brief: 'Client-side capture of a user speech utterance within a live, bidirectional speech-to-speech session, delimited by the provider''s user voice-activity (start/stop) events.'
```
lmolkova: do we need to be this specific? 

```suggestion
          brief: Describes the process of recording user speech input'
```

then it can be used for traditional inference too
?

model/gen-ai/registry.yaml
```diff
@@ -608,6 +629,30 @@ attributes:
...
+          note: >
+            This operation is optional and provider-dependent. It is recorded only when the provider
+            exposes user voice-activity detection (VAD) events that bracket when the user started and
+            stopped speaking (for example OpenAI Realtime `input_audio_buffer.speech_started` /
+            `input_audio_buffer.speech_stopped`). It is speech-specific: text and other non-speech
+            input have no such speaking interval and are carried on the `realtime_inference`
+            operation through `gen_ai.input.messages` instead. Providers without user VAD events
+            (such as the Gemini Developer API) likewise carry the user input on the
+            `realtime_inference` operation through `gen_ai.input.messages`.
```
lmolkova, line 654:
```suggestion
          note: >
            This operation SHOULD be reported when instrumentation can reliably determine
            start and end of user speech either through user voice-activity detection (VAD) events
            in model response or via client-side events when application requires user to manually
            start and stop recording.
            
            Examples:
            - OpenAI Realtime `input_audio_buffer.speech_started` /
            `input_audio_buffer.speech_stopped`
            - ...
            - 
```

model/gen-ai/spans.yaml
```diff
@@ -270,6 +270,159 @@ spans:
...
+        client-side tool runtime does not inflate model-generation latency.
+      - Child pattern (Gemini Live 3.x): a single generation stays open across the tool call and
+        the `execute_tool` span is nested as its child.
+
+      Both patterns are valid; a single normative recommendation across providers is not yet
+      established.
+
+      When the provider exposes user voice-activity events, the user speech MAY additionally be
+      modeled by an optional `gen_ai.user_speech.internal` span; otherwise the user input is carried
+      on this span through `gen_ai.input.messages`.
```
lmolkova, line 328: can you please de-AI it? it has way too much text and includes details about other operations that are also expressed in the design doc.

model/gen-ai/spans.yaml
```diff
@@ -270,6 +270,159 @@ spans:
...
+      `gen_ai.request.model`) are not recorded on this span; the user speech is correlated to the
+      session and its generations through `gen_ai.live_session.id`.
+
+      Providers that do not expose user VAD events (such as the Gemini Developer API) cannot
+      record this span. In that case the user input is carried on the
+      `gen_ai.realtime_inference.client` span through `gen_ai.input.messages` instead.
+
+      The `gen_ai.operation.name` SHOULD be `user_speech`. The generations that respond to this
+      speech are correlated through `gen_ai.live_session.id` when the provider exposes a session
+      identifier.
```
lmolkova, line 411: same here, most of it is covered in user_speech operation and design doc too

reference/reports/execute-tool-span.md
```diff
@@ -6,8 +6,8 @@
 
 | Attribute | Supporting Libraries |
 | --- | --- |
-| gen_ai.operation.name | [agent-framework], [autogen], [crewai], [google-adk], [google-genai], [langchain], [llamaindex], [openai-agents], [openai-assistants], [pydantic-ai], [vertexai] |
-| gen_ai.tool.name | [agent-framework], [autogen], [crewai], [google-adk], [google-genai], [langchain], [llamaindex], [openai-agents], [openai-assistants], [pydantic-ai], [vertexai] |
+| gen_ai.operation.name | [agent-framework], [autogen], [crewai], [gemini-live], [gemini-live-3], [google-adk], [google-genai], [langchain], [llamaindex], [openai-agents], [openai-assistants], [openai-realtime], [pydantic-ai], [vertexai] |
```
lmolkova: why do we need to separate [gemini-live], [gemini-live-3],[google-genai] ? it's all in one library

model/gen-ai/registry.yaml
```diff
@@ -443,6 +443,27 @@ attributes:
       SHOULD leave it unset otherwise.
     examples: [true]
     stability: development
+  - key: gen_ai.live_session.id
+    type: string
+    brief: The unique identifier for a live, bidirectional streaming session (such as OpenAI Realtime or Gemini Live), used to correlate the generations, user inputs, and lifecycle events that belong to the same session.
```
```suggestion
    brief: The unique identifier for a live, bidirectional streaming session.
```

model/gen-ai/registry.yaml
```diff
@@ -443,6 +443,27 @@ attributes:
...
+
+      Instrumentations SHOULD populate the live session id when the provider exposes a session
+      identifier (for example the OpenAI Realtime `session.id`). When the provider exposes no
+      session identifier (for example the Gemini Developer API), instrumentations MAY use a
+      client-created per-connection identifier — the live connection is the session boundary —
+      or leave the attribute unset.
+
+      `gen_ai.live_session.id` is orthogonal to `gen_ai.conversation.id`: a single conversation
+      can span multiple live sessions, so the live session id is not a substitute for the
+      conversation id.
```
```suggestion
      A live session is a single long-lived streaming connection during which the client
      streams input continuously and the server streams incremental output.
      
      <Please put examples here>
      - OpenAI Realtime 
      -  Gemini - no session when using developer API. There is session id in enterprise
      
      Instrumentation MAY generate or reuse per-connection identifier or leave it unset when instrumented 
      library does not provide one.
      
      `gen_ai.live_session.id` is orthogonal to `gen_ai.conversation.id`: a single conversation
      can span multiple live sessions, so the live session id is not a substitute for the
      conversation id.
```

model/gen-ai/spans.yaml
```diff
@@ -270,6 +270,159 @@ spans:
       - ref: gen_ai.request.model
         sampling_relevant: true
 
+  - type: gen_ai.realtime_inference.client
+    requirement_level: recommended
```
lmolkova, line 274: given it can last for hours (translation / transcription) we should probably make it opt-in or leave a note saying something like

> Instrumentation SHOULD NOT create this span by default when expected duration can be long (minutes?). For example, when translating speech or in other cases when there is no conversation between user and model. Instrumentations MAY use model name or other heuristics available in
instrumented library to determine whether the call is expected to be conversational.
> 
> Instrumentation that instrument live SHOULD provide configuration option to disable realtime spans.

(not sure about last one, feel free to change to MAY or drop)

model/gen-ai/spans.yaml
```diff
@@ -270,6 +270,159 @@ spans:
...
+        requirement_level: recommended
+      - ref: gen_ai.usage.reasoning.output_tokens
+        requirement_level:
+          recommended: When applicable.
+        note: >
+          The value SHOULD be included in `gen_ai.usage.output_tokens`.
+      - ref: gen_ai.live_session.id
+        requirement_level:
+          conditionally_required: If the provider exposes a session identifier, or the instrumentation uses a client-created per-connection identifier.
+        note: >
```
lmolkova: is this note important here on this span? it seems pretty generic and covers what's already present in the definition

model/gen-ai/spans.yaml
singankit: Added the long-duration opt-out note to the `realtime_inference` span: instrumentation SHOULD NOT create it by default when the expected generation duration can be long (e.g. speech translation/transcription with no user-model conversation), and MAY use the model name or other library heuristics to decide. Dropped the second sentence (the SHOULD-provide-a-disable-config-option) per your note.

model/gen-ai/registry.yaml
singankit: Applied — brief trimmed to `The unique identifier for a live, bidirectional streaming session.` in https://github.com/open-telemetry/semantic-conventions-genai/pull/394/commits/43171ce34db20259730197670b26c9f60ea12669

model/gen-ai/registry.yaml
singankit: Applied your rewrite and filled the examples placeholder: OpenAI Realtime exposes a session id (`session.id`); Gemini has none on the Developer API but does on the enterprise API. Kept the per-connection-identifier MAY guidance and the `conversation.id` orthogonality note. https://github.com/open-telemetry/semantic-conventions-genai/pull/394/commits/43171ce34db20259730197670b26c9f60ea12669

reference/reports/execute-tool-span.md
singankit: Fixed. The report keyed `Supporting Libraries` off the scenario directory name; it now groups by the `instrumented_library` declared in each `conformance.yaml` and merges scenarios that share a library. `gemini-live`, `gemini-live-3`, and `google-genai` now collapse to a single `[google-genai]` (and `openai-realtime` folds into `[openai]`). https://github.com/open-telemetry/semantic-conventions-genai/pull/394/commits/43171ce34db20259730197670b26c9f60ea12669

model/gen-ai/registry.yaml
singankit: Standardized on `live`: renamed the operation `realtime_inference` -> `live_inference` and the span `gen_ai.realtime_inference.client` -> `gen_ai.live_inference.client` across the model, scenarios, docs, and reports. `gen_ai.live_session.id` and the `live_session.started`/`.ended` events are unchanged, so the whole family is now `live`. https://github.com/open-telemetry/semantic-conventions-genai/pull/394/commits/43171ce34db20259730197670b26c9f60ea12669

model/gen-ai/spans.yaml
singankit: Good point — dropped the span-level `note` override so the `gen_ai.live_session.id` footnote now falls back to the attribute's registry definition instead of restating it. https://github.com/open-telemetry/semantic-conventions-genai/pull/394/commits/2467062e60aa612ed2552d9ea71e3fde3f824c80

model/gen-ai/registry.yaml
singankit: Standardized the whole family on `realtime` (reversing my earlier `live` pass): operation `realtime_inference`, span `gen_ai.realtime_inference.client`, attribute `gen_ai.realtime_session.id`, and events `gen_ai.client.realtime_session.started`/`.ended`. No more `realtime`/`live` mix across operation, span, attribute, and event names. Scenarios, docs, reports, and the changelog are regenerated to match. https://github.com/open-telemetry/semantic-conventions-genai/pull/394/commits/38c671922000c6090591eee4563557f039b200a1

docs/gen-ai/non-normative/models.py
singankit: Added a dedicated `transcription` message part (`type: transcription`) and dropped the attached `transcript` field from the blob/file/uri parts. Modeling the transcript as its own part lets audio and transcription complete on independent boundaries (OpenAI Realtime/Grok audio vs. transcript `.done` deltas, Gemini Live `inputTranscription`/`outputTranscription`) without artificial coupling, and avoids the extra discriminator a reused `text` part would need. The three realtime scenarios now emit the audio blob and the transcription as separate parts. https://github.com/open-telemetry/semantic-conventions-genai/pull/394/commits/035224cd5898236a054d4dc1d0da87927e24599a

reference/scenarios/openai-realtime/scenario.py
singankit: Added a hand-written `README.md` to each new scenario (`openai-realtime`, `gemini-live`, `gemini-live-3`) with the library-boundary description and an operation-coverage table (realtime_inference / user_speech / execute_tool), following the format in the reference skill. `user_speech` is marked ➖ Not instrumentable for both Gemini Live scenarios since the Gemini Developer API exposes no user voice-activity (VAD) events.

https://github.com/open-telemetry/semantic-conventions-genai/pull/394/commits/f8ad4d8026c6d937eb2cd036d1c8ef25b7a04d7d

reference/scenarios/openai-realtime/scenario.py
singankit: Done in df9bfd3.

- Removed the `execute_tool` spans from all three scenarios — the client runs the tool, so it's application code, not a model operation. Coverage reports/READMEs updated accordingly.
- The tool exchange is now captured as message parts: a `tool_call` output part on the requesting generation and a `tool_call_response` input part on the answer generation (sibling pattern — openai-realtime, gemini-live). Gemini Live 3.x (child pattern) resolves the round-trip inside one generation and omits both, which I've documented as an honest capture gap.
- Switched the response part to the schema's `response` field (was `result`) in openai-realtime and gemini-live.

I also updated the `realtime_inference` span note and the realtime-spans doc prose so the convention describes the message-part representation rather than an `execute_tool` span shape, and refreshed the changelog.

https://github.com/open-telemetry/semantic-conventions-genai/pull/394/commits/df9bfd3

reference/scenarios/openai-realtime/scenario.py
singankit: Reworked the driver into an event-driven dispatch loop (`_RealtimeDriver`) that reacts to server events instead of the imperative turn-by-turn flow, matching how a real server-VAD realtime client is structured.

This also lets the scenario demonstrate a barge-in: a `user_speech` span now opens while a `realtime_inference` generation is still in flight, which cancels that generation. The mock's `interrupted` behavior brackets the barge-in utterance (`speech_stopped` + `committed`) after the cancelled response so the client answers it with a follow-up generation.

Addressed in https://github.com/open-telemetry/semantic-conventions-genai/pull/394/commits/1ee2ff5e3a4bdc77f1bc0b50d9159d794c2528ff

singankit:
Raw media (audio/video) is handled by the same tiered content model as chat spans, so it isn't a realtime-specific problem:

1. **Default (Tier 1):** message content — including audio/video parts — is **not** captured. `gen_ai.input.messages` / `gen_ai.output.messages` are `Opt-In`, gated behind an explicit flag (e.g. `OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT`). So nothing large hits the span by default.
2. **Opt-in on span (Tier 2):** if an operator opts in, content is recorded on the span/event as structured message parts. Instrumentation MAY truncate individual contents.
3. **External storage (Tier 3):** for large/sensitive media, the documented pattern is the in-process **content upload hook**, which operates independently of the opt-in flags and can offload the bytes and record a `uri`/`file` reference on the span instead of inline bytes.

Given this, I'd prefer **not** to state "leave raw audio/video out of telemetry entirely." Capturing it is valuable for agent-internals visibility, and the framework already keeps it safe: off by default, opt-in inline for small cases, and external-storage references (the hook — essentially your "completion hook" idea) for large media. Realtime just inherits this model, so audio/video parts are subject to the same tiers as any other message content.

The one genuine gap is that recording references to externally stored content is still a spec TODO, and streamed realtime audio has no provider-supplied `file_id`/`uri` for the hook to reuse — so the hook would have to persist the media and mint the reference itself.

changelog.d/394.enhancement.md
singankit: Good catch — "operation" here is the GenAI operation type (a `gen_ai.operation.name` value), realized as a span, not a log event. So `realtime_inference` is an operation name that maps to the `gen_ai.realtime_inference.client` span (and `user_speech` to `gen_ai.user_speech.internal`).

I've anchored the term on first use in the changelog so it isn't mistaken for a log event: https://github.com/open-telemetry/semantic-conventions-genai/pull/394/commits/b215381a19728c705ffb0bc2cd649f246d4124e5

singankit: "VAD-gated" referred to Voice Activity Detection — the provider's server-side detection of when the user starts/stops speaking (e.g. OpenAI Realtime's `input_audio_buffer.speech_started` / `speech_stopped`). The `user_speech` span is only created when those events are surfaced to bound the utterance; providers that don't emit them (such as the Gemini Developer API) carry the user input on the generation's `input.messages` instead.

I've dropped the unexplained jargon from the changelog and spelled this out in plain language: https://github.com/open-telemetry/semantic-conventions-genai/pull/394/commits/3bc29d22e961cdb753060bc097da844a7a74a0a2

reference/scenarios/openai-realtime/scenario.py
```diff
@@ -0,0 +1,164 @@
...
+        "gen_ai.operation.name": "chat",
+        "gen_ai.provider.name": "openai",
+        "gen_ai.request.model": request_model,
+        # The realtime session produces speech output for this turn.
+        "gen_ai.output.type": "speech",
+    }
+    if host:
+        chat_attributes["server.address"] = host
+    if port is not None:
+        chat_attributes["server.port"] = port
```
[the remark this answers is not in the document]
singankit: Fixed — `server.port` now reflects the realtime WebSocket endpoint the instrumented request actually reaches (`http_port + 1`), not the HTTP `/health` port. https://github.com/open-telemetry/semantic-conventions-genai/pull/394/commits/df9bfd35dbc7ecd4c4c4f9c5840534551b2aa682

reference/scenarios/openai-realtime/scenario.py
```diff
@@ -0,0 +1,164 @@
...
+                    },
+                ],
+                "finish_reason": status,
+            }
+        ]
+        chat_span.set_attribute("gen_ai.response.model", response_model)
+        if response_id:
+            chat_span.set_attribute("gen_ai.response.id", response_id)
+        chat_span.set_attribute("gen_ai.response.finish_reasons", [status])
+        chat_span.set_attribute("gen_ai.output.messages", json.dumps(output_messages))
```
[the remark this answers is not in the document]
singankit: Fixed — added a `_STATUS_TO_FINISH_REASON` map (`completed`→`stop`, `incomplete`→`length`, `failed`→`error`, `cancelled`→`stop`) so `gen_ai.response.finish_reasons` and the `finish_reason` in `gen_ai.output.messages` carry the model's stop reason rather than the Realtime transport status. https://github.com/open-telemetry/semantic-conventions-genai/pull/394/commits/df9bfd35dbc7ecd4c4c4f9c5840534551b2aa682

reference/scenarios/openai-realtime/scenario.py
```diff
@@ -0,0 +1,164 @@
...
+        if usage:
+            chat_span.set_attribute("gen_ai.usage.input_tokens", usage.input_tokens)
+            chat_span.set_attribute("gen_ai.usage.output_tokens", usage.output_tokens)
+            if usage.input_token_details and usage.input_token_details.audio_tokens:
+                chat_span.set_attribute("gen_ai.usage.input_audio_tokens", usage.input_token_details.audio_tokens)
+            if usage.output_token_details and usage.output_token_details.audio_tokens:
+                chat_span.set_attribute("gen_ai.usage.output_audio_tokens", usage.output_token_details.audio_tokens)
```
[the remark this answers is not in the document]
singankit: Fixed — the audio-token attributes are now guarded with `is not None` instead of a truthiness check, so a reported `0` is still emitted (matching how the other usage subfields are handled). https://github.com/open-telemetry/semantic-conventions-genai/pull/394/commits/df9bfd35dbc7ecd4c4c4f9c5840534551b2aa682

reference/scenarios/openai-realtime/scenario.py
singankit: Addressed by not modeling the session as a span. Because instrumentation can't reliably bound a session span (a realtime session is long-lived and mostly idle, and there's no clean single start/end operation), the session lifecycle is captured as two events instead — `gen_ai.client.realtime_session.started` (on `session.created`) and `gen_ai.client.realtime_session.ended` (on close). The per-generation `realtime_inference` spans and the `user_speech` span are correlated back to the session via the `gen_ai.realtime_session.id` attribute rather than span parenting.

DylanRussell:
I think audio / video content differs from what exists today in that it is a order of magnitude bigger.. A few seconds of audio / video is too big to put in.. Saying that instrumentations "may truncate" is vague.. Also the fact that it is opt-in doesn't solve the problem either

reference/scenarios/gemini-live-3/mock_key.pem
singankit: Good call — dropped the checked-in certs. The mock server now generates an ephemeral self-signed cert in memory at startup (the client already connects with validation disabled, so its contents don't matter). Done in 81c1891.

reference/scenarios/openai-realtime/scenario.py
singankit: Plan is to keep behavior similar to text model inference span. We briefly discussed it in GenAI SIG meeting.

reference/scenarios/openai-realtime/scenario.py
singankit: Addressed by modeling the transcript as its own part rather than a field on `BlobPart`. `BlobPart` no longer carries a `transcript`, so it can't show up on image or document parts. Spoken audio and its transcription are now emitted as two sibling parts:

```json
{"type": "blob", "modality": "audio", "mime_type": "audio/pcm", "content": "..."},
{"type": "transcription", "content": "..."}
```

There's a dedicated `TranscriptionPart` (`type: "transcription"`) whose docstring explains why they're separate parts: providers stream/complete audio and transcript on independent boundaries (OpenAI Realtime/Grok emit separate `.done` events; Gemini Live delivers `inputTranscription`/`outputTranscription` separately from the audio). It's a general transcription part rather than an audio-specific `AudioPart` so we don't need an extra identifier to tell a transcript apart from ordinary text.

model/gen-ai/spans.yaml
```diff
@@ -270,6 +270,166 @@ spans:
...
+          then the value must be the exact name of the model actually used. If the model is a
+          fine-tuned custom model, the value should have a more specific name than the base model that's been fine-tuned.
+      - ref: gen_ai.response.finish_reasons
+        requirement_level: recommended
+      - ref: gen_ai.usage.reasoning.output_tokens
+        requirement_level:
+          recommended: When applicable.
+        note: >
+          The value SHOULD be included in `gen_ai.usage.output_tokens`.
+      - ref: gen_ai.realtime_session.id
```
allen-stephen, line 361: `gen_ai.conversation.id` is absent from this span, though the inference and agent spans carry it as Conditionally Required. `realtime_session.id` isn't a substitute, since a conversation can span multiple sessions, so a voice conversation that reconnects or transfers has no correlator. Suggest adding `gen_ai.conversation.id` here alongside it.

model/gen-ai/spans.yaml
```diff
@@ -270,6 +270,166 @@ spans:
...
+
+      The `gen_ai.operation.name` SHOULD be `realtime_inference`.
+
+      The realtime session itself is not modeled as a span. It is represented by the
+      `gen_ai.client.realtime_session.started` and `gen_ai.client.realtime_session.ended` events and by
+      `gen_ai.realtime_session.id` correlating the generations within it. Turn-level spans (a full
+      user-and-model exchange) are out of scope: turn boundaries cannot be detected reliably
+      across providers.
+
+      When a generation calls a tool, the realtime API returns the tool call to the client, which
```
allen-stephen, line 308: The rationale here, "application code, not a model operation," applies equally to non-live chat, where `execute_tool` is normative. Whether the span can be emitted depends on the instrumentation layer. A provider-SDK instrumentation can't observe the tool run; a framework that dispatches it can (like ADK which emits execute_tool on its non-live path today).

Suggested replacement:

> Tool execution in a realtime session is performed by the client, so a provider-SDK instrumentation cannot observe it and will not emit a `gen_ai.execute_tool.internal` span; the tool exchange is captured as `tool_call` / `tool_call_response` parts instead. Instrumentation at a layer that dispatches the tool, such as an agent framework, should emit it as usual: a sibling of the two generations under the sibling pattern, or a child of the single generation under the child pattern.

lmolkova:
> I think audio / video content differs from what exists today in that it is a order of magnitude bigger.. A few seconds of audio / video is too big to put in.. Saying that instrumentations "may truncate" is vague.. Also the fact that it is opt-in doesn't solve the problem either

@DylanRussell  I think it's an existing problem we have with request-response inference - someone can easily put in a few seconds of audio and video into messages.

Created https://github.com/open-telemetry/semantic-conventions-genai/issues/515 to track it.

model/gen-ai/spans.yaml
singankit: Good catch — added `gen_ai.conversation.id` to the `realtime_inference.client` span alongside `gen_ai.realtime_session.id`, with the same `Conditionally Required` wording the `inference.client` / `invoke_agent` spans use. Agreed that `realtime_session.id` isn't a substitute: it correlates generations within a connection, while a conversation can span multiple sessions (reconnect/transfer), so the cross-session correlator was missing. For the record it was never added when the span was authored rather than removed. Done in 9960e11.

model/gen-ai/spans.yaml
singankit: Good point — reframed the guidance around what the instrumentation layer can observe: a provider-SDK instrumentation can't observe the client-side tool run and won't emit the span, while a dispatching layer (agent framework) should emit it as usual. Folded the sibling/child placement into the distribution bullets. Done in 493d94f.

model/gen-ai/spans.yaml
singankit: The span is now `realtime_inference` (renamed from `generate_live_content`), and the note states the value directly: `gen_ai.operation.name` SHOULD be `realtime_inference`. Keeping it as an attribute ref follows the existing pattern all single-operation spans use.

model/gen-ai/spans.yaml
singankit: Clarified in the note: it now reads "token usage (`gen_ai.usage.*`)" instead of bare "usage" — i.e. the `gen_ai.usage.*` token counts some providers report after the completion event.

model/gen-ai/spans.yaml
singankit: Trimmed the note in 5fc01d2. The brief now mirrors `gen_ai.inference.client`, and the note keeps only the realtime-specific normative rules (span boundaries + usage timing, the long-running/non-conversational heuristic, `operation.name`, and client-side tool-call handling). The session-modeling, sibling/child tool-exchange patterns, and user_speech details are covered by the design doc, the `operation.name` enum notes, and the `user_speech` span, so they're dropped here. Please review.

model/gen-ai/spans.yaml
singankit: Trimmed this note in 70259d4, keeping only the record-when/span-boundaries rule and `gen_ai.operation.name`; the rest is covered by the `user_speech` operation enum notes and the design doc. Please review the content.

model/gen-ai/registry.yaml
singankit: Applied in ecc2154. Reframed the note so `user_speech` SHOULD be reported when instrumentation can reliably determine speech start/end -- via provider VAD events or client-side manual start/stop recording -- and filled the examples (OpenAI Realtime VAD events; push-to-talk client-side start/stop).

model/gen-ai/registry.yaml
singankit: Done in ecc2154 -- simplified the brief to `Describes the process of recording user speech input.` so it isn't tied to a realtime session.

changelog.d/448.enhancement.md
singankit: That is a PR I started with which is still in draft. It is not relevant to this PR right now.

model/gen-ai/spans.yaml
singankit: Addressed in f8ad4d8026c6d937eb2cd036d1c8ef25b7a04d7d. With #440 merged, this PR no longer adds the duplicate audio-token attributes or modifies the shared usage group. The realtime span reuses `attributes.gen_ai.usage.breakdown` from #440 instead.

model/gen-ai/spans.yaml
singankit: Clarified in [f8ad4d8](https://github.com/open-telemetry/semantic-conventions-genai/commit/f8ad4d8026c6d937eb2cd036d1c8ef25b7a04d7d): a generation is one server-side model response (the model's spoken or textual turn) within the realtime session.

model/gen-ai/spans.yaml
singankit: Agreed that the resumption token is separate from `gen_ai.conversation.id`. We can model it as a Gemini-specific attribute in a follow-up rather than add it to the generic realtime conventions here.

PR #395, opened 2026-07-23
Make `gen_ai.retrieval.documents` `id`/`score` optional (#344)

### LINKED ISSUE #344: Remove retrieval document id and score as required
https://github.com/trask/semantic-conventions-genai/blob/243e57bb0718f1ae2515c2814203613f2e5524cb/docs/gen-ai/gen-ai-retrieval-documents.json#L19-L20
1) LangChain's Document.id is Optional[str], and other retrieval libraries may similarly not always provide a document ID. Requiring it forces instrumentations to fabricate a value, which is worse than omitting the attribute.
2) on_retriever_end only receives Sequence[Document] — no scores. LangChain's Document has id, metadata, page_content, and type — no score field. The [similarity_search_with_relevance_scores](https://reference.langchain.com/python/langchain-core/vectorstores/base/VectorStore/similarity_search_with_relevance_scores) method returns list[tuple[Document, float]], but that score never flows into the callback handler interface; on_retriever_end only gets the plain documents.

nitishagar:
## What

Relax `gen_ai.retrieval.documents` `id` and `score` from recommended to optional.

Retrieval libraries may not always expose a document id or a relevance score. For example, LangChain's `Document.id` is `Optional[str]`, and `on_retriever_end` receives `Sequence[Document]` with no `score` field. Requiring these properties forces instrumentations to fabricate values, which is worse than omission.

## Changes

- Make `RetrievalDocument.id` / `score` `Optional` with `default=None` in the pydantic source in `docs/gen-ai/non-normative/models.py`. The JSON schema is generated from this model.
- Update the registry note from "SHOULD contain at least the following properties" to "MAY contain the following properties when available" (`model/gen-ai/registry.yaml`).
- Regenerate the affected artifacts: `model/gen-ai/gen-ai-retrieval-documents.json` (fields now nullable and removed from `required`), `docs/registry/attributes/gen-ai.md`, `docs/gen-ai/gen-ai-spans.md`.

The fields and their descriptions are retained, so existing consumers that read `id` / `score` when present keep working, and documents that populate them still validate.

## Verification

- `make generate-all` — clean diff (only the intended generated files).
- `make check-policies` — passes (no policy violation).
- `cd reference && uv run run-scenario langchain` — the langchain scenario emits retrieval documents without `id`/`score` and validates (`data.json` unchanged).

## Related

- Closes #344
- Unblocks the downstream instrumentation work tracked in opentelemetry-python-genai#124.

### DIFF
--- a/changelog.d/395.clarification.md
+++ b/changelog.d/395.clarification.md
@@ -1,0 +1,1 @@
+Relax `gen_ai.retrieval.documents` `id` and `score` from recommended to optional, since retrieval libraries (e.g. LangChain) may not always expose a document id or a relevance score.
--- a/docs/gen-ai/gen-ai-spans.md
+++ b/docs/gen-ai/gen-ai-spans.md
@@ -573,7 +573,7 @@ applicable `aws.bedrock.*` attributes and are not expected to include
 
 **[7] `server.address`:** When observed from the client side, and when communicating through an intermediary, `server.address` SHOULD represent the server address behind any intermediaries, for example proxies, if it's available.
 
-**[8] `gen_ai.retrieval.documents`:** Each document object SHOULD contain at least the following properties:
+**[8] `gen_ai.retrieval.documents`:** Each document object MAY contain the following properties when available:
 `id` (string): A unique identifier for the document, `score` (double): The relevance score of the document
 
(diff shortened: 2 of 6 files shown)


JWinermaSplunk:
Hi @nitishagar,
Apologies, but any chance we could consolidate into #396? I had to close #392 due to a branching issue.
Thanks!

PR #396, opened 2026-07-23
#344 Make retrieval document id and score optional

### LINKED ISSUE #344: Remove retrieval document id and score as required

JWinermaSplunk:
## Description

Fixes https://github.com/open-telemetry/semantic-conventions-genai/issues/344
Closes #395 

Make id and score optional properties of gen_ai.retrieval.documents.

This allows instrumentations to omit values instead of fabricating them.

## Motivation

Retrieval frameworks do not consistently expose document identifiers or relevance scores. For example, LangChain document identifiers are optional, and its retriever callback receives documents without their relevance scores.

## Prototype

The existing LangChain reference scenario invokes the public retriever API and emits retrieval documents containing available content and metadata without id or score.

## Checklist

See [CONTRIBUTING.md]

### DIFF
--- a/changelog.d/+.bugfix.md
+++ b/changelog.d/+.bugfix.md
@@ -1,0 +1,1 @@
+Make retrieval document `id` and `score` optional so instrumentations can omit values that are unavailable at runtime.
--- a/docs/gen-ai/non-normative/models.py
+++ b/docs/gen-ai/non-normative/models.py
@@ -34,7 +34,6 @@ from pydantic import (
     GetJsonSchemaHandler,
     RootModel,
 )
-from pydantic.json_schema import SkipJsonSchema
 from pydantic_core import core_schema
 
 
@@ -465,9 +464,8 @@ class RetrievalDocument(BaseModel):
     Represents a single document retrieved from a vector database or search system.
     """
 
-    id: str | SkipJsonSchema[None] = Field(
-        default_factory=lambda: None,
-        description="A unique identifier for the document.",
+    id: str | None = Field(
+        default=None, description="A unique identifier for the document."
     )
     score: float | None = Field(
         default=None, description="The relevance score of the document."
--- a/docs/gen-ai/non-normative/models.py
+++ b/docs/gen-ai/non-normative/models.py
@@ -464,8 +464,12 @@ class RetrievalDocument(BaseModel):
     Represents a single document retrieved from a vector database or search system.
     """
 
-    id: str = Field(description="A unique identifier for the document.")
-    score: float = Field(description="The relevance score of the document.")
+    id: Optional[str] = Field(
+        default=None, description="A unique identifier for the document."
+    )
+    score: Optional[float] = Field(
+        default=None, description="The relevance score of the document."
+    )
 
     model_config = ConfigDict(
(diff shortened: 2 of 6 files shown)


wrisa (approved):

docs/gen-ai/non-normative/models.py
```diff
@@ -464,8 +465,14 @@ class RetrievalDocument(BaseModel):
...
-    id: str = Field(description="A unique identifier for the document.")
-    score: float = Field(description="The relevance score of the document.")
+    id: str | SkipJsonSchema[None] = Field(
+        default_factory=lambda: None,
+        description="A unique identifier for the document.",
+    )
+    score: float | SkipJsonSchema[None] = Field(
+        default_factory=lambda: None,
+        description="The relevance score of the document.",
+    )
```
lmolkova:
```suggestion
    score: float | None = Field(default=None, description="The relevance score of the document.")
```

- `null` is a valid value, someone validating payload against json schema should not fail when null is there. I'd love to be strict, but not allowing null will be too strict
- `default` set to `None` gives a hint that absence and null are equivalent
- since it has a default, it's not required

changelog.d/+.bugfix.md
```diff
@@ -0,0 +1 @@
+Make retrieval document `id` and `score` optional so instrumentations can omit values that are unavailable at runtime.
```
lmolkova, line 1: seems there is a similar PR #395,  @nitishagar @JWinermaSplunk could you folks please decide which one we'll go ahead with? thanks!

JWinermaSplunk:
Hi @lmolkova,
Addressed your comments, if you have a moment for rereview. Thank you!

docs/gen-ai/non-normative/models.py
```diff
@@ -464,8 +465,13 @@ class RetrievalDocument(BaseModel):
     Represents a single document retrieved from a vector database or search system.
     """
 
-    id: str = Field(description="A unique identifier for the document.")
-    score: float = Field(description="The relevance score of the document.")
+    id: str | SkipJsonSchema[None] = Field(
```
lmolkova: could you please update this one too?

JWinermaSplunk: Done!

lmolkova (approved):

aabmass (approved):

PR #400, opened 2026-07-24
weaver 0.25.0: remove now unnecessary workarounds

lmolkova:
Weaver 0.25.0 fixed a couple of issues we workaround here, specifically 
- https://github.com/open-telemetry/weaver/pull/1414 - makefile updates 
- https://github.com/open-telemetry/weaver/pull/1403 (span name can now be rendered by templates and we can remove duplicated text) - markdown, yaml and jinja updates


model/gen-ai/spans.yaml
```diff
@@ -642,13 +615,14 @@ spans:
 span_refinements:
   - id: openai.inference.client
     ref: gen_ai.inference.client
+    name:
+      note: >
+        **Span name** SHOULD be `{gen_ai.operation.name} {gen_ai.request.model}`.
```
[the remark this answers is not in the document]
lmolkova: out of scope. and not right.

templates/registry/markdown/attribute_macros.j2
```diff
@@ -23,13 +23,14 @@
...
-   imported attributes link to the pinned upstream registry page. #}
+{# v2 link rendering: locally-defined attributes (their YAML file lives under
+   ./model/) carry `provenance.path`, while attributes imported from a dependency
+   carry `provenance.source` (the upstream schema URL). Local attributes link to
+   /docs/registry/attributes/<ns>.md; imported attributes link to the pinned
+   upstream registry page. #}
 {% macro name_with_link(attribute, attribute_registry_base_url, upstream_docs_base) -%}
 {%- set ns_id = name(attribute) | trim | split_id | first -%}
-[`{{ name(attribute) | trim }}`]({% if attribute.provenance %}{{ attribute_registry_base_url }}/{{ ns_id | kebab_case }}.md{% else %}{{ upstream_docs_base }}/docs/registry/attributes/{{ ns_id | kebab_case }}.md{% endif %})
+[`{{ name(attribute) | trim }}`]({% if attribute.provenance and attribute.provenance.path %}{{ attribute_registry_base_url }}/{{ ns_id | kebab_case }}.md{% else %}{{ upstream_docs_base }}/docs/registry/attributes/{{ ns_id | kebab_case }}.md{% endif %})
```
trask: This same fix already landed on main in #405, as `attribute.provenance.path` -- that's what the conflict is. Suggest taking main's form so the only delta left here is your (clearer) comment above.

Related, and worth checking before merge: #405 also widened the in-sync check from `git diff --exit-code -- ./docs/registry` to `-- ./docs`, and this branch's base predates that commit. So the green `generated-docs` run here never verified `docs/gen-ai/*.md` -- exactly the files this PR regenerates. Worth re-running `make generate-all` after merging main.

```suggestion
[`{{ name(attribute) | trim }}`]({% if attribute.provenance.path %}{{ attribute_registry_base_url }}/{{ ns_id | kebab_case }}.md{% else %}{{ upstream_docs_base }}/docs/registry/attributes/{{ ns_id | kebab_case }}.md{% endif %})
```

model/gen-ai/spans.yaml
trask: Now that refinements inherit `name` from the base span, this override is the duplication the PR is removing. `azure.ai.inference.client` and `aws.bedrock.inference.client` have no `name` and fall through to the base note -- which is why aws-bedrock.md and azure-ai-inference.md gained the guidance in this PR.

Keeping the override leaves openai/anthropic with the one-line wording and bedrock/azure with the full three-sentence one, for the same span, in sibling docs. The first sentence is identical to the base anyway; the override only drops the "MAY specify different span name format" and "MUST follow the overall guidelines" sentences.

Deleting it means openai.md picks up the base text, so this needs a `make generate-all`.

```suggestion
```

model/gen-ai/spans.yaml
```diff
@@ -734,13 +708,14 @@ span_refinements:
 
   - id: anthropic.inference.client
     ref: gen_ai.inference.client
+    name:
+      note: >
+        **Span name** SHOULD be `{gen_ai.operation.name} {gen_ai.request.model}`.
```
trask: Same as the openai refinement above -- this drops to the base `name.note` if removed, which is what bedrock and azure already do. Needs a `make generate-all` along with it.

```suggestion
```

templates/registry/markdown/span_macros.j2
```diff
@@ -5,21 +5,19 @@
...
 
 {{ span.note | trim }}{% endif %}
-{#- if span.name and span.name.note %}
+{%- if span.name and span.name.note %}
 
-**Span name** SHOULD be {{ span.name.note | trim }}.{% endif #}
+{{ span.name.note | trim }}{% endif %}
 {%- if "**Span kind" not in (span.note or "") %}
 
 **Span kind** SHOULD be `{{ span.kind | upper }}`.{% endif %}
```
trask, line 23: The `**Span kind`/`**Span status` guards below let a span inline that prose in its `note` instead. When it does, the prose renders as part of `note` -- ahead of the span name -- and the generated line is suppressed. So the span name lands in a different spot per span:

- `gen_ai.inference.client` (inlines kind): kind -> **Span name** -> status
- `mcp.server` (inlines status): status -> **Span name** -> kind
- `invoke_agent` (inlines neither): **Span name** -> kind -> status

Emitting `name.note` before `note` puts it in the same place everywhere. Needs a `make generate-all`.

```suggestion
{%- if span.name and span.name.note %}

{{ span.name.note | trim }}{% endif %}
{%- if span.note %}

{{ span.note | trim }}{% endif %}
{%- if "**Span kind" not in (span.note or "") %}

**Span kind** SHOULD be `{{ span.kind | upper }}`.{% endif %}
```

Makefile
```diff
@@ -129,12 +113,12 @@ $(SC_UPSTREAM_STAMP): $(VERSION_PINS_FILE)
...
 
-# Validate the model and run shared policies
-check-policies: $(LOCAL_POLICY_STAMP) $(SC_UPSTREAM_STAMP)
+# Validate the model and run shared policies from otel-weaver-packages.
+check-policies: $(SC_UPSTREAM_STAMP)
 	$(WEAVER) registry check \
 		-r ./model \
 		--v2 \
-		--policy $(LOCAL_POLICIES)/policies/check \
+		--policy '$(POLICY_REPO_URL)@$(POLICY_REPO_REF)[policies/check]' \
```
trask, line 121: Fine by me if this is the accepted cost of dropping the workaround, but flagging the trade explicitly: there's no caching left here at all. Weaver clones a remote vdir into a `TempDir` under `~/.weaver/vdir_cache` and deletes it when the `VirtualDirectory` is dropped (`create_tmp_repo` in `weaver_common/src/vdir.rs`), so nothing is reused between runs, local weaver or Docker -- the container mounts only `$(CURDIR):/workspace`. `make check-policies` now clones the policy pack from GitHub on every invocation and can't run offline, where the stamped `.build/weaver-policies` checkout it replaces was fetched once per `POLICY_REPO_REF`.

trask (approved):

templates/registry/markdown/span_macros.j2
lmolkova: i'll push back on this one. This would put free-form text for span.note under free-form text about span name resulting in something like 

> **Span name** SHOULD be `invoke_agent {gen_ai.agent.name}` if `gen_ai.agent.name` is readily available.
When `gen_ai.agent.name` is not available, it SHOULD be `invoke_agent`.
Semantic conventions for individual GenAI systems and frameworks MAY specify different span name format.
>
> Examples: LangChain agents, CrewAI agent

"Examples: LangChain agents, CrewAI agent" is a span note that has nothing to do with span name, but reads as a part of it.

Makefile
lmolkova: weaver can start caching, simplicity outweights tiny perf concern

PR #404, opened 2026-07-28
MCP server status: don't set to errors for client errors

### LINKED ISSUE #389: MCP: don't mark client problems as errors on the server
https://github.com/open-telemetry/semantic-conventions-genai/blob/150760c6252a4bb63c49c9915bad11997d316a15/docs/gen-ai/mcp.md?plain=1#L363-L364

does not align with 

HTTP approach

> For HTTP status codes in the 4xx range span status MUST be left unset in case of SpanKind.SERVER and SHOULD be set to Error in case of SpanKind.CLIENT.

Client-side errors are not errors on the server, so we'd set RPC status code but not error.type/span status for 

-32700 | Parse error | Invalid JSON was received by the server.An error occurred on the server while parsing the JSON text.
-32600 | Invalid Request | The JSON sent is not a valid Request object.
-32601 | Method not found | The method does not exist / is not available.
-32602 | Invalid params | Invalid method parameter(s).

lmolkova:
Fixes #389

## Motivation

HTTP and RPC conventions keep server span status unset for client-side errors
(like not found). MCP should follow the same pattern: on the server side,
`-32700`, `-32600`, `-32601`, `-32602`, and `-32002`
mean the caller sent a request the server could not serve, and are not errors.
Any other code is.

Stated as an exclusion list rather than an allowlist because JSON-RPC's
"implementation-defined server-errors" range holds a caller fault (`-32002`),
control flow (`-32042`), and SDK transport codes (`-32000`, `-32001`) — and an
allowlist would record unrecognized codes as success.

On the client side all error codes are errors: the client is the caller, so a
caller-fault code still means its own operation failed.

See https://www.jsonrpc.org/specification#error_object for list of json rpc errors codes

## Checklist

See [CONTRIBUTING.md]

-----------

PS: it's the perfect PR number for this change


lmolkova:
CI is fixed in #405

docs/gen-ai/mcp.md
```diff
@@ -158,27 +158,37 @@ option to enable it.
...
 | --- | --- | --- | --- | --- | --- |
 | [`mcp.method.name`](/docs/registry/attributes/mcp.md) | ![Development](https://img.shields.io/badge/-development-blue) | `Required` | string | The name of the request or notification method. | `notifications/cancelled`; `initialize`; `notifications/initialized` |
-| [`error.type`](https://github.com/open-telemetry/semantic-conventions/blob/v1.43.0/docs/registry/attributes/error.md) | ![Stable](https://img.shields.io/badge/-stable-lightgreen) | `Conditionally Required` If and only if the operation fails. | string | Describes a class of error the operation ended with. [1] | `timeout`; `java.net.UnknownHostException`; `server_certificate_invalid`; `500` |
+| [`error.type`](https://github.com/open-telemetry/semantic-conventions/blob/v1.43.0/docs/registry/attributes/error.md) | ![Stable](https://img.shields.io/badge/-stable-lightgreen) | `Conditionally Required` [1] | string | Describes a class of error the operation ended with. [2] | `timeout`; `java.net.UnknownHostException`; `server_certificate_invalid`; `500` |
```
lmolkova: TODO: return back

changelog.d/389.bugfix.md
```diff
@@ -0,0 +1 @@
+Fix MCP server span status and `error.type` conventions to not mark client-side errors (such as parse errors, invalid requests, method not found, and invalid parameters) as server-side failures.
```
trask: Parking my notes on the PR description here.

- The description lists the four JSON-RPC pre-defined client codes, but MCP also defines `-32002` (`Resource not found`) — the canonical "not found" case the motivation calls out. It falls in the `-32000`..`-32099` range this PR classifies as a *server* error, so it still gets reported as a server-side failure. Details in my comment on `model/mcp/spans.yaml`.
- "Reference scenarios updated for affected libraries" is unchecked. There are no MCP scenarios under `reference/scenarios/` today (nothing emits `mcp.method.name`), so it's N/A — worth saying so in the description so nobody blocks on it.
- The fragment only describes the server-side half. The PR also newly defines the client-side classification ("All JSON-RPC error codes SHOULD be considered errors") and splits `error.type` guidance by client/server, which consumers will want to know about.

```suggestion
Clarify MCP error reporting: MCP server telemetry no longer reports client-side JSON-RPC errors (parse errors, invalid requests, method not found, and invalid params) as failures, while MCP client telemetry reports all JSON-RPC error codes as errors.
```

model/mcp/common.yaml
```diff
@@ -26,10 +26,20 @@ attribute_groups:
         requirement_level: opt_in
       - ref: error.type
         requirement_level:
-          conditionally_required: If and only if the operation fails.
+          conditionally_required: >
+            On client spans, if and only if the operation fails. On server spans,
+            if and only if the operation fails with a server-side error, timeout, or transport error.
```
trask: Two issues with this requirement level.

**"spans" leaks into the metric tables.** `mcp.common.attributes` is also pulled in by `mcp.operation.metrics.attributes`, so this text renders verbatim into the `mcp.client.operation.duration` and `mcp.server.operation.duration` attribute tables, where there are no spans. Signal-neutral wording reads correctly on both.

**It contradicts the note below.** The enumeration ("server-side error, timeout, or transport error") reads as exhaustive, but the note still says that when the JSON-RPC call succeeds and an error is returned inside the result payload (including `CallToolResult` with `isError: true`), `error.type` SHOULD be set to `tool_error` / the payload error. On a server span that isn't a server-side error, a timeout, or a transport error, so the two rules disagree about the most common MCP failure mode. Tool failures are server-side, so I'd fold them in:

```suggestion
          conditionally_required: >
            On the client side, if and only if the operation fails. On the server side,
            if and only if the operation fails with a server-side error, an error reported
            in the result payload, a timeout, or a transport error.
```

model/mcp/common.yaml
```diff
@@ -26,10 +26,20 @@ attribute_groups:
...
         note: |
-          This attribute SHOULD be set to the string representation of the JSON-RPC
-          error code, if one is returned.
+          On client spans, `error.type` SHOULD be set for any failed request or error response,
+          including both client-side and server-side errors.
+
+          On server spans, `error.type` SHOULD NOT be set for client-side errors. See
+          `rpc.response.status_code` for which error codes are considered server-side.
```
trask: Same "spans" problem here — this note is rendered into the `mcp.client.operation.duration` and `mcp.server.operation.duration` tables too, where "client spans" / "server spans" doesn't apply.

```suggestion
          On the client side, `error.type` SHOULD be set for any failed request or error
          response, including both client-side and server-side errors.

          On the server side, `error.type` SHOULD NOT be set for client-side errors. See
          `rpc.response.status_code` for which error codes are considered server-side.
```

model/mcp/spans.yaml
```diff
@@ -43,11 +43,11 @@ spans:
       as `target` in the span name when it is available but SHOULD NOT include it by default
       to avoid high cardinality span names.
 
-      **Span status** SHOULD be set to `ERROR` when `error.type` attribute is present.
-      The status description SHOULD match the `JSONRPCError.message` if the message is available.
+      **Span status**: refer to the [Recording Errors](https://github.com/open-telemetry/semantic-conventions/blob/v1.41.0/docs/general/recording-errors.md)
+      document for details on how to record span status. See also `rpc.response.status_code` attribute
+      for the details on which values classify as errors.
```
trask: These lines get rewritten anyway, and the `error.type` link this PR adds a few lines down in `model/mcp/common.yaml` points at `v1.43.0`. Worth pinning them consistently — `docs/general/recording-errors.md` exists at `v1.43.0`. Same on line 119 for the server span.

```suggestion
      **Span status**: refer to the [Recording Errors](https://github.com/open-telemetry/semantic-conventions/blob/v1.43.0/docs/general/recording-errors.md)
      document for details on how to record span status. See also `rpc.response.status_code` attribute
      for the details on which values classify as errors.
```

model/mcp/spans.yaml
```diff
@@ -61,6 +61,9 @@ spans:
     attributes:
       - ref_group: trace.mcp.common.attributes
       - ref_group: mcp.client.address_and_port.attributes
+      - ref: rpc.response.status_code
+        note: |
+          All JSON-RPC error codes SHOULD be considered errors.
```
trask: The classification only exists on the two span definitions, but the `error.type` note that points at it lives in `mcp.common.attributes`, which `mcp.operation.metrics.attributes` also references. In the generated `mcp.client.operation.duration` / `mcp.server.operation.duration` tables the result is a dangling reference: `error.type` says "See `rpc.response.status_code` for which error codes are considered server-side", while `rpc.response.status_code` still carries the generic upstream note ("Usually it represents an error code... Semantic conventions for individual RPC frameworks SHOULD document what `rpc.response.status_code` means..."). So `error.type` on `mcp.server.operation.duration` — the signal that most needs this rule — is unimplementable as written.

Can't express a cross-file suggestion, so concretely: move both overrides into shared internal groups in `model/mcp/common.yaml`, e.g.

```yaml
  - id: mcp.client.status.attributes
    visibility: internal
    attributes:
      - ref: rpc.response.status_code
        note: |
          All JSON-RPC error codes SHOULD be considered errors.

  - id: mcp.server.status.attributes
    visibility: internal
    attributes:
      - ref: rpc.response.status_code
        note: |
          <server-side classification>
```

then replace the inline overrides in `spans.yaml` with `- ref_group: mcp.client.status.attributes` / `- ref_group: mcp.server.status.attributes`, and reference the matching group from each metric in `model/mcp/metrics.yaml` (`mcp.operation.metrics.attributes` is shared by both metrics today, so either split it or add the `ref_group` per metric).

Separately, worth a second look at "all error codes" on the client: 2025-11-25 defines `-32042` (`URLElicitationRequiredError`), which tells the client to run a URL elicitation and retry. That's control flow, not a failure, and this rule turns it into an `ERROR` client span.

model/mcp/spans.yaml
```diff
@@ -113,14 +116,20 @@ spans:
...
+      If the span status is set to `ERROR`, the status description SHOULD match the `JSONRPCError.message` if the message is available.
     stability: development
     attributes:
       - ref_group: trace.mcp.common.attributes
+      - ref: rpc.response.status_code
+        note: |
+          The following status codes SHOULD be considered errors:
+
+          - `-32603` (`Internal error`)
+          - Any error code in the range `-32000` to `-32099` (`Server error`)
```
trask: Two problems with this list.

**`-32002` is misclassified.** MCP tells servers to return `-32002` for "Resource not found" ([2025-11-25 `server/resources` → Error Handling](https://modelcontextprotocol.io/specification/2025-11-25/server/resources#error-handling), same in 2025-06-18). It sits inside `-32000`..`-32099`, so this rule marks the MCP equivalent of an HTTP 404 as a server-side failure — the exact case the motivation says we want to stop flagging. Every `resources/read` for a URI the caller got wrong will set `error.type` and `ERROR` status on the server span and count as a failure in `mcp.server.operation.duration`.

**The allowlist is closed, so unrecognized codes default to "not an error".** MCP and SDKs define codes outside the `-3200x` block (2025-11-25 adds `-32042` `URLElicitationRequiredError`), and applications define their own. Under this rule a genuine server failure carrying a code the instrumentation doesn't recognize is silently recorded as success. HTTP defaults the other way: "any other code the client failed to interpret" is an error.

Inverting the list fixes both, and the excluded set becomes exactly the caller-fault codes:

```suggestion
      - ref: rpc.response.status_code
        note: |
          The following error codes indicate an error on the caller side and
          SHOULD NOT be considered errors:

          - `-32700` (`Parse error`)
          - `-32600` (`Invalid Request`)
          - `-32601` (`Method not found`)
          - `-32602` (`Invalid params`)
          - `-32002` ([`Resource not found`](https://modelcontextprotocol.io/specification/2025-11-25/server/resources#error-handling))

          Any other error code SHOULD be considered an error.
```

If you'd rather not blanket-classify unknown codes as errors, then at minimum `-32002` needs to come out of the server range, and the note should say what an instrumentation does with a code matching none of the listed ranges.

trask (commented):

model/mcp/spans.yaml
lmolkova: addressed but not URLElicitationRequiredError - when HTTP returns 429 and asks to retry after we mark it as error. when server returns 401 and we refresh token and retry on client, it's also an error. so this one is a client error too. request was not processed - an error.

model/mcp/common.yaml
```diff
@@ -40,11 +44,43 @@ attribute_groups:
...
+      - ref: rpc.response.status_code
+        brief: The error code from the JSON-RPC response.
+        requirement_level:
+          conditionally_required: If response contains an error code.
+        note: |
+          This attribute records the [JSON-RPC error code](https://www.jsonrpc.org/specification#error_object)
+          returned in the error response.
+
+          The following error codes indicate that the caller sent a request the
+          server could not serve and SHOULD NOT be considered errors:
```
trask: This group also lands on the `mcp.server` span emitted by the MCP *client* when the server initiates the operation (`sampling/createMessage`, `roots/list`, elicitation) — the span note in `spans.yaml` calls that direction out explicitly. There the party that couldn't serve the request is the MCP client, so "the server" reads as if the rule is scoped to the MCP endpoint role rather than to the receiving side of the call. "Receiver" stays true both ways.

```suggestion
          The following error codes indicate that the caller sent a request the
          receiver could not serve and SHOULD NOT be considered errors:
```

trask (approved):

model/mcp/common.yaml
aabmass: It should still be recorded on `rpc.response.status_code` in this case right?

aabmass (approved):

model/mcp/common.yaml
lmolkova: yep, this attribute is populated regardless and this paragraph is only about which of them are errors.

PR #410, opened 2026-07-29
Allow conversation ID on workflow spans

### LINKED ISSUE #409: Allow `gen_ai.conversation.id` on `invoke_workflow` spans for conversation-scoped evaluation
### Problem

When a workflow invocation represents a conversation turn, backends need a stable identifier to correlate it with other turns in the same conversation. However, `invoke_workflow` spans do not currently allow `gen_ai.conversation.id`, even though they can capture the turn input and final output through `gen_ai.input.messages` and `gen_ai.output.messages`.

As a result, workflow invocations cannot be directly grouped by conversation for conversation-scoped evaluation.

### Motivating example

A conversational travel-planning workflow may create one trace per user turn:

```text
Trace A
└── invoke_workflow travel_planner
      gen_ai.conversation.id = conv_abc
      gen_ai.input.messages  = [user: "Find flights to Paris"]
      gen_ai.output.messages = [assistant: "I found three options..."]
      ├── invoke_agent flight_search
      └── chat

Trace B
└── invoke_workflow travel_planner
      gen_ai.conversation.id = conv_abc
      gen_ai.input.messages  = [user: "Also find a hotel"]
      gen_ai.output.messages = [assistant: "Here are matching hotels..."]
      ├── invoke_agent hotel_search
      └── chat
```

The shared conversation identifier would allow a backend to group the workflow turns and evaluate properties such as cross-turn consistency, use of prior context, and overall task completion. The current `invoke_workflow` span convention does not permit recording this identifier.

### Proposed change

Add `gen_ai.conversation.id` to the `gen_ai.invoke_workflow.internal` span:

```yaml
- ref: gen_ai.conversation.id
  requirement_level:
    conditionally_required: >
      If the workflow invocation is unambiguously associated with a single
      conversation and the conversation identifier is available.
```

This does not require `gen_ai.conversation.id` on every workflow. Workflows that are not associated with a conversation, or that span multiple conversations, should omit it.

The existing guidance for `gen_ai.conversation.id` should continue to apply: instrumentation must not synthesize a fallback identifier from a trace ID, random UUID, request hash, or similar value when the application or framework does not expose one.

### Related issues

- #356 proposes identifying the root span of each conversation turn and includes `invoke_workflow` as an eligible turn root. Its proposed query relies on `gen_ai.conversation.id`, but that attribute is not currently part of the workflow span convention.
- #79 discusses session/trace-level evaluation and grouping multi-turn conversations by `gen_ai.conversation.id`.
- #51 discusses the distinction between application sessions and GenAI conversations.

steverao:
## Description

Allow `gen_ai.conversation.id` on `invoke_workflow` spans when a conversation identifier is available. The existing guidance against synthesizing fallback identifiers also applies to workflow spans.

This PR also updates the Google ADK reference scenario to populate the workflow span from the library-owned `session.id`, regenerates the workflow coverage report and documentation, and adds a Towncrier fragment.

Fixes #409.

## Motivation

**User journey:** conversation-scoped quality evaluation.

A workflow invocation can represent the externally observable boundary of one conversation turn and already supports opt-in `gen_ai.input.messages` and `gen_ai.output.messages`. Without `gen_ai.conversation.id` on that span, a backend cannot directly group workflow turns across traces to evaluate cross-turn consistency, use of prior context, or overall task completion.

**Prior art:** Google ADK models a session as the conversation thread and exposes `session.id` to `Runner.run_async`. The OpenAI Agents SDK similarly supports SDK-managed sessions and provider-managed conversation IDs. `gen_ai.conversation.id` is already used on inference and agent invocation spans; this change makes it available at the workflow boundary as well.

- [Google ADK sessions](https://adk.dev/sessions/)
- [OpenAI Agents SDK sessions](https://openai.github.io/openai-agents-python/sessions/)

## Prototype

The Google ADK reference scenario now records the current `session.id` on the `invoke_workflow` span. The regenerated coverage report shows Google ADK support for `gen_ai.conversation.id` on workflow spans.

The CrewAI reference scenario is unchanged because its existing non-conversational Crew invocation does not expose a conversation identifier at the current call boundary; no value is synthesized or hardcoded.

Validation:

- `uv run run-scenario google-adk`
- `make generate-all`
- `make check-policies`
- `uv tool run --from ruff ruff check reference/scenarios/google-adk/scenario.py`
- `uv tool run --from ruff ruff format --check reference/scenarios/google-adk/scenario.py`
- `git diff --check`

## Checklist

See [CONTRIBUTING.md]


model/gen-ai/spans.yaml
```diff
@@ -598,6 +598,26 @@ spans:
...
+          conditionally_required: When available.
+        note: |
+          Instrumentations SHOULD populate conversation id when they have an identifier
+          for the conversation readily available for a given operation, for example:
+
+          - when client framework being instrumented manages conversation history
+          (see [LlamaIndex chat store](https://docs.llamaindex.ai/en/stable/module_guides/storing/chat_stores/))
+          - when instrumenting GenAI client libraries that maintain conversation on the backend side
+          (see [AWS Bedrock agent sessions](https://docs.aws.amazon.com/bedrock/latest/userguide/agents-session-state.html),
+          [OpenAI Assistant threads](https://platform.openai.com/docs/api-reference/threads))
```
lmolkova: I think this belongs in the attribute definition rather than in reference on the workflow. 

consider also mentioning Langchain with session_id https://reference.langchain.com/python/langchain-core/runnables/history/RunnableWithMessageHistory

and ADK with https://adk.dev/sessions/session

model/gen-ai/spans.yaml
```diff
@@ -598,6 +598,26 @@ spans:
       - ref: gen_ai.workflow.name
         requirement_level:
           conditionally_required: When available.
+      - ref: gen_ai.conversation.id
+        requirement_level:
+          conditionally_required: When available.
```
```suggestion
          conditionally_required: If and only if instrumented library has one readily available or when user application provided it though OpenTelemetry context or library-specific mechanisms.
          
```

model/gen-ai/spans.yaml
```diff
@@ -598,6 +598,26 @@ spans:
...
+          (see [AWS Bedrock agent sessions](https://docs.aws.amazon.com/bedrock/latest/userguide/agents-session-state.html),
+          [OpenAI Assistant threads](https://platform.openai.com/docs/api-reference/threads))
+
+          When no identifier for the conversation is available, instrumentations SHOULD NOT
+          populate conversation id. For example, a new UUID, a trace identifier, or a hash
+          of request content SHOULD NOT be used as a fallback value.
+
+          Application developers that manage conversation history MAY add conversation id to GenAI and other
+          spans or logs using custom span or log record processors or hooks provided by instrumentation
+          libraries.
```
```suggestion
```

Suggested a shorter form on the requirement_level

lmolkova (approved):

model/gen-ai/spans.yaml
```diff
@@ -112,23 +112,6 @@ attribute_groups:
       - ref: gen_ai.conversation.id
         requirement_level:
           conditionally_required: When available.
```
trask: The same attribute now ends up with two different conditions. This ref and the `invoke_agent` one below still say `When available.`, while the new `invoke_workflow` ref spells out exactly when to set it. You can see it in the generated docs: the workflow table puts its condition in a footnote, every other `gen_ai.conversation.id` row just says `When available.` inline.

Now that the shared note lives on the attribute itself, it'd be good for the requirement level to read the same way everywhere. Same change for the `invoke_agent` ref around line 226.

```suggestion
          conditionally_required: If and only if the instrumented library has one readily available, or the user application provides one through OpenTelemetry context or library-specific mechanisms.
```

trask (approved):

lmolkova (approved):

PR #415, opened 2026-07-31
Track client token usage and operation duration in the reference report

AgentGymLeader:
Split out of #324, which is now migration-only.

`METRIC_SPECS` doesn't include `gen_ai.client.token.usage` or
`gen_ai.client.operation.duration`, so they never reach `data.json` or the
coverage report. Adding them surfaces two libraries that already emit them:

| Metric | Libraries |
| --- | --- |
| Client Token Usage | agent-framework, anthropic |
| Client Operation Duration | agent-framework, anthropic |

Both reach `data.json` through the parser change alone. The google-adk
scenario change in this PR is separate: it stops ADK's fallback instruments
from crediting the delegated model call to google-adk, which is why it no
longer appears above.

`METRIC_SPECS` carries a comment from #336 marking these two as deliberately
omitted, "trivially derivable from the same information already present on
spans". That holds: both are a single-span read, unlike the `invoke_agent`
counters #336 added, which need each call attributed to exactly one invocation
across the call tree.

It reads as a rule for the whole list, though, and the list records what the
reference scenarios emit rather than what instrumentations should emit.
agent-framework and anthropic emit both metrics today and the report says
nothing either way, so this PR re-scopes the comment instead of dropping it.

@RKest — flagging since this rewrites your comment from #336.


reference/src/semconv_genai/semconv_model.py
```diff
@@ -217,10 +217,22 @@ def _from_yaml(
...
 }
 
-# The `gen_ai.client.operation.duration` and `gen_ai.client.token.usage`
-# metrics are deliberately omitted: they are trivially derivable from the same
-# information already present on spans.
+# `gen_ai.client.operation.duration` and `gen_ai.client.token.usage` are a
+# single-span read, unlike the `invoke_agent` counters below, which need each
+# call attributed to exactly one invocation across the call tree (#336). They
+# are tracked here because google-adk and agent-framework emit them; this list
+# records what the reference scenarios emit, not what instrumentations should.
```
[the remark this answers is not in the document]
AgentGymLeader: Fixed — "not what instrumentations should emit."

reference/src/semconv_genai/parse_results.py
```diff
@@ -67,35 +74,43 @@ def _attribute_names(
...
+                data_point,
+            )
+        return data_points
+
+    _LOGGER.warning(
+        "Skipping metric %r with invalid data_points: expected dict or list, got %s",
+        metric.get("name", ""),
+        type(raw_points).__name__,
+    )
+    return None
```
AgentGymLeader: Carrying over an open point from #324 ([original thread](https://github.com/open-telemetry/semantic-conventions-genai/pull/324#discussion_r3687655828)).

@trask found that `data_points` has no `skip_serializing_if`, so the key is always present and `null` when a metric has no data set — which kills the marker branch, the dict branch and the `metric["data"]` lookup, and leaves `null` falling into the catch-all where a valid Weaver serialization gets logged as invalid.

Applied his suggestion as-is in cd07fc2, plus the `test_null_data_points_is_skipped_without_warning` case he wrote — it fails on the old function with the warning he quoted and passes after. Verified against the Weaver this branch pins, v0.25.1.

One deviation: the test uses a local log-capture helper instead of `caplog`, so it also runs under `python tests/test_metrics.py`. CI doesn't run `reference/tests/` today, so that path is the only one that executes it.

reference/README.md
```diff
@@ -46,6 +46,8 @@ Run `uv run update-reports` to regenerate.
 
 | Metric | Libraries |
 | --- | --- |
+| [Client Token Usage](reports/gen-ai-client-token-usage-metric.md) | agent-framework, google-adk |
```
trask: This is missing anthropic. #324 landed on main after this PR's last CI run, so the green anthropic check is stale — the anthropic scenario now goes through `opentelemetry-util-genai`, which records the duration histogram on every inference invocation and token usage whenever token counts are set, as that scenario does. On a rebase, `run-scenario anthropic` will produce a `metrics` block in `scenarios/anthropic/data.json` and the up-to-date check will fail.

Worth doing beyond the CI fix: anthropic is the only library here that passes `server_port`, so both new report pages currently show `server.port | (none)` when they shouldn't. Please rebase on main and re-run `uv run run-scenario anthropic` plus `uv run update-reports`.

reference/scenarios/google-adk/data.json
```diff
@@ -68,6 +68,19 @@
     ]
   },
   "metrics": {
+    "gen_ai.client.token.usage": [
```
trask: Should google-adk be credited with these? ADK records both metrics for the model call it delegates to `google-genai` — the same call whose inference span #351 removed and whose README row says "correctly not emitted". `_suppress_adk_native_tracing` only patches tracers and the `trace_*` helpers, so ADK's own metrics pass through.

Is that intended, or should the suppression cover the metrics too?

reference/README.md
AgentGymLeader: Fixed in 108d88f. Merged main, re-ran `run-scenario anthropic` and `update-reports` — anthropic now has a `metrics` block, and `server.port` went from `(none)` to anthropic on both pages.

reference/scenarios/google-adk/data.json
AgentGymLeader: Neither, I think — it's a conditional fallback.

ADK emits them itself (meter scope `gcp.vertex.agent`; `google/genai/` has no OTel code at all), but the attributes describe the delegated call: `gen_ai.operation.name = generate_content`, `gen_ai.provider.name = gemini`.

And ADK treats them as model-client telemetry. `_should_emit_native_telemetry` returns False when the google-genai instrumentation is active on a Gemini agent, "then the lib already emits inference metrics" — so ADK only emits these when no real client instrumentation is present.

That leaves the span and the metric disagreeing about the same call: one says "delegates, correctly not emitted", the other credits google-adk. I'd rather close that on the suppression side — the helper already patches the tracers and the `trace_*` helpers, so covering the meter is the same kind of change.

Happy to do it here or leave it for a separate change, whichever you prefer.

trask: sounds good, let's do it here, thanks

AgentGymLeader: @trask done in 17402e9. The two client instruments are now disabled alongside the tracers, and the invoke_agent counters are untouched.

The scenario README already said inference here is "correctly not emitted". That was only true of spans until now.

reference/src/semconv_genai/semconv_model.py
```diff
@@ -223,10 +223,22 @@ def _from_yaml(
     ),
 }
 
-# The `gen_ai.client.operation.duration` and `gen_ai.client.token.usage`
-# metrics are deliberately omitted: they are trivially derivable from the same
-# information already present on spans.
+# `gen_ai.client.operation.duration` and `gen_ai.client.token.usage` are a
+# single-span read, unlike the `invoke_agent` counters below, which need each
+# call attributed to exactly one invocation across the call tree (#336). They
+# are tracked here because google-adk and agent-framework emit them; this list
```
trask: google-adk doesn't emit these in this PR — the google-adk scenario change here explicitly disables ADK's `_client_token_usage` and `_client_operation_duration` instruments, and its `data.json` still shows only the two `invoke_agent` metrics. The libraries that actually surface are agent-framework and anthropic, per the README table and the two new report files. The PR description's table has the same mistake and is worth fixing too.

```suggestion
# are tracked here because agent-framework and anthropic emit them; this list
```

trask (approved):

reference/src/semconv_genai/semconv_model.py
AgentGymLeader: @trask applied your suggestion in c205a6d. The description had it in two more places beyond the table, including the "parser change alone" line that cited `run-scenario google-adk` as evidence for both metrics — rewritten to say what the scenario change actually does.

lmolkova (approved):

PR #417, opened 2026-07-31
docs: add governed tool action metadata example

jw-ond:
## Summary

Adds a non-normative GenAI example showing how applications can attach governance metadata to an `execute_tool` span when a tool call is reviewed before execution.

The example uses application-specific attributes (`app.governance.*`) so it does not introduce new normative semantic convention attributes. It illustrates action references, approval verdicts, replay/proof URLs, and external verifier references as telemetry that can correlate tool execution with an approval or policy checkpoint.

## Validation

- Checked Markdown code fence balance locally
- Confirmed the example is vendor-neutral and uses non-normative `app.governance.*` attributes
- Added a clarification changelog fragment


jw-ond:
Updated after review: the example SHA-256 value is now a full 64-character digest, and the changelog fragment has been renamed to 417.clarification.md per changelog.d/README.md. The old +.clarification.md fragment was removed from the branch.

Silentpartnercoding:
One trust boundary may be worth stating explicitly: these `app.governance.*` attributes are application-emitted correlation claims. The span alone does not prove that the approval gate ran, that the executed inputs matched the approved `action.hash`, or that the referenced proof/verifier was independently validated.

Could the example add one sentence such as: “These attributes are correlation claims, not authorization or verification evidence; consumers must independently validate the referenced proof before treating the execution as approved.”

That would preserve the example’s purpose without making an `execute_tool` span look like an authorization receipt.

jw-ond:
Thanks, that trust-boundary wording is right. Updated the example to state that the `app.governance.*` attributes are correlation claims, not authorization or verification evidence, and that consumers must independently validate the referenced approval/verifier/proof material before treating them as governance evidence.

lmolkova (changes requested):
None of the attributes added in the example are existing attributes defined in this repo. Applications are free to add them, but this is not something they should expect from OpenTelemetry GenAI instrumentations, hence it's controversial to have them in examples. 

While the current language says that it's non-normative, the example is very specific and reads more prescriptive than 'attach whatever attributes you want'.

The example can be changed in a more generic way: do not mention specific attributes, use something like `acme.custom.attribute`, emphasize that applications can add them on any spans, tool calls or not.

jw-ond:
Addressed the requested change in the latest commits. The example no longer defines or suggests `app.governance.*` attributes; it now uses a generic `acme.custom.attribute` placeholder, states that custom attributes are outside the OpenTelemetry GenAI semantic conventions, and notes that they can be added to any span, including GenAI client spans, execute-tool spans, or surrounding application spans. It also keeps the trust-boundary wording that custom attributes are application claims, not authorization or verification evidence.

Also fixed the generated TOC anchor after the section rename. All current PR checks are passing.

/dashboard route:reviewers

jw-ond:
Following up after the latest update. The PR now keeps the governed-action fields framed as optional custom metadata rather than asking semantic conventions to standardize OSuite-specific attributes, and the current checks are green.

If this still feels too concrete for the semantic-conventions repo, I can narrow it further or move the example toward a more vendor-neutral cookbook shape.

PR #419, opened 2026-08-01
reference: emit execute_tool only where the library actually runs the tool

lmolkova:
Follow-up to #351

Remove: Generic instrumentation of these clients could never emit it

- `aws-bedrock`, `cohere`, `groq`, `litellm`, `mistralai`, `openai`
- `instructor` deleted — nothing left to emit once removed

Added: the library does run the tool:  `google-genai`, `vertexai`, `langchain`

`execute_tool` coverage drops from 13 libraries to the 11 that genuinely
execute tools.


reference/scenarios/vertexai/scenario.py
```diff
@@ -118,22 +127,41 @@ def run_chat():
...
+
+    def get_weather(location: str) -> str:
+        """Get the current weather.
+
+        Args:
+            location: City name
+        """
+        # The SDK invokes this callable as part of automatic function calling.
+        # The public FunctionCall wrapper exposes only name/args, so the tool call
+        # id is not recorded here.
```
trask: vertexai ends up as the only one of the three execute_tool scenarios added here without `gen_ai.tool.call.id`. That follows from emitting inside `get_weather`: Vertex AI's automatic function calling hands the callable only the model args, so the id never reaches it.

It is available one level up. `AutomaticFunctionCallingResponder.respond_to_model_response` is the dispatch site and has the `FunctionCall` in scope, and `function_call.to_dict().get("id")` picks up the `call_mock_001` the mock server already returns on the v1beta1 path this PR wires up — the v1beta1 `FunctionCall` proto carries `id`, and `Part.function_call` preserves it through `_from_gapic`. Patching there, the same move as `_patch_automatic_function_calling` in the google-genai scenario, would put vertexai in the `gen_ai.tool.call.id` row alongside google-genai and langchain.

trask (approved):

Issue #320, opened 2026-06-18
Agent Harness Hook Semantic Conventions


michaelsafyan:
Various agent harnesses support extension via "hooks". As more and more AI usage leverages "harnesses" rather than "frameworks" and more logic shifts into "hooks", it becomes important to:

1. Ensure that context is propagated into hooks (and that there are standards for _how_ context is propagated to them; for example, is it a standard set of parameters to hooks? Is it via OTel context ENV carriers?)

2. Provide conventions around spans and other telemetry that hook invocation should generate.

Here are a sample of some hooks for consideration:

| Semantic Stage | **Claude Code** | **Google Antigravity SDK** | **OpenCode** | Semantic Description | Documentation Sources |
| :--- | :--- | :--- | :--- | :--- | :--- |
| **Session Initialization** | `SessionStart`<br>`Setup` | Managed via `SessionContext` / Trigger runner start | `session-start` | Fires when an agent session begins, initializes environment states, or prepares a workflow. | [Claude Code Docs](https://code.claude.com/docs/en/hooks)<br>[Antigravity Docs](https://antigravity.google/docs/hooks) |
| **Pre-LLM Invocation** | `UserPromptSubmit`<br>`UserPromptExpansion` | `PreInvocation` | *Implicit turn start* | Fires after a user submits a prompt or before the harness passes the conversation history/context to the LLM. Useful for pre-processing input or appending system guidance. | [Claude Code Docs](https://code.claude.com/docs/en/hooks)<br>[Antigravity Docs](https://antigravity.google/docs/hooks) |
| **Pre-Tool Execution** | `PreToolUse`<br>`PermissionRequest` | `PreToolUse`<br>(`PreToolCallDecideHook`) | `tool.execute.before` | Fires after the model decides to invoke a tool but before it runs. Can intercept parameters, block execution, or request explicit human confirmation. | [Claude Code Docs](https://code.claude.com/docs/en/hooks)<br>[Antigravity Docs](https://antigravity.google/docs/hooks) |
[...]

PR #427, opened 2026-08-04
gen-ai: add run guardrail span and security finding

habibam:
## Description

Defines semantic conventions for observing GenAI **guardrails** policy, security findings and other controls that inspect GenAI content or actions before, during, or after a GenAI operation.

Adds:

- `gen_ai.run_guardrail.client` for out-of-process guardrail services.
- `gen_ai.run_guardrail.internal` for in-process guardrail checks.
- `gen_ai.security.finding` event for findings produced by a guardrail run.
- `gen_ai.security.guardrail.*` (evaluating service/component),
  `gen_ai.security.verdict.*` (result returned), `gen_ai.security.action.type`
  (caller enforcement), `gen_ai.security.policy.*` /
  `gen_ai.security.policy.rule.id` (policy/rule attribution),
  `gen_ai.security.risk.*` (classification/scoring), and
  `gen_ai.security.external_finding_id` (SIEM correlation).
- OpenAI reference coverage for the client/internal guardrail spans and finding.

Successor to [#262](https://github.com/open-telemetry/semantic-conventions-genai/pull/262) by Nagkumar Arkalgud.

## Motivation

**User journey:** security investigation and incident triage. When a GenAI application blocks, modifies, or escalates a request, responders need to answer from telemetry alone which guardrail ran, what it evaluated, what verdict it returned, what the caller actually enforced, which policy/rule fired, and how to correlate the finding with the external security system (SIEM). Also supports sampling decisions (verdict/action) and audit.

Separating **verdict** (the guardrail's judgment) from **action** (the caller's enforcement) lets telemetry represent cases where a guardrail returns one result but the caller enforces another (e.g. verdict `deny`, action `escalate`).

**Prior art:** Google's Model Armor emits these guardrail spans in production ([public reference](https://docs.cloud.google.com/gemini-enterprise-agent-platform/govern/view-model-armor-spans)).
Content-safety services, DLP filters, prompt-injection detectors, and policy engines all expose equivalent concepts (guardrail identity, verdict, policy, finding) today in vendor-specific forms.

## Prototype

The reference scenarios under `reference/scenarios/openai/` are updated to emit both the `run_guardrail` client/internal spans and the `gen_ai.security.finding` event, with generated reports under `reference/reports/`.

**Caveat / follow-up:** the current scenario is hand-authored rather than produced by an auto-instrumented guardrail library. Per @lmolkova top level review on [#262](https://github.com/open-telemetry/semantic-conventions-genai/pull/262), a survey of instrumentable open-source server-side guardrail libraries and adding a real one to the reference scenarios is the main remaining work item and is tracked as an open question tobe addressed at the SIG meeting.

## Checklist


AgentGymLeader:
Saw the open item in your description — a real instrumented guardrail library in the reference scenarios, rather than hand-authored spans.

I have one running in #368. `reference/scenarios/agent-governance/` drives the OpenAI Agents SDK with a plain-function input guardrail: one allowed run that executes a tool, and one tripwire run with zero model calls and no child execute span. The decision attributes are read off the SDK's own guardrail runtime objects, so it is genuine instrumentation rather than spans written by hand.

It currently sits on a draft PR of mine, but nothing in the scenario is specific to my proposal — it is just a guardrail library being instrumented for real. Happy to move it here, or leave it where it is and reference it from this PR. Whichever is less work for you.

One practical note while you are in here: EasyCLA is still failing on this PR, and reviewers generally will not start until that goes green.

sjain700 (approved):

AgentGymLeader:
My offer above is moot now that the PR has a real-library scenario; consider it withdrawn. No reply needed.

docs/gen-ai/gen-ai-guardrail-systems.md
```diff
@@ -0,0 +1,134 @@
+<!--- Hugo front matter used to generate the OpenTelemetry.io documentation --->
```
habibam, line 1: @lmolkova this is the systems research doc you requested in your top level comment on #262. It addresses your ask to list the systems evaluated and share how their properties map to the attributes being defined. It covers:

• Systems surveyed - OpenAI Guardrails, NeMo Guardrails, Guardrails AI, AWS Bedrock Guardrails, Azure AI Content Safety, and standalone detection engines.
• What each exposes at an instrumentable boundary.
• Property-to-attribute mapping - base  gen_ai.guardrail.*  plus the  security.*  overlay.
• Capturability findings that motivate keeping the security overlay separable from the base guardrail attributes.

Would appreciate a review when you get a chance.

model/gen-ai/gen-ai-security-content.json
```diff
@@ -0,0 +1,4 @@
+{
+    "description": "Represents content evaluated or produced by a security guardrail.",
+    "title": "SecurityGuardrailContent"
+}
```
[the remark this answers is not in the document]
habibam: Addressed, added a permissive root  type  of  ["string", "object", "array", "boolean", "number", "null"]  so the schema now imposes a real (if permissive) structure and the "MUST follow schema" guidance is meaningful.

reference/scenarios/openai/scenario.py
```diff
@@ -511,6 +515,90 @@ def get_weather(location: str) -> str:
...
+    }
+    if host and port is not None:
+        guardrail_attrs["server.address"] = host
+        guardrail_attrs["server.port"] = port
+    with _reference_tracer.start_as_current_span(
+        "run_guardrail Azure Content Safety",
+        kind=SpanKind.CLIENT,
+        attributes=guardrail_attrs,
+    ):
+        with urlopen(health_url, timeout=5) as response:
```
lmolkova, line 597: The client span has no reference coverage. This calls the mock server's `/health` endpoint rather than a guardrail API, and every guardrail result attribute is a constant. The generated client-span report consequently lists `(none)` for every attribute.

Could this use a real client operation such as [AWS `ApplyGuardrail`](https://docs.aws.amazon.com/sdk-for-python/v1/reference/clients/bedrock-runtime/operations/apply_guardrail/), [Azure `analyze_text`](https://learn.microsoft.com/en-us/python/api/azure-ai-contentsafety/azure.ai.contentsafety.contentsafetyclient?view=azure-python), or [Model Armor `sanitize_user_prompt`](https://docs.cloud.google.com/model-armor/reference/rest/v1/projects.locations.templates/sanitizeUserPrompt), with values read from the request and response?

model/gen-ai/spans.yaml
```diff
@@ -611,6 +665,111 @@ spans:
...
+        brief: Guardrail service address.
+        requirement_level: recommended
+        sampling_relevant: true
+      - ref: server.port
+        brief: Guardrail service port.
+        requirement_level:
+          conditionally_required: If `server.address` is set.
+        sampling_relevant: true
+
+  - type: gen_ai.run_guardrail.internal
```
lmolkova, line 725: The SDK survey shows that dedicated in-process guardrail APIs exist, but they are not a general agent-framework primitive:

| Library | Dedicated guardrail API | Observable boundary |
| --- | --- | --- |
| [OpenAI Agents](https://openai.github.io/openai-agents-python/guardrails/) | Yes | Input, output, tool-input, and tool-output guardrails return dedicated result types |
| [OpenAI Guardrails](https://openai.github.io/openai-guardrails-python/) | Yes | `GuardrailsOpenAI` runs configured input/output guardrails |
| [Guardrails AI](https://github.com/guardrails-ai/guardrails#-quick-start) | Yes | `Guard.validate()` returns validation outcomes |
| [NeMo Guardrails](https://docs.nvidia.com/nemo/guardrails/latest/index.html) | Partial | `LLMRails.generate()` runs the complete guarded generation, but does not expose an independent per-rail operation at the public boundary |
| [Claude Agent SDK](https://code.claude.com/docs/en/agent-sdk/python#hooks) | No | General lifecycle and tool hooks |
| [Google ADK](https://adk.dev/callbacks/types-of-callbacks/) | No | General agent, model, and tool callbacks |
| [LangChain](https://docs.langchain.com/oss/python/langchain/middleware) | No | General agent, model, and tool middleware |

The current reference uses Claude `PreToolUse`, so it demonstrates a general hook rather than a dedicated guardrail boundary. Hook classification and telemetry are being handled separately in [#320](https://github.com/open-telemetry/semantic-conventions-genai/issues/320). The dedicated APIs can be covered later without treating every hook as a guardrail.

model/gen-ai/spans.yaml
```diff
@@ -611,6 +665,111 @@ spans:
       - ref: gen_ai.tool.call.result
         requirement_level: opt_in
 
+  - type: gen_ai.run_guardrail.client
```
lmolkova, line 668: This client span defines a new operation class without its corresponding duration metric.

[#249](https://github.com/open-telemetry/semantic-conventions-genai/issues/249) defines a 1:1 relationship between span definitions and duration metrics. It specifically sketches:

| Span type | Duration metric |
| --- | --- |
| `gen_ai.apply_guardrail.client` | `gen_ai.client.apply_guardrail.duration` |
| `gen_ai.apply_guardrail.internal` | `gen_ai.apply_guardrail.duration` |

Under that design, the client convention added here needs a dedicated client duration metric with the same operation-class segment.

model/gen-ai/events.yaml
```diff
@@ -38,6 +38,54 @@ events:
           completion being evaluated. This attribute helps correlate the evaluation
           event with the corresponding operation when span id is not available.
 
+  - name: gen_ai.guardrail.security.finding
```
lmolkova, line 41: The researched guardrail APIs expose different result shapes:

| API | Available result properties | Not available |
| --- | --- | --- |
| [AWS `ApplyGuardrail`](https://docs.aws.amazon.com/sdk-for-python/v1/reference/clients/bedrock-runtime/operations/apply_guardrail/) | Overall `action`, `actionReason`, output content, assessments, guardrail ID/version, `source` | Caller enforcement |
| [AWS `InvokeGuardrailChecks`](https://docs.aws.amazon.com/bedrock/latest/userguide/guardrails-use-invoke-guardrail-checks-using.html) | Per-check category/entity, severity or confidence score, message/content location | Overall verdict, policy ID, modified output, caller action |
| [Azure `analyze_text`](https://learn.microsoft.com/en-us/python/api/azure-ai-contentsafety/azure.ai.contentsafety.models.analyzetextresult?view=azure-python) | Per-category severity and blocklist matches | Overall verdict, reason, policy ID, caller action |
| [Azure Prompt Shields](https://learn.microsoft.com/en-us/rest/api/contentsafety/text-operations/shield-prompt?view=rest-contentsafety-2024-09-01) | `attackDetected` for the prompt and each document | Score, policy ID, reason, caller action |
| [Google Model Armor](https://docs.cloud.google.com/python/docs/reference/google-cloud-modelarmor/latest/google.cloud.modelarmor_v1.types.SanitizationResult) | Match state, invocation state, per-filter results, errors, optional deidentified content | Caller enforcement |
| [OpenAI Agents](https://openai.github.io/openai-agents-python/guardrails/) | Guardrail name, tripwire state, output information, dedicated input/output result collections | Common policy or security-risk identity |

A security finding with required risk and policy attributes cannot represent several of these result types. A `gen_ai.guardrail.result` event with properties conditional on what the API reports covers both remote and dedicated SDK guardrails.

model/gen-ai/registry.yaml
```diff
@@ -586,6 +586,285 @@ attributes:
...
+  - key: gen_ai.guardrail.component.name
+    type: string
+    brief: Human readable name of the guardrail service or component.
+    note: >
+      This identifies the guardrail evaluation service or component, not the
+      policy. For the policy that produced the verdict, use
+      `gen_ai.guardrail.security.policy.id`.
+    examples: ['Azure Content Safety', 'Bedrock Guardrails', 'Custom PII Filter']
+    stability: development
+  - key: gen_ai.guardrail.verdict.type
```
lmolkova, line 599: The enum combines service results that have different semantics:

- AWS `GUARDRAIL_INTERVENED` means the configured guardrail intervened.
- Model Armor `MATCH_FOUND` means at least one filter matched, while `invocation_result` separately reports execution success.
- Azure severity is a score. It becomes allow or deny only after the application applies a threshold.
- AWS inline checks have no overall result to translate.

Consequently, `allow`, `deny`, and `modify` would be inferred application decisions for some providers and direct service values for others. Telemetry using the same enum member would not describe the same observed fact across providers. Preserving the native outcome keeps the value attributable to the instrumented API.

model/gen-ai/registry.yaml
```diff
@@ -586,6 +586,285 @@ attributes:
...
+          brief: The caller escalated the content or action for human review, approval, or another workflow.
+          stability: development
+    brief: The actual enforcement action taken by the caller or framework.
+    note: >
+      Use this attribute when the guardrail verdict is enforced by the caller or
+      framework and the action is known. For example, a guardrail may return a
+      `warn` verdict while the caller chooses to `block` the operation.
+    examples: ['allow', 'block', 'modify', 'escalate']
+    stability: development
+  - key: gen_ai.guardrail.target.type
```
lmolkova, line 680: `type` and `subtype` divide the target along two unrelated dimensions, but neither name identifies which dimension it represents. The resulting combinations are also artificial:

- `llm + input` means a user prompt, not an LLM.
- `agent + input` still means input supplied to an agent.
- `tool_call + output` may mean either the tool result or a rejected tool invocation.

Providers expose different native concepts:

- [AWS `ApplyGuardrail`](https://docs.aws.amazon.com/sdk-for-python/v1/reference/clients/bedrock-runtime/operations/apply_guardrail/) exposes `source=INPUT|OUTPUT`.
- [Model Armor](https://docs.cloud.google.com/model-armor/reference/rest/v1/projects.locations.templates/sanitizeUserPrompt) exposes separate `sanitizeUserPrompt` and `sanitizeModelResponse` operations.
- [OpenAI Agents](https://openai.github.io/openai-agents-python/guardrails/) exposes input and output guardrails, plus distinct [tool-input and tool-output guardrails](https://openai.github.io/openai-agents-python/ref/tool_guardrails/).
- Azure `analyze_text` does not identify whether the text is an LLM, agent, message, or tool target.

None of these APIs reports `llm`, `agent`, or `message` as a target subtype. Those values would be invented by instrumentation, while consumers would still need the provider-native target value.

model/gen-ai/registry.yaml
```diff
@@ -586,6 +586,285 @@ attributes:
...
+      - 'PII detected in output'
+      - 'Prompt injection attempt detected'
+      - 'Action exceeds agent permission scope'
+    stability: development
+  - key: gen_ai.guardrail.verdict.code
+    type: string
+    brief: Provider-specific code for the guardrail verdict.
+    examples: ['112', '403', 'AACS-PII', 'MODEL_ARMOR_SAFETY']
+    stability: development
+  - key: gen_ai.guardrail.action.type
```
lmolkova, line 650: `gen_ai.guardrail.action.type` describes an action taken by the caller, but remote client instrumentation observes only the service request and response:

| API | Observable service result |
| --- | --- |
| [AWS `ApplyGuardrail`](https://docs.aws.amazon.com/sdk-for-python/v1/reference/clients/bedrock-runtime/operations/apply_guardrail/) | `action` reports whether the service intervened. It does not report what the application did afterward. |
| [AWS `InvokeGuardrailChecks`](https://docs.aws.amazon.com/bedrock/latest/userguide/guardrails-use-invoke-guardrail-checks-using.html) | Returns per-check categories, entities, and scores without an enforcement action. |
| [Azure `AnalyzeTextResult`](https://learn.microsoft.com/en-us/python/api/azure-ai-contentsafety/azure.ai.contentsafety.models.analyzetextresult?view=azure-python) | Returns category severities and blocklist matches without the caller's decision. |
| [Google Model Armor `SanitizationResult`](https://docs.cloud.google.com/python/docs/reference/google-cloud-modelarmor/latest/google.cloud.modelarmor_v1.types.SanitizationResult) | Returns filter match and invocation states without the caller's decision. |

The application can block, redact, retry, escalate, or ignore the same service result. Therefore this value is only observable at an application or framework enforcement boundary, not from generic instrumentation of the client call.

model/gen-ai/registry.yaml
```diff
@@ -586,6 +586,285 @@ attributes:
...
+    stability: development
+  - key: gen_ai.guardrail.target.id
+    type: string
+    brief: Identifier of the specific target the guardrail is applied to.
+    note: >
+      For example, a tool call identifier, request identifier, memory record
+      identifier, or message identifier, if applicable.
+    examples: ['call_xyz789', 'msg_abc123', 'mem_abc456']
+    stability: development
+  - key: gen_ai.guardrail.security.risk.category
```
lmolkova, line 746: Policy, risk, content, and findings are already scoped to a guardrail. The additional `security` level does not identify a different object.

Could these use `gen_ai.guardrail.risk.*`, `gen_ai.guardrail.policy.*`, `gen_ai.guardrail.content.*`, and `gen_ai.guardrail.external_finding_id`?

model/gen-ai/spans.yaml
```diff
@@ -167,6 +167,60 @@ attribute_groups:
         requirement_level:
           conditionally_required: When applicable.
```
lmolkova: Could the guardrail attribute groups and signal definitions move to `model/gen-ai/guardrails.yaml`? This adds about 160 lines to the general span model and will grow as provider mappings are added.

docs/gen-ai/gen-ai-guardrail-systems.md
```diff
@@ -0,0 +1,134 @@
...
+| `gen_ai.guardrail.security.external_finding_id` | External finding record id | Services that return a finding id or emit to a SIEM |
+
+## Capturability findings
+
+Classifying each mapping by how directly the boundary yields it (`direct` when it
+is readable at the boundary, `derivable` from library-owned semantics, `weak`
+when it needs app-specific naming or an enum guess, `capture gap` when the
+boundary cannot produce it) gives a consistent picture across systems:
+
+- **Base attributes are `direct` or `derivable`.** Stage gives `target.type`; the
```
lmolkova, line 108: The capturability conclusion conflicts with the systems described above:

- Azure severity becomes a verdict only after applying an application threshold.
- Detection engines require the caller to supply target, action, and policy.
- AWS APIs expose different result shapes, and inline checks have no overall verdict.

Could this section classify each attribute separately for client and internal boundaries instead of stating that all base attributes are direct or derivable?


Or maybe we can remove it and just make sure the important details are covered in actual conventions rather than here?

docs/gen-ai/gen-ai-guardrail-systems.md
```diff
@@ -0,0 +1,134 @@
+<!--- Hugo front matter used to generate the OpenTelemetry.io documentation --->
+# GenAI Guardrail Systems: Evaluated Systems and Attribute Mapping
+
+**Status**: [Development][DocumentStatus]
+
+This is a non-normative companion to the [GenAI guardrail conventions](gen-ai-security.md).
```
lmolkova, line 6: if there are some non-normative design considerations that are useful to keep, they should be in the non-normative folder

lmolkova (changes requested):
Thanks! I believe internal guardrail design is more tricky since most libraries provide generic hooks not limited to guardrails, we can't record custom hook created e.g. to log things as a guardrail.

So one suggestion is to limit this PR to remote guardrails and to SDKs that provide dedicated internal guardrail API. Fo the latter, the span might not even make sense and I'd propose to start with event capturing the result only.

Another important step is to write reference scenarios that show how to capture client guardrail spans and events for 2+ providers. Please use or read the skill in this repo for this for the instructions on how to do it to ensure capturability.

PR #434, opened 2026-08-04
Keep upstream doc links pinned to the version in `model/manifest.yaml`

trask:
Links from model briefs and notes into `open-telemetry/semantic-conventions` are hardcoded at a specific ref, so they drift from the pinned dependency on every bump. After #433 moved the pin to v1.44.0, `docs/gen-ai/mcp.md` ended up rendering v1.44.0 links in the generated tables next to v1.43.0 links in the notes right beside them. A couple of even older v1.40.0 links were sitting in the hand-written prose too.

This adds `.github/scripts/update-upstream-links.sh`, which rewrites the version in every such link to whatever `model/manifest.yaml` pins. It runs as part of `make generate-all`, and CI's existing generated-output sync check now covers `model/` and `AGENTS.md` — so the next dependency bump can't merge with stale links. Same shape as upstream's [`update-spec-repo-links.sh`](https://github.com/open-telemetry/semantic-conventions/blob/main/.github/scripts/update-spec-repo-links.sh), except it reads the pinned version instead of polling for the latest release.

An earlier revision of this PR put a `{{upstream_docs_base}}` placeholder in the model YAML and let the markdown templates substitute it. That doesn't work: `weaver registry resolve` copies brief/note text verbatim into the published `resolved.yaml`, so the raw placeholder leaked out to every downstream repo that reads the registry directly. The YAML has to carry real URLs.

All 14 rewritten URLs were checked to exist at v1.44.0, and the script is idempotent.

### DIFF
--- a/.github/instructions/model-style.instructions.md
+++ b/.github/instructions/model-style.instructions.md
@@ -10,8 +10,9 @@ first word and end with a period.
 
 # Links to upstream docs
 
-Model YAML is data, not a template, so a link hardcoded against a specific
-upstream ref goes stale on the next version bump. Write links into
-open-telemetry/semantic-conventions as `{{upstream_docs_base}}/docs/...`. The
-markdown templates substitute the placeholder with the ref pinned by the
-`model/manifest.yaml` dependency.
+Write links into open-telemetry/semantic-conventions as ordinary URLs pinned to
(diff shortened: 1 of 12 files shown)


trask:
no

trask:
Superseded by #435, same commit but from a fork branch.

PR #436, opened 2026-08-05
Pin the signal requirement level link to the upstream version

trask:
`signal_requirement_level.j2` hardcoded `blob/main` for the upstream signal-requirement-level link, so 39 links across `docs/gen-ai/*.md` pointed at `main` instead of the version pinned in `model/manifest.yaml`. It now uses the existing `upstream_docs_base` param, like the other templates that link upstream.


lmolkova (approved):
TY!

PR #439, opened 2026-08-07
Add budget governance attributes to invoke_agent span

### LINKED ISSUE #425: Budget governance attributes for invoke_agent span
## Problem

Agent frameworks expose per-invocation budget caps (token limits, iteration/turn limits) but there is no semantic convention for recording these in traces. Operators instrumenting agent loops today can see individual inference calls via `gen_ai.inference.client` spans but cannot answer "how close was this agent run to its configured budget?" or alert on agents approaching exhaustion.

This gap has real production impact. In openai/openai-agents-python#2848 an agent loop triggered thousands of unexpected API calls before detection; engineers confirmed that org-level monthly budgets are too coarse and post-run usage checks arrive too late. CrewAI users report similar runaway behavior with delegation ping-pong and tool loops (crewAIInc/crewAI#6414, #6219).

## What this enables

Today, an `invoke_agent` span shows you that an agent ran and how long it took. If it consumed too many tokens, you only find out when the monthly bill arrives or when you manually sum child spans after the fact (which breaks under sampling).

With these attributes, the parent `invoke_agent` span carries both the configured cap and actual consumption. An operator can write a single query:

```
gen_ai.agent.iteration_budget.consumed / gen_ai.agent.iteration_budget > 0.9
```

to find agents consistently running near their limit. An alert on `gen_ai.invoke_agent.token_budget.utilization > 0.95` catches runaway loops in seconds instead of hours. And because the values live on the parent span, they survive head-based sampling even when child inference spans are dropped.

## Proposal

Add four attributes to the `invoke_agent` span and one derived metric.

**Attributes**

`gen_ai.agent.token_budget` (int) records the cumulative token cap configured for the invocation.
`gen_ai.agent.token_budget.consumed` (int) records actual tokens consumed across all inference calls.
`gen_ai.agent.iteration_budget` (int) records the configured iteration/turn limit.
`gen_ai.agent.iteration_budget.consumed` (int) records completed iterations at invocation end.

**Metric**

`gen_ai.invoke_agent.token_budget.utilization` (histogram, unit 1) emits consumed/budget ratio per invocation, following the same pattern as `system.memory.utilization`.

## Framework mapping

Every major agent framework already has equivalent configuration:

| Framework | Token budget | Iteration budget |
|-----------|-------------|-----------------|
| OpenAI Agents SDK | n/a | `max_turns` |
| LangChain | n/a | `max_iterations` |
| LangGraph | n/a | `recursion_limit` |
| CrewAI | `max_tokens` (aggregate) | `max_iter` |
| Google ADK | n/a | `max_llm_calls` |

## Relationship to existing conventions

`gen_ai.request.max_tokens` is a per-call limit passed to the model API. The proposed attributes capture the invocation-level envelope that the framework enforces across multiple calls.

Complementary to #171 (agent finish_reason): that tells you WHY an agent stopped, this tells you HOW MUCH budget was allocated and consumed.

Instrumentations MUST NOT synthesize a token budget by multiplying iteration limits by per-call max_tokens.

elang2:
## Description

Add `gen_ai.agent.iteration_budget.limit` attribute for the `invoke_agent` span.
Resolves #425. Continuation of #426 (accidentally closed — I made my forked repo private which caused GitHub to close the PR and sever the fork relationship, making it non-reopenable).

New attribute on `invoke_agent` spans:

- `gen_ai.agent.iteration_budget.limit` (int) — configured iteration limit for a single agent's control loop.

## Changes since #426

- Made sub-agent token exclusion explicit in the attribute `brief` (not just the note) per reviewer feedback on summability
- Dropped `.consumed` attributes (derivable from same span tokens and the `inference_calls` metric)
- Dropped `token_budget.limit` and utilization metric (deferred to follow-up with per-invocation evidence)
- Tightened scope: excluded graph depth, run-scoped caps, and runner-wide inference counts
- Rebased onto current main

## Motivation

An SRE operating a multi-agent pipeline needs to know which agents are consistently running near their configured budget and alert before a runaway loop exhausts it. Today the `invoke_agent` span shows duration but not budget consumption. Operators only discover overruns when the monthly bill arrives or by manually summing child spans (which breaks under sampling since the `inference_calls` metric survives but child spans may not).

Agent frameworks that expose per-agent iteration caps:

| Framework | Iteration budget |
|-----------|-----------------|
| CrewAI | `max_iter` |
| Microsoft Agent Framework | `max_iterations` |
| AutoGen | `max_tool_iterations` |

Explicitly excluded (different semantic scope):

| Framework | Parameter | Why excluded |
|-----------|-----------|-------------|
| LangGraph | `recursion_limit` | Graph traversal depth, not loop iterations |
| OpenAI Agents SDK | `max_turns` | Run-scoped, persists across agent handoffs |
| Google ADK | `max_llm_calls` | Runner-wide inference count spanning multiple agents |

There is no vendor-neutral telemetry standard for recording per-agent iteration caps on traces. This attribute captures the invocation-level envelope that frameworks enforce.

## Prototype

Reference scenarios updated for AutoGen and CrewAI (`reference/scenarios/`). ADK scenario includes a capture-gap comment explaining why `max_llm_calls` is out of scope.

## Checklist

---

cc @meshailabs — this replaces #426 which was accidentally closed (I made my fork private which severed the fork relationship).

## Changes since last review

- Rebased on latest main (workflow.name noise and merge conflicts resolved)
- Removed `token_budget.limit` and `token_budget.utilization` (deferred to follow-up with Pydantic AI demonstration)
- Tightened `iteration_budget.limit` scope: excluded `recursion_limit`, `max_turns`, `max_llm_calls` via MUST NOT clause
- Added Microsoft Agent Framework `max_iterations` to examples
- Added CrewAI scenario emission, kept AutoGen scenario emission
- Added ADK capture-gap comment explaining scope mismatch
- Added SHOULD clause for per-framework documentation (matches `gen_ai.workflow.name` pattern)

### DIFF
--- a/changelog.d/439.enhancement.md
+++ b/changelog.d/439.enhancement.md
@@ -1,0 +1,1 @@
+Add `gen_ai.agent.iteration_budget.limit` attribute for agent loop-budget telemetry.
--- a/changelog.d/439.enhancement.md
+++ b/changelog.d/439.enhancement.md
@@ -1,1 +1,1 @@
-Add `gen_ai.agent.iteration_budget.limit` attribute for agent loop-budget telemetry.
+Add `gen_ai.agent.iteration_budget.limit` attribute for agent iteration budget telemetry.
--- a/model/gen-ai/registry.yaml
+++ b/model/gen-ai/registry.yaml
@@ -1025,3 +1025,38 @@ attributes:
       what `gen_ai.workflow.name` means in the context of that framework.
     examples: ["multi_agent_rag", "customer_support_pipeline"]
     stability: development
+  - key: gen_ai.agent.iteration_budget.limit
+    type: int
+    brief: Maximum iterations configured for this agent invocation.
+    note: |
+      The maximum number of iterations configured for this agent
+      invocation. An iteration is one pass through the agent's control
+      loop: a model request together with any resulting tool executions.
+
+      Instrumentations SHOULD record the framework's configured value
+      verbatim, without normalization. Exact bounds may differ by one
+      cycle across frameworks (for example, a tool-iteration cap may
+      permit one additional final model request).
+
+      Examples of framework parameters that map to this attribute:
+      CrewAI `max_iter`, Microsoft Agent Framework
+      `max_iterations`, AutoGen `max_tool_iterations`.
+
+      Semantic conventions for individual GenAI frameworks MUST
+      document what an iteration means in the context of that
+      framework.
+
+      Limits counting something other than control-loop cycles — graph
+      traversal depth (e.g. LangGraph `recursion_limit`), run-scoped
+      caps that persist across agent handoffs (e.g. OpenAI Agents SDK
+      `max_turns`), or runner-wide inference counts spanning multiple
+      agents (e.g. Google ADK `max_llm_calls`) — MUST NOT be recorded
+      in this attribute.
+
+      Instrumentations MUST NOT synthesize an iteration budget by
+      multiplying other limits.
+
+      If no iteration budget is configured, this attribute SHOULD NOT
+      be set.
+    examples: [10, 25, 50]
+    stability: development
--- a/model/gen-ai/registry.yaml
+++ b/model/gen-ai/registry.yaml
@@ -1042,7 +1042,7 @@ attributes:
       CrewAI `max_iter`, Microsoft Agent Framework
       `max_iterations`, AutoGen `max_tool_iterations`.
 
-      Semantic conventions for individual GenAI frameworks MUST
+      Semantic conventions for individual GenAI frameworks SHOULD
       document what an iteration means in the context of that
       framework.
 
--- a/model/gen-ai/spans.yaml
+++ b/model/gen-ai/spans.yaml
@@ -113,12 +113,20 @@ attribute_groups:
             library allows to set only a single model per agent. It SHOULD NOT
             be populated for agents that support multiple models or dynamic selection.
 
+  - id: attributes.gen_ai.agent.budget
+    visibility: internal
+    attributes:
+      - ref: gen_ai.agent.iteration_budget.limit
+        requirement_level:
+          recommended: When the framework exposes a configured iteration limit.
+
   - id: attributes.gen_ai.invoke_agent.common
     visibility: internal
     attributes:
       - ref_group: attributes.gen_ai.usage
       - ref_group: attributes.gen_ai.content
       - ref_group: attributes.gen_ai.error
+      - ref_group: attributes.gen_ai.agent.budget
       - ref_group: attributes.gen_ai.invoked_agent.internal.common
       - ref: gen_ai.operation.name
         requirement_level: required
--- a/reference/scenarios/autogen/scenario.py
+++ b/reference/scenarios/autogen/scenario.py
@@ -147,9 +147,9 @@ def run_agent_reference():
                     json.dumps([{"role": "user", "parts": [{"type": "text", "content": input_text}]}]),
                 )
                 agent_span.set_attribute("gen_ai.tool.definitions", json.dumps(tool_defs))
-                configured_iters = getattr(agent, "_max_tool_iterations", None)
-                if configured_iters is not None:
-                    agent_span.set_attribute("gen_ai.agent.iteration_budget.limit", configured_iters)
+                # In production, read the configured value from the framework's public API.
+                if max_tool_iterations is not None:
+                    agent_span.set_attribute("gen_ai.agent.iteration_budget.limit", max_tool_iterations)
                 # AutoGen delegates the LLM call to the underlying openai client,
                 # whose own instrumentation owns the inference span and the
                 # inference-operation-details event. This scenario emits only the
--- a/reference/scenarios/autogen/scenario.py
+++ b/reference/scenarios/autogen/scenario.py
@@ -109,13 +109,14 @@ def run_agent_reference():
     try:
         agent_name = "test_agent"
         agent_description = "Reference AutoGen assistant."
+        max_tool_iterations = 2
         agent = AssistantAgent(
             name=agent_name,
             model_client=model_client,
             description=agent_description,
             system_message=system_message,
             tools=[get_weather],
-            max_tool_iterations=2,
+            max_tool_iterations=max_tool_iterations,
         )
 
         async def _run():
@@ -146,6 +147,9 @@ def run_agent_reference():
                     json.dumps([{"role": "user", "parts": [{"type": "text", "content": input_text}]}]),
                 )
                 agent_span.set_attribute("gen_ai.tool.definitions", json.dumps(tool_defs))
+                configured_iters = getattr(agent, "_max_tool_iterations", None)
+                if configured_iters is not None:
+                    agent_span.set_attribute("gen_ai.agent.iteration_budget.limit", configured_iters)
                 # AutoGen delegates the LLM call to the underlying openai client,
                 # whose own instrumentation owns the inference span and the
                 # inference-operation-details event. This scenario emits only the
--- a/reference/scenarios/google-adk/scenario.py
+++ b/reference/scenarios/google-adk/scenario.py
@@ -189,6 +189,10 @@ def run_agent_reference():
         )
 
         session_service = InMemorySessionService()
+        # Capture gap: gen_ai.agent.iteration_budget.limit is not emitted here.
+        # ADK's max_llm_calls (RunConfig) counts every inference call across the
+        # entire runner including sub-agents, not per-agent loop iterations.
+        # That scope does not match iteration_budget.limit's definition.
         runner = Runner(agent=agent, app_name="test_app", session_service=session_service)
 
         async def _run():
--- a/docs/gen-ai/gen-ai-agent-spans.md
+++ b/docs/gen-ai/gen-ai-agent-spans.md
@@ -224,33 +224,34 @@ Semantic conventions for individual GenAI systems and frameworks MAY specify dif
 | [`gen_ai.request.choice.count`](/docs/registry/attributes/gen-ai.md) | ![Development](https://img.shields.io/badge/-development-blue) | `Conditionally Required` If available, in the request, and !=1. | int | The target number of candidate completions to return. | `3` |
 | [`gen_ai.request.seed`](/docs/registry/attributes/gen-ai.md) | ![Development](https://img.shields.io/badge/-development-blue) | `Conditionally Required` If applicable and if the request includes a seed. | int | Requests with same seed value more likely to return same result. | `100` |
 | [`server.port`](https://github.com/open-telemetry/semantic-conventions/blob/v1.44.0/docs/registry/attributes/server.md) | ![Stable](https://img.shields.io/badge/-stable-lightgreen) | `Conditionally Required` If `server.address` is set. | int | GenAI server port. [10] | `80`; `8080`; `443` |
+| [`gen_ai.agent.iteration_budget.limit`](/docs/registry/attributes/gen-ai.md) | ![Development](https://img.shields.io/badge/-development-blue) | `Recommended` [11] | int | Maximum iterations configured for this agent invocation. [12] | `10`; `25`; `50` |
 | [`gen_ai.request.frequency_penalty`](/docs/registry/attributes/gen-ai.md) | ![Development](https://img.shields.io/badge/-development-blue) | `Recommended` | double | The frequency penalty setting for the GenAI request. | `0.1` |
 | [`gen_ai.request.max_tokens`](/docs/registry/attributes/gen-ai.md) | ![Development](https://img.shields.io/badge/-development-blue) | `Recommended` | int | The maximum number of tokens the model generates for a request. | `100` |
-| [`gen_ai.request.model`](/docs/registry/attributes/gen-ai.md) | ![Development](https://img.shields.io/badge/-development-blue) | `Recommended` If applicable. | string | The name of the GenAI model configured for the agent. [11] | `gpt-4` |
+| [`gen_ai.request.model`](/docs/registry/attributes/gen-ai.md) | ![Development](https://img.shields.io/badge/-development-blue) | `Recommended` If applicable. | string | The name of the GenAI model configured for the agent. [13] | `gpt-4` |
 | [`gen_ai.request.presence_penalty`](/docs/registry/attributes/gen-ai.md) | ![Development](https://img.shields.io/badge/-development-blue) | `Recommended` | double | The presence penalty setting for the GenAI request. | `0.1` |
-| [`gen_ai.request.previous_response.id`](/docs/registry/attributes/gen-ai.md) | ![Development](https://img.shields.io/badge/-development-blue) | `Recommended` [12] | string | The unique identifier of a previous response or interaction used to provide context for the current operation. [13] | `resp_0123456789aBCdef`; `interaction-123` |
+| [`gen_ai.request.previous_response.id`](/docs/registry/attributes/gen-ai.md) | ![Development](https://img.shields.io/badge/-development-blue) | `Recommended` [14] | string | The unique identifier of a previous response or interaction used to provide context for the current operation. [15] | `resp_0123456789aBCdef`; `interaction-123` |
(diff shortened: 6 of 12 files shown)


model/gen-ai/registry.yaml
```diff
@@ -944,3 +944,84 @@ attributes:
...
+    brief: Cumulative token budget configured for this agent invocation.
+    note: |
+      The cumulative token budget (input + output, across all inference
+      calls) configured for this agent invocation. When the budget is
+      exhausted the agent loop terminates.
+
+      Instrumentations SHOULD populate this attribute only when the
+      configured limit acts as a cumulative cap across the full
+      invocation, not when it constrains a single inference call
+      (use `gen_ai.request.max_tokens` for per-call limits).
```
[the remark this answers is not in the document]
elang2: The note scopes it to "configured for this agent invocation." No referenced framework supports mid-flight budget changes, so a last-applied-value sentence would be speculative. Happy to add one if a reviewer feels it's ambiguous.

model/gen-ai/spans.yaml
```diff
@@ -77,12 +77,29 @@ attribute_groups:
             library allows to set only a single model per agent. It SHOULD NOT
             be populated for agents that support multiple models or dynamic selection.
 
+  - id: attributes.gen_ai.agent.budget
+    visibility: internal
+    attributes:
+      - ref: gen_ai.agent.token_budget.limit
+        requirement_level:
```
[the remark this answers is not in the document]
elang2: AutoGen scenario emits iteration_budget.limit, .consumed, and token_budget.consumed (see autogen/scenario.py in this diff). token_budget.limit is intentionally unset because neither framework exposes a cumulative token cap. There's an explicit capture-gap comment in the ADK scenario explaining this.

model/gen-ai/metrics.yaml
```diff
@@ -229,6 +229,31 @@ metrics:
     attributes:
       - ref: gen_ai.agent.name
         requirement_level: recommended
+  - name: gen_ai.invoke_agent.token_budget.utilization
+    requirement_level: recommended
+    annotations:
+      code_generation:
+        metric_value_type: double
```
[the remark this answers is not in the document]
elang2: Covered in google-adk/data.json and scenario.py in this PR.

changelog.d/439.enhancement.md
```diff
@@ -0,0 +1 @@
+Add `gen_ai.agent.token_budget.limit`, `gen_ai.agent.token_budget.consumed`, `gen_ai.agent.iteration_budget.limit`, and `gen_ai.agent.iteration_budget.consumed` attributes and `gen_ai.invoke_agent.token_budget.utilization` metric for agent budget governance telemetry.
```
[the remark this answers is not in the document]
elang2: File is named 439.enhancement.md. The #426 in the description refers to the predecessor PR, not the fragment.

reference/scenarios/autogen/scenario.py
```diff
@@ -146,6 +146,7 @@ async def _run():
                     json.dumps([{"role": "user", "parts": [{"type": "text", "content": input_text}]}]),
                 )
                 agent_span.set_attribute("gen_ai.tool.definitions", json.dumps(tool_defs))
+                agent_span.set_attribute("gen_ai.agent.iteration_budget.limit", 2)
```
lmolkova: this seems to be a magic number. Please demonstrate where this actually can come from in this and other instrumentations.

reference/scenarios/autogen/scenario.py
```diff
@@ -183,6 +184,14 @@ async def _capture_create(messages, **kwargs):
                     agent_span.set_attribute("gen_ai.usage.input_tokens", total_input_tokens)
                 if total_output_tokens:
                     agent_span.set_attribute("gen_ai.usage.output_tokens", total_output_tokens)
+                agent_span.set_attribute("gen_ai.agent.iteration_budget.consumed", len(captured_results))
+                agent_span.set_attribute("gen_ai.agent.token_budget.consumed", total_input_tokens + total_output_tokens)
```
lmolkova: is there a point in capturing a derived valus since it's already exposed via individual attributes?

reference/scenarios/autogen/scenario.py
```diff
@@ -183,6 +184,14 @@ async def _capture_create(messages, **kwargs):
                     agent_span.set_attribute("gen_ai.usage.input_tokens", total_input_tokens)
                 if total_output_tokens:
                     agent_span.set_attribute("gen_ai.usage.output_tokens", total_output_tokens)
+                agent_span.set_attribute("gen_ai.agent.iteration_budget.consumed", len(captured_results))
```
lmolkova: this looks confusing - why number of return results has anything to do with budget?

reference/scenarios/google-adk/scenario.py
```diff
@@ -260,16 +271,19 @@ async def _run():
...
-                            candidate_token_count = usage_metadata.get("candidates_token_count")
-                        if prompt_token_count is not None:
-                            agent_span.set_attribute("gen_ai.usage.input_tokens", prompt_token_count)
-                        if candidate_token_count is not None:
-                            agent_span.set_attribute("gen_ai.usage.output_tokens", candidate_token_count)
+                    if total_input_tokens > 0:
+                        agent_span.set_attribute("gen_ai.usage.input_tokens", total_input_tokens)
+                    if total_output_tokens > 0:
+                        agent_span.set_attribute("gen_ai.usage.output_tokens", total_output_tokens)
+                    agent_span.set_attribute("gen_ai.agent.iteration_budget.consumed", call_counts["inference"])
```
lmolkova: there is a metric [gen_ai.invoke_agent.inference_calls](https://github.com/open-telemetry/semantic-conventions-genai/blob/main/docs/gen-ai/gen-ai-metrics.md#metric-gen_aiinvoke_agentinference_calls) - would it make sense to just define an attribute with the same name? Or, even better, consumers can count number of inference spans under this agent and then no attribute on a span is needed.

lmolkova (changes requested):

model/gen-ai/spans.yaml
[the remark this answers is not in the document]
elang2: Correction: AutoGen does have a token cap (`TokenUsageTermination`), it's just team level. Updated in a53f0eb.

reference/scenarios/autogen/scenario.py
elang2: Fixed in a53f0eb. The attribute pulls straight from what you pass to the constructor (`AssistantAgent(max_tool_iterations=N)` for AutoGen, `RunConfig(max_llm_calls=N)` for ADK). No magic numbers.

reference/scenarios/autogen/scenario.py
elang2: Agreed, gone. You can already sum it from the same span.

reference/scenarios/autogen/scenario.py
elang2: You're right. Bad proxy on my part. Line's gone entirely now that `iteration_budget.consumed` is dropped.

reference/scenarios/google-adk/scenario.py
elang2: Agreed. `inference_calls` covers consumption and survives sampling, so both `.consumed` attributes are gone.

The configured cap is a different story though. That only exists in framework config (`max_tool_iterations`, `max_llm_calls`), nowhere in spans or metrics. So the `.limit` pair stays.

On the token side, AutoGen does have a cap via `TokenUsageTermination`, but it's team level so the single agent scenario can't emit `token_budget.limit`. Note added. If you'd rather see it demonstrated before it lands, happy to split the token budget stuff into a follow up. The iteration budget half stands on its own.

elang2: Agreed. `inference_calls` covers consumption and survives sampling, so both `.consumed` attributes are gone.

The configured cap is a different story though. That only exists in framework config (`max_tool_iterations`, `max_llm_calls`), nowhere in spans or metrics. So the `.limit` pair stays.

On the token side, AutoGen does have a cap via `TokenUsageTermination`, but it's team level so the single agent scenario can't emit `token_budget.limit`. Note added. If you'd rather see it demonstrated before it lands, happy to split the token budget stuff into a follow up. The iteration budget half stands on its own.

elang2:
Pushed a53f0eb. Both `.consumed` attributes are gone (you can derive them from same span tokens and `inference_calls`). The `.limit` pair stays. `iteration_budget.limit` is demonstrated from framework config in both scenarios. `token_budget.limit` has a capture gap note since the scenario can't emit it yet. Happy to split that into a follow up if you want it demonstrated first.

Motivation table, changelog, and generated docs updated.

elang2:
Pushed a53f0eb. Both `.consumed` attributes are gone (you can derive them from same span tokens and `inference_calls`). The `.limit` pair stays. `iteration_budget.limit` is demonstrated from framework config in both scenarios. `token_budget.limit` has a capture gap note since the scenario can't emit it yet. Happy to split that into a follow up if you want it demonstrated first.

Motivation table, changelog, and generated docs updated.

model/gen-ai/registry.yaml
```diff
@@ -935,19 +935,56 @@ attributes:
     type: string
     brief: Human-readable name of the GenAI workflow provided by the application.
     note: |
-      The workflow name is usually a static, application-unique identifier defined
-      in a framework-specific way.
+      This attribute can be populated in different frameworks; for example, as the name of the first chain in LangChain or the name of the crew in CrewAI.
```
lmolkova: this does not seem to be related

model/gen-ai/registry.yaml
```diff
@@ -935,19 +935,56 @@ attributes:
...
-      the name of the crew in CrewAI, or the entry point agent in ADK or
-      OpenAI Agents when no explicit workflow name is provided.
+      `gen_ai.workflow.name` MUST have low cardinality.
+      Semantic conventions for individual Generative AI frameworks SHOULD document what `gen_ai.workflow.name` means in the context of that framework.
+      If there is no low-cardinality workflow name available for a given framework, this attribute MUST NOT be captured by default.
+    examples: ["multi_agent_rag", "customer_support_pipeline"]
+    stability: development
+  - key: gen_ai.agent.token_budget.limit
+    type: int
+    brief: Cumulative token budget configured for this agent invocation.
```
lmolkova: there is no framework that has such a configuration option, at least none were demonstrated in this PR

model/gen-ai/registry.yaml
```diff
@@ -935,19 +935,56 @@ attributes:
...
+      Instrumentations SHOULD populate this attribute only when the
+      configured limit acts as a cumulative cap across the full
+      invocation, not when it constrains a single inference call
+      (use `gen_ai.request.max_tokens` for per-call limits).
 
-      Semantic conventions for individual Generative AI frameworks SHOULD document
-      what `gen_ai.workflow.name` means in the context of that framework.
-    examples: ["multi_agent_rag", "customer_support_pipeline"]
+      Examples of framework parameters that map to this attribute:
+      AutoGen `TokenUsageTermination`.
```
lmolkova: is this framework deprecated / replaced?

model/gen-ai/registry.yaml
```diff
@@ -935,19 +935,56 @@ attributes:
...
+    note: |
+      The maximum number of iterations configured for this agent
+      invocation. The definition of "iteration" is framework-specific;
+      instrumentations SHOULD record whatever value the framework uses
+      as its primary loop-termination counter without conversion.
+
+      Examples of framework parameters that map to this attribute:
+      LangChain `max_iterations`, CrewAI `max_iter`, OpenAI Agents SDK
+      `max_turns`, LangGraph `recursion_limit`, Google ADK
+      `max_llm_calls`, AutoGen `max_tool_iterations`.
```
lmolkova: these things seem to be way too different - number of llm calls != recursion limit != tool iterations.

An attribute that attempts to represent them all is not usable in practice without specific library context.  The iteration needs to be defined as something common, or, if there is no such generic concept, the limits could be recorded with framework specific attributes

lmolkova (changes requested):

elang2:
The workflow.name diff is rebase noise from a stale base. Gone now.

I pulled `token_budget.limit` into a follow-up. You're right that no scenario here demonstrates it at the agent-invocation level (AutoGen's `TokenUsageTermination` is team-level). Pydantic AI's `UsageLimits(total_tokens_limit=…)` is the one I'll use for that PR. The utilization metric I mentioned Aug 14 isn't in the ADK scenario either, so I dropped that claim.

AutoGen entered maintenance mode late 2025, Microsoft Agent Framework is its successor. MAF's `FunctionInvocationConfiguration.max_iterations` bounds the same per-invocation loop, so I added it to the examples. Kept the AutoGen scenario since this repo already maintains it.

Heterogeneity concern is fair. `recursion_limit` (graph depth), `max_llm_calls` (runner-wide), `max_turns` (run-scoped across handoffs) don't map to "one pass through the agent's decision loop." All three are excluded in the note, ADK scenario doesn't emit it. Run-scoped caps could become a workflow-level attribute later but out of scope here.

What's left is CrewAI `max_iter`, MAF `max_iterations`, and AutoGen `max_tool_iterations`. They can differ by one cycle depending on where the framework exits its loop, which is why the note says "record verbatim, without normalization" rather than defining a canonical count. The distinction I care about is bounded vs unbounded. Each of these ticks at most once per model request within its scope. `recursion_limit` doesn't have that property (a superstep can contain zero or many model requests). That's the line.

Why unified rather than framework-specific: an ops team running CrewAI in one service and Agent Framework in another can't write one alert for "agent approaching ceiling" without enumerating every possible attribute name. `gen_ai.workflow.name` merged with the same kind of latitude ("first chain in LangChain, crew in CrewAI, entry-point agent in ADK") plus the per-framework documentation clause.

Diff reflects all of this. If you'd still rather see framework-specific attributes I can go that route.

elang2:
Empirical update on the counting-method question. I modeled eleven frameworks from source code, then executed eight of them against a deterministic mock LLM with a request ledger. The mock gives ground truth (identical work across all frameworks, no network variance, fully reproducible). The source-code model was right for exactly half.

The executed table, for identical ground truth of 4 LLM calls and 3 tool calls:

| Framework | Version | Counting unit | consumed |
|-----------|---------|--------------|----------|
| AutoGen | 0.4.7 | composite_messages | 5 |
| OpenAI Agents | 0.22.0 | llm_invocations | 4 |
| LangChain | 0.3.14 | tool_cycles | 3 |
| LangGraph | 1.2.11 | graph_nodes | 8 |
| Semantic Kernel | 1.44.1 | auto_invoke_rounds | 3 |
| CrewAI | 1.15.16 | tool_cycles | 3 |
| LlamaIndex | 0.14.24 | llm_invocations | 4 |
| Agno | 1.2.5 | NOT ENFORCED | n/a |

Every number is machine-derived from results/S2-executed.json in the linked repo. The `consumed` column is the counter value for the full ground-truth workload (counter-at-budget-stop is a second field in the artifact for anyone diffing against the JSON). The artifact includes mock-ledger cross-checks (actual_llm_calls == mock_confirmed_calls on all rows).

I expected a majority cluster in the unit taxonomy. There isn't one. Across enforcing frameworks the observed units break down as composite_messages (1), llm_invocations (2), tool_cycles (2), graph_nodes (1), auto_invoke_rounds (1). Five distinct counting methods, no winner. Without a unit attribute on the telemetry, none of these numbers are comparable to each other. A consumed>3 alert fires for 4 of the 7 enforcing frameworks but not the other 3, for the exact same work.

This supports "record verbatim, without normalization" in the current PR. If a consumed attribute comes back later it needs a mandatory counting_method enum. The measured chaos is the argument for the attribute.

Side finding on Agno 1.2.5: tool_call_limit is a real propagated dataclass field and the enforcement code runs, but it doesn't actually stop the agent. Nine tool calls past a limit of three. Known bug (agno-agi/agno#8304), independently reproduced here blind before we found the existing reports. Deterministic mock removes model behavior as a variable.

Harness is public at https://github.com/elang2/agent-budget-semantics (v0.3.0, pip-installable, runs without API keys). Happy to add any framework or scenario that would help here.

lmolkova:
thanks for the research. what I believe it proves is that a single attribute won't be able to describe it in any reasonable way. With or without unit these numbers are not comparable. It can only be captured in framework-specific way like openai.agent.max_invocations

elang2:
> thanks for the research. what I believe it proves is that a single attribute won't be able to describe it in any reasonable way. With or without unit these numbers are not comparable. It can only be captured in framework-specific way like openai.agent.max_invocations

You are right! A shared attribute can't carry the weight operator the operator use case wants from it. Closing this.  Thanks for engaging !

PR #440, opened 2026-08-08
Add span/event attributes for usage breakdown per modality / cache / phases

lmolkova:
Extracts non-controversial piece from https://github.com/open-telemetry/semantic-conventions-genai/pull/374 and https://github.com/open-telemetry/semantic-conventions-genai/pull/197

Defined attributes to stamp on spans and events

- `gen_ai.usage.cache_write.input_tokens` - The number of input tokens written to a provider-managed cache.
- `gen_ai.usage.text.input_tokens` - The number of text input tokens.
- `gen_ai.usage.image.input_tokens` - The number of image input tokens.
- `gen_ai.usage.audio.input_tokens` - The number of audio input tokens.
- `gen_ai.usage.text.output_tokens` - The number of text output tokens.
- `gen_ai.usage.image.output_tokens` - The number of image output tokens.
- `gen_ai.usage.audio.output_tokens` - The number of audio output tokens.
- `gen_ai.usage.text.cache_read.input_tokens` - The number of text input tokens served from a provider-managed cache.
- `gen_ai.usage.image.cache_read.input_tokens` - The number of image input tokens served from a provider-managed cache.
- `gen_ai.usage.audio.cache_read.input_tokens` - The number of audio input tokens served from a provider-managed cache.

Metrics will be addressed in #374 
No tools, video, or document token usage for now due to limited support across providers - intentionally skipping and waiting if anyone complains. 
Also added image / audio reference scenarios.

## Motivation

Provide token usage breakdown per inference operation on spans and events - gives almost precise cost for individual operations, can be used for rough cost estimates across the system.

## Checklist

- [ ] Towncrier fragment added under `changelog.d/` for any change to the conventions that a consumer would care about. Editorial changes (typos, pure rewording, repo tooling) don't need an entry.

See [CONTRIBUTING.md]


model/gen-ai/spans.yaml
```diff
@@ -42,10 +42,30 @@ attribute_groups:
         requirement_level: recommended
       - ref: gen_ai.usage.cache_read.input_tokens
         requirement_level: recommended
-      - ref: gen_ai.usage.cache_creation.input_tokens
+      - ref: gen_ai.usage.cache_write.input_tokens
```
aabmass, line 46: Dumb question but is it OK that these are all `Recommended`, what if a provider doesn't support any of these specific breakdowns?

reference/reports/invoke-agent-client-span.md
aabmass, line 1: Where did we settle on including these on the agent spans and overcounting issues when doing rollups?

model/gen-ai/registry.yaml
```diff
@@ -308,6 +336,48 @@ attributes:
...
+    brief: The number of audio output tokens.
+    note: >
+      The value SHOULD be included in `gen_ai.usage.output_tokens`.
+    examples: [240]
+    stability: development
+  - key: gen_ai.usage.text.cache_read.input_tokens
+    type: int
+    brief: The number of text input tokens served from a provider-managed cache.
+    note: >
+      The value SHOULD be included in `gen_ai.usage.cache_read.input_tokens`.
```
aabmass: Should the note on `gen_ai.usage.<modality>.cache_read.input_tokens` also mention that it's included in `gen_ai.usage.<modality>.input_tokens`?

Right now it only mentions `gen_ai.usage.cache_read.input_tokens`. For reference, that's what we do in:
- `reference/src/semconv_genai/mock_server/google_genai.py:155-158` (cached tokens are modeled as a subset of prompt tokens)
- `reference/scenarios/google-genai/scenario.py:88-101` (mapping `prompt_tokens_details` to `<modality>.input_tokens` and `cache_tokens_details` to `<modality>.cache_read.input_tokens`)
- #374 ("Each detailed attribute is a subset of its aggregate")

```yaml
    note: >
      The value SHOULD be included in `gen_ai.usage.cache_read.input_tokens` and in `gen_ai.usage.<modality>.input_tokens`.
```

aabmass (approved):

reference/reports/invoke-agent-client-span.md
lmolkova: good point. I removed new attributes from internal span, but kept on the client spans.

I don't believe we landed anywhere on https://github.com/open-telemetry/semantic-conventions-genai/issues/19

so just not making it worse for now by not including breakdowns on internal agents

model/gen-ai/spans.yaml
lmolkova: right, updated to `recommended: when available`

model/gen-ai/registry.yaml
```diff
@@ -284,13 +284,41 @@ attributes:
...
+  - key: gen_ai.usage.cache_write.input_tokens
     type: int
     brief: The number of input tokens written to a provider-managed cache.
     note: >
       The value SHOULD be included in `gen_ai.usage.input_tokens`.
     examples: [25]
     stability: development
+  - key: gen_ai.usage.tool.input_tokens
+    type: int
+    brief: The number of input tokens used for tool definitions and tool call results.
```
alexmojaki: This seems to be only for google's tool_use_prompt_token_count. that field is not about tool definitions. it's also not about all tool calls, just ones that happen server side.

i think that having this as a provider-agnostic field might not be a good idea, if other providers grow a similar sounding field it's likely to mean something slightly different.

model/gen-ai/spans.yaml
```diff
@@ -41,12 +41,51 @@ attribute_groups:
...
+          recommended: When applicable.
+      - ref: gen_ai.usage.cache_write.input_tokens
+        requirement_level:
+          recommended: When applicable.
       - ref: gen_ai.usage.output_tokens
         requirement_level: recommended
 
+  # Kept out of `attributes.gen_ai.usage`: an internal agent span reports the
+  # usage its own call tree accumulated, and summing breakdowns across that
+  # tree would count the same tokens more than once.
```
alexmojaki, line 54: i don't understand how this reasoning doesn't apply to the other token usage attributes, especially cache and reasoning tokens.

model/gen-ai/registry.yaml
```diff
@@ -308,6 +336,51 @@ attributes:
...
+    note: >
+      The value SHOULD be included in `gen_ai.usage.output_tokens`.
+    examples: [240]
+    stability: development
+  - key: gen_ai.usage.text.cache_read.input_tokens
+    type: int
+    brief: The number of text input tokens served from a provider-managed cache.
+    note: >
+      The value SHOULD be included in `gen_ai.usage.cache_read.input_tokens`
+      and in `gen_ai.usage.text.input_tokens`.
```
alexmojaki, line 366: i think an example that explicitly clarifies these intersection cases is important, including how to translate from a raw API response to attributes, especially when the translation is nontrivial

model/gen-ai/spans.yaml
lmolkova: it absolutely applies, but taking baby steps. Will follow up via https://github.com/open-telemetry/semantic-conventions-genai/issues/460

model/gen-ai/registry.yaml
lmolkova: agreed, removed for now - can get back to it if anyone asks for it to be added

alexmojaki (approved):

PR #441, opened 2026-08-08
Run the reference scenarios under the conformance runner

lmolkova:
Moves the reference scenarios onto the runner, mock LLM server and GenAI advice
policies from open-telemetry/semantic-conventions-conformance#12, dropping this
repo's copies. Nothing is vendored: the checkout is fetched at the
`CONFORMANCE_REF` pinned in `versions.env`.

**Demo until that PR merges** — the pin is a branch SHA.

Each `scenarios/<library>/` gains a `conformance.yaml` saying how to run the
scenario. Runs validate against the working tree's `model/`, not the registry
that package pins. `run-scenario` keeps its CLI; violations are reported, not
failed. 24 of 28 scenarios report violations — mostly unexpected span kinds,
missing `server.address`, and content-schema mismatches. Declaring them under
`expected_violations` is a follow-up.

Dropped: the mock server, the weaver pipeline, the result parser and the
hand-written YAML model walker (~1900 lines).

`data.json` is now keyed by registry span type (`gen_ai.inference.client`), so
those diffs are renames. The only content change:

- google-adk: `gen_ai.execute_tool.duration` and `gen_ai.invoke_agent.duration`
  now recorded — ADK emits them natively, and the reduction covers every metric
  the registry declares rather than a hand-kept list of four.

Also:

- `reference/coverage-model.json` is committed — weaver's resolution of
  `model/`, generated by `make generate-all`. The reports read it, so they and
  `data.json` share one view of what each signal declares.
- google-genai scenario updated for the mock's Interactions usage fields.
- The CodeQL config existed only to exclude the mock server; removed.


.github/codeql/codeql-config.yml
lmolkova, line 1: was needed for mock server, not needed since mock is in conformance repo

lmolkova:
Reopening from a fork branch.

PR #442, opened 2026-08-08
Run the reference scenarios under the conformance runner

lmolkova:
Moves the reference scenarios onto the runner, mock LLM server and GenAI advice
policies from open-telemetry/semantic-conventions-conformance#12, dropping this
repo's copies. Nothing is vendored: the checkout is fetched at the
`CONFORMANCE_REF` pinned in `versions.env`.

**Demo until that PR merges** — the pin is a branch SHA.

Each `scenarios/<library>/` gains a `conformance.yaml` saying how to run the
scenario. Runs validate against the working tree's `model/`, not the registry
that package pins. `run-scenario` keeps its CLI; violations are reported, not
failed. 24 of 28 scenarios report violations — mostly unexpected span kinds,
missing `server.address`, and content-schema mismatches. Declaring them under
`expected_violations` is a follow-up.

Dropped: the mock server, the weaver pipeline, the result parser and the
hand-written YAML model walker (~1900 lines).

`data.json` is now keyed by registry span type (`gen_ai.inference.client`), so
those diffs are renames. The only content change:

- google-adk: `gen_ai.execute_tool.duration` and `gen_ai.invoke_agent.duration`
  now recorded — ADK emits them natively, and the reduction covers every metric
  the registry declares rather than a hand-kept list of four.

Also:

- The reports are now built from weaver's resolution of `model/` rather than a
  hand-written YAML walker, so they and `data.json` share one view of what each
  signal declares. It is resolved on demand into `reference/.cache/`.
- google-genai scenario updated for the mock's Interactions usage fields.
- The CodeQL config existed only to exclude the mock server; removed.


versions.env
```diff
@@ -4,3 +4,7 @@ WEAVER_VERSION=v0.25.1
 # renovate: datasource=git-refs depName=open-telemetry/opentelemetry-weaver-packages versioning=git ref=main
 POLICY_REPO_URL=https://github.com/open-telemetry/opentelemetry-weaver-packages.git
 POLICY_REPO_REF=d84341cf20a1fef1a833ef44d318c41a770e6e64
+
+# renovate: datasource=git-refs depName=open-telemetry/semantic-conventions-conformance versioning=git ref=main
```
trask, line 8: This `# renovate:` annotation won't actually be picked up. The custom manager in `.github/renovate.json5` lists explicit `matchStrings` for `WEAVER_VERSION=` and for the `POLICY_REPO_URL=`/`POLICY_REPO_REF=` pair, and Renovate's regex manager has no generic fallback for a bare `# renovate:` comment. As written, `CONFORMANCE_REF` will silently never be bumped even though the annotation says it is managed. Add a matching `matchStrings` entry alongside the `POLICY_REPO_*` one, or drop the annotation until the pin is wired up.

reference/src/semconv_genai/conformance.py
```diff
@@ -0,0 +1,151 @@
...
+
+
+def _environment() -> dict[str, str]:
+    """A clean environment with the pinned weaver ahead of anything on PATH."""
+    env = {k: v for k, v in os.environ.items() if k not in ("VIRTUAL_ENV", "PYTHONHOME", "PYTHONPATH")}
+    env["PATH"] = os.pathsep.join([str(ensure_weaver().parent), env.get("PATH", "")])
+    return env
+
+
+def _is_stale(model: Path) -> bool:
```
trask, line 110: The cached coverage model is invalidated only when `model/` changes, but its content also depends on `CONFORMANCE_REF` (which owns the weaver templates that produce it) and on `WEAVER_VERSION`. A warm `reference/.cache/coverage-model.json` short-circuits before `checkout()` and `ensure_weaver()` ever run, so after a pin bump `make generate-all` keeps building the reports from the previous resolution until someone touches `model/` or runs `make clean`. Consider folding both pins into the staleness check -- treating `versions.env` as another source, or keying the cached file on the ref.

trask (approved):
Lightly filtered AI-generated feedback

Issue #373, opened 2026-07-11
Proposal: tool risk attributes for execute_tool and MCP tool call telemetry


JM-Lab:
### Area(s)

area:mcp, area:tools

### What's missing?

The conventions describe what a tool call did (method, tool name, arguments,
result, duration, error), but not what the host assessed the call could do.
Hosts that gate tool execution increasingly compute a risk posture per tool
before exposing or invoking it: from enforced sandbox capabilities for local
tools, or from declared annotations (readOnlyHint, destructiveHint,
idempotentHint, openWorldHint), side-effect scope, and provenance for MCP
tools. That assessment drives real runtime decisions (exposure ceilings,
human-approval requirements), yet there is no conventional attribute to carry
it on the resulting spans and metrics.

Without it, operators cannot answer basic safety questions from telemetry
alone: how many high-risk tool calls ran today, did a floor override fire,
which sessions invoked tools above a given level, or how approval rates
correlate with assessed risk (the risk-approval correlation was also raised
in #95). Each producer invents a vendor-prefixed attribute, which defeats
cross-vendor dashboards for exactly the signals security teams ask about.

### Describe the solution you'd like

A minimal, payload-free set of three attributes, applicable to execute_tool
spans and MCP tools/call client/server spans, and usable as metric
dimensions:

- gen_ai.tool.risk.level (string enum, ordered): the producer-assessed risk
  band for this tool invocation at call time, reflecting the host's
  assessment under the context and band mapping in force at emission time.
  The same tool may carry different levels across invocations as sandbox
  posture, provenance, or exposure context change. Suggested well-known
  values, lowest to highest: minimal, low, moderate, high, critical.
  Producers with numeric or house scales document a mapping onto the shared
  band; recorded levels are not reinterpreted under a later mapping (the
  stability concern discussed in #368).
- gen_ai.tool.risk.override (string, opt-in, low cardinality): the rule id
  that pinned the level when a floor/override short-circuited the normal
  scoring. Absent when no override fired. A single string attribute: a small
  set of documented well-known identifiers where there is broad
  cross-producer agreement, while producer-specific rules use opaque
  producer-defined identifiers rather than descriptive rule names that would
  expose policy structure. Producer-defined values SHOULD come from a
  stable, bounded vocabulary and MUST NOT encode per-request, per-user,
  tenant-specific, or otherwise unbounded policy detail. Initial well-known
  candidates, one from each floor-rule family observed in practice (action
[...]

PR #443, opened 2026-08-09
Add per-operation cost conventions (gen_ai.usage.cost.*)

### LINKED ISSUE #287: Add convention for operation costs
### Propose new conventions

Cost tracking seems to be an important use case in observability, especially for GenAI / LLMs where costs can be unpredictable (depend on input and output token counts).

But cost tracking in general is something that's always been useful, it's very common for calling external APIs to have a cost associated with it.

I'd like to propose we add an `operation.cost: int` and `operation.cost_units: str` attributes (open to suggestions on naming, units, etc.) to generalize this.

With these attributes in place it would be easy to make generic queries such as:
- Most expensive operations
- Total cost of LLM calls
- Spend per service

elang2:
## Summary

Adds per-operation cost attributes and a metric, implementing the
consensus from #287.

There's no standard OTel convention for reporting GenAI cost today; each
vendor is rolling their own (LiteLLM, Pydantic AI + Logfire, OpenRouter —
different shapes for the same USD amount). This PR proposes the shared
schema.

## What's included

Three span attributes on the `attributes.gen_ai.usage.cost` group, wired
to **inference-only** spans (moved out of agent/workflow spans after
review feedback, to avoid double-counting):

- `gen_ai.usage.cost.amount` (double) — the monetary cost computed at
  recording time. When a cost value is present in the response the client
  received (from the provider or an intermediary that priced the request,
  e.g. OpenRouter), it is passed through unmodified. Otherwise the
  emitting component computes it from token counts and pricing data it
  owns.
- `gen_ai.usage.cost.currency` (string) — the ISO 4217 currency code.
  Conditionally required when `cost.amount` is set.
- `gen_ai.usage.cost.source` (enum: `provider | local`) — provenance on a
  receive-vs-derive test: `provider` when the value was present in the
  response received (whoever answered — from the client's perspective a
  gateway is the provider), `local` when the recorder computed it.
  Backend enrichment (post-hoc reconstruction against pricing that may
  have drifted since the call) is out of scope.

One histogram metric `gen_ai.client.operation.cost`, in `{cost}`, with
`ExplicitBucketBoundaries` advice on a roughly US-dollar scale.

### Who is expected to record cost

This is what @lmolkova's review asked for and the proposal did not answer
anywhere. Now normative text on `cost.amount`:

> Cost is therefore recorded by whichever component can determine it. Any
> instrumentation of the client can record it when the value is present in
> the response, since doing so copies a field the client already received.
> When it is absent, only a component configured with pricing data can
> record it. A component with neither is **not expected to acquire pricing
> data in order to record cost**.

So a plain auto-instrumentation is covered for the pass-through case and is
never asked to price anything.

## Design decisions

- **Histogram over counter.** Gives p50/p99 per call and sum from one
  instrument.
- **Floating point for `.amount`.** This is observability, not accounting.
- **Frozen at recording time.** The recorded amount reflects the pricing
  data in force when it was computed; consumers never recompute against
  later pricing.
- **`.amount` suffix** avoids the namespace-prefix collision the registry
  policy forbids. Matches the `gen_ai.evaluation.score.value`/`.label`
  precedent.
- **Effective price** for the recorded amount (what was actually charged
  after discounts). Keeps a FOCUS-aligned bridge open for the future
  without overcomplicating v0.1.
- **Enum kept at two values** (`provider | local`) after review. The
  finer question of *which* pricing data and *what* version is not
  answered here; I owe @meshailabs an issue for it and have not filed it.

- **Own-operation rule.** A record covers only the operation being
  recorded and MUST NOT include cost attributable to nested operations.
  This does not make cost summable across a trace — de-duplicating
  overlapping records is the consumer's job. Requested by @meshailabs.

## Follow-ups, not in this PR

An earlier revision carried four normative changes nobody asked for;
738270b removes them per CONTRIBUTING.md. The argument for them stays in
80325bb's message.

- Requirement level on `cost.amount` (`conditionally_required` vs
  `recommended`), with the instruction the narrow condition needs.
- Restricting the metric to operations that perform model inference, and
  `cost.source` becoming `required` on it.
- The metric admits any operation while the attributes live only on
  `gen_ai.inference.client`, so `embeddings` has nowhere to record the same
  number on the span.

## What this leaves out (intentionally)

- **Cost allocation** (which team/budget-owner gets charged). Bigger
  design surface; better as a follow-up.
- **FOCUS integration** at the collector level. Orthogonal to the
  semconv.
- **Backend enrichment** as a source. Post-hoc reconstruction has
  different reliability characteristics from request-time computation and
  is out of scope for these attributes entirely.
- **Per-token-class cost breakdown** (input/output/cache-read/cache-write/
  reasoning/tool). Follow-up issue, driven by evidence from LiteLLM
  #36875 shipping this and Langfuse #13807 documenting cache-token
  accounting drift. This is also the answer to @alexmojaki's non-token
  charges question: a `provider` amount may already bundle them where a
  `local` one computed from token counts cannot, and `cost.source` is the
  only signal of which you hold.

## Reference emission

The LiteLLM reference scenario in `reference/scenarios/litellm/` emits
both the span attributes and the histogram, exercising the `local` path
end-to-end (the SDK derives cost internally; the mock upstream returns
none).

## Validation

- `make check-policies` passes (no `after_resolution` violations)
- `make generate-all` passes, `git diff --exit-code` clean afterwards
- 28 of 28 reference scenarios run. 15 `genai_expected_attribute_missing`
  findings for `cost.amount` are recorded in the snapshots: it is
  `recommended`, and about half the scenarios cannot determine cost, which
  is the point the review made.

Resolves #287


meshailabs:
The `cost.source` enum is the piece that makes this consumable: keeping billed actuals and estimates from silently mixing in one chart is the difference between a dashboard and an invoice dispute. Two points from running cost attribution in production, both small:

1. For `cost.source=pricing_table`, the definition should state that the recorded amount reflects the pricing table in force at computation time, and that consumers never recompute cost from token counts and current prices. Pricing tables change; a later table edit must not rewrite what an old span appears to say. Same emission-time stability concern that landed in the budget limit note on #439, and in #368 and #373 before it.
2. Worth stating explicitly whether an agent or workflow span's `cost.amount` includes the cost of sub-agent operations. Tokens already answer this in #439: own consumption only, sub-agents excluded, which is what makes the attribute summable. Cost should aggregate by the same rule, otherwise team-level rollups double-count exactly the way token rollups used to.

lmolkova (changes requested):
the most important question is how to collect this informantion:
- where does the information comes from
- who collects it
- at which stage

Typical otel auto-instrumentations can't collect this data on their own - cost is not visible on the OpenAI / anthropic / gemini / etc client lib level. 

We can't move this proposal forward without addressing how this information is collected.

meshailabs:
@lmolkova this is the right question to insist on, and it happens to be one we can answer from experience, since we operate one of the components in question in production.

You are right that the client libraries cannot do it: token usage arrives in the provider response, but turning tokens into money takes pricing knowledge, and that knowledge lives elsewhere. The good news is it lives in a small number of well-defined places, and they map onto the proposed `cost.source` enum almost one to one.

Some providers return cost in the response itself (OpenRouter reports it in generation usage). That is `provider`, and there client-level instrumentation genuinely can record it. Everywhere else, the natural producer is the gateway, proxy, or router in the request path: it knows the model, reads the token counts from the response, owns the pricing table as configuration, and can record cost at response completion. LiteLLM already ships per-request cost tracking on exactly this basis, and enterprise LLM gateways do the same. That is `pricing_table`. And backend or collector-side enrichment that joins token counts against a table after the fact is where `estimate` fits, filling a value no in-path producer recorded, once, labeled with its provenance. That last one is the complement of the stability point raised earlier rather than an exception to it: whichever component produces the amount computes it once against the pricing knowledge in force at that moment and labels its source. Filling an absence is fine; rewriting a recorded amount with later prices is the thing to rule out.

So the collection story is less "auto-instrumentation of the client lib" and more "the request-path components that already terminate provider credentials and enforce routing," with labeled downstream enrichment as the fallback. If it would help, a sentence in the attribute note saying which class of instrumentation is expected to emit these would also let client-lib instrumentations know they are off the hook.

changelog.d/287.enhancement.md
```diff
@@ -0,0 +1 @@
+Add `gen_ai.client.operation.cost`, `gen_ai.client.operation.cost.currency`, and `gen_ai.client.operation.cost.source` attributes and `gen_ai.client.operation.cost` metric for per-operation cost tracking.
```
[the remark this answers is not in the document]
elang2: Yup, wrong namespace. Fixed in 26b70e0.

model/gen-ai/registry.yaml
```diff
@@ -944,3 +944,56 @@ attributes:
...
+      any applicable discounts or commitments.
+    examples: [0.0023, 0.015, 0.42]
+    stability: development
+  - key: gen_ai.usage.cost.currency
+    type: string
+    brief: ISO 4217 currency code for the cost value.
+    note: |
+      MUST be a three-letter ISO 4217 alphabetic code (e.g. "USD", "EUR").
+      When the cost source does not specify a currency, instrumentations
+      SHOULD default to "USD" and document this assumption.
```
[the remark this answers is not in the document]
elang2: Good catch. Switched to requiring explicit user config, otherwise the attribute stays unset.

model/gen-ai/spans.yaml
```diff
@@ -46,6 +46,13 @@ attribute_groups:
         requirement_level: recommended
       - ref: gen_ai.usage.output_tokens
         requirement_level: recommended
+      - ref: gen_ai.usage.cost.amount
+        requirement_level: recommended
+      - ref: gen_ai.usage.cost.currency
+        requirement_level:
+          conditionally_required: If `gen_ai.usage.cost.amount` is set.
```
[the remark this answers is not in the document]
elang2: Added cost emission to the litellm scenario. LiteLLM already computes per-request cost so it's the natural fit.

model/gen-ai/metrics.yaml
```diff
@@ -43,6 +43,34 @@ metrics:
       - ref_group: metric_attributes.gen_ai
       - ref: gen_ai.token.type
         requirement_level: required
+  - name: gen_ai.client.operation.cost
+    requirement_level: recommended
+    instrument: histogram
```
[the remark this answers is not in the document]
elang2: Same commit covers this. The litellm scenario now emits both span attributes and the histogram metric.

elang2:
Fair point. Client libs can't do this alone since pricing knowledge doesn't exist at that layer. Agree the spec needs to say that explicitly.

Three collection paths that already exist:

1. Some providers return cost directly in the response. Client instrumentation just passes it through. That's `cost.source=provider`.

2. Gateways and proxies compute it at response time from token counts + a locally configured pricing table. This is the most common production pattern. That's `cost.source=pricing_table`.

3. Collector-side enrichment after the fact using OTTL arithmetic or backend ingest pipelines. That's `cost.source=estimate`.

Added collection guidance to the attribute note in 47b267e. Client-lib instrumentations are off the hook unless the provider response includes cost directly.

Let me know if there's anything else you'd want tightened up or if I'm missing something.

elang2:
Both good points, and both align with decisions already made for tokens.

1. Added a note to `pricing_table` that the value is frozen at computation time. Consumers don't recompute from a later table.

2. `cost.amount` records own operation cost only. Sub agent costs excluded. Same aggregation rule as `gen_ai.usage.input_tokens` and `gen_ai.usage.output_tokens`.

meshailabs:
This lands the collection story cleanly, and the frozen-at-emission-time note on pricing_table is exactly the guarantee consumers need. One small tightening while you are in the file: the enrichment path says it fills the attribute when no in-path component recorded it, which is the right rule, but nothing yet says it normatively. A sentence on the estimate member like "Enrichment MUST NOT replace a cost value already recorded by an in-path component" closes the last hole: without it, a collector pipeline configured with a newer pricing table can silently overwrite a provider-returned actual, and the record stops being trustworthy. With that, nothing else from our side.

elang2:
Done. Added in 6c5401f.

model/gen-ai/registry.yaml
```diff
@@ -951,3 +951,97 @@ attributes:
       what `gen_ai.workflow.name` means in the context of that framework.
     examples: ["multi_agent_rag", "customer_support_pipeline"]
     stability: development
+  - key: gen_ai.usage.cost.amount
```
wolfgangcodes, line 1042: Providers bill differently for input/output or by modality have we considered breaking up cost similarly? Pricing providers like Pydantic do this: https://github.com/pydantic/genai-prices/blob/main/packages/python/genai_prices/types.py#L102-L109

```
class PriceCalculation:
    input_price: Decimal
    output_price: Decimal
    total_price: Decimal
...
```


And if folks are instrumenting (or using a library like Pydantic to do this) it would be useful to have a place to put that breakdown.

alexmojaki: I actually want to deprecate the input/output_price fields in genai-prices because their meaning isn't that clear outside of tokens. Is a sever side tool call like a web search input or output? What about the charging of the request itself?

wolfgangcodes: @alexmojaki That's fair. The thinking is that it's complex and opaque so let's not pretend to have more insights that we do?

When do you expect that change to land? Is there somewhere I can follow along that work/thinking I did a quick search of the Pydantic issues and I didn't see anything that jumped out?

elang2: Yeah this was intentional. The split gets murky fast once you're past simple per-token pricing (tool fees, request charges, cached discounts), so a total felt like the right starting point. Happy to add optional breakdown attributes later if a clean decomposition emerges.

elang2:
@lmolkova Addressed your collection path question. Spec note now covers three paths (provider headers, pricing table lookup, gateway enrichment), added a LiteLLM reference scenario that emits both span attrs and the metric, and removed the USD default. @alexmojaki confirmed total-only fits where genai-prices is going.

Anything else needed or good for another pass?

model/gen-ai/spans.yaml
```diff
@@ -46,6 +46,13 @@ attribute_groups:
         requirement_level: recommended
       - ref: gen_ai.usage.output_tokens
         requirement_level: recommended
+      - ref: gen_ai.usage.cost.amount
```
alexmojaki, line 49: Why have these under `gen_ai.usage.*`? They're not specific to genai, and as noted in the issue, people do claim to want to track other costs similarly.

docs/gen-ai/gen-ai-agent-spans.md
alexmojaki, line 1: Putting these attributes on agent spans raises the same problem as in https://github.com/open-telemetry/semantic-conventions-genai/issues/19. It becomes more difficult to correctly get the total cost of e.g. a trace.

Also, what if the child costs have different sources?

model/gen-ai/registry.yaml
```diff
@@ -951,3 +951,97 @@ attributes:
...
+          brief: Cost computed from token counts and a locally configured pricing table.
+          note: |
+            The recorded value reflects the pricing table in force at the
+            time cost was computed. Consumers MUST NOT recompute cost from
+            token counts and a later pricing table. The value is frozen at
+            emission time.
+        - id: estimate
+          stability: development
+          value: "estimate"
+          brief: Cost is a rough estimate (e.g. from cached pricing that may be stale).
```
alexmojaki: it's really not clear how this differs from pricing_table. tables may always be stale.

model/gen-ai/registry.yaml
```diff
@@ -951,3 +951,97 @@ attributes:
...
+      `gen_ai.usage.output_tokens` and ensures the attribute is safely
+      summable across a trace without double-counting.
+
+      ## Collection guidance
+
+      Cost is typically not available to client-library auto-instrumentations
+      because pricing knowledge does not exist at that layer. The expected
+      producers of this attribute are:
+
+      1. **Provider response** (`cost.source=provider`): Some providers
```
alexmojaki: why describe sources here again? these differ from what the cost.source attribute says which is confusing

model/gen-ai/registry.yaml
```diff
@@ -951,3 +951,97 @@ attributes:
...
+         return cost directly in the response body or headers. Client-level
+         instrumentation CAN record this value when present.
+      2. **Gateway, proxy, or router** (`cost.source=pricing_table`): LLM
```
alexmojaki: pricing tables are not limited to gateways, this header is confusing.

alexmojaki: from the client's perspective, the gateway pretty much is the provider. if they return a cost, why shouldn't i use cost.source=provider? i didn't compute the cost myself. and i don't necessarily know what the source of the gateway's cost is.

model/gen-ai/registry.yaml
```diff
@@ -951,3 +951,97 @@ attributes:
...
+         gateways that terminate provider credentials typically own a pricing
+         table as configuration. They compute cost at response-completion
+         time from token counts in the response. This is the most common
+         production pattern.
+      3. **Backend enrichment** (`cost.source=estimate`): A collector
```
alexmojaki: again, this is confusing me. this just sounds like pricing_table. why is `estimate` described as 'enrichment'?

model/gen-ai/registry.yaml
```diff
@@ -951,3 +951,97 @@ attributes:
       what `gen_ai.workflow.name` means in the context of that framework.
     examples: ["multi_agent_rag", "customer_support_pipeline"]
     stability: development
+  - key: gen_ai.usage.cost.amount
+    type: double
+    brief: The monetary cost of a single GenAI operation.
+    note: |
+      The cost in the currency specified by `gen_ai.usage.cost.currency`.
+      When the provider includes cost in its response (e.g. via a header or
+      response field), that value SHOULD be used directly. When cost is not
```
alexmojaki: Consider https://docs.merge.dev/merge-gateway/cost/cost-governance-and-savings#per-call-cost-in-the-api-response

> Use it for per-request attribution, not for reconciliation: it is provider cost, so it does not include your plan rate, the list-price basis used for BYOK requests, or server-tool charges such as web search. The dashboard above remains the authority on what you are billed.

So this gateway returns a cost and explicitly says it's not what's *actually* charged. I could theoretically calculate a more accurate cost with my own pricing table that accounts for web search tool calls. How should this be handled?

model/gen-ai/spans.yaml
elang2: You're right. Nothing in the schema ties this to genai semantically. Kept the naming generic enough that a namespace move wouldn't require breaking changes.

docs/gen-ai/gen-ai-agent-spans.md
elang2: Valid concern. The spec says parent spans record only their own cost, not a sum of children. For trace totals, consumers sum spans that have gen_ai.usage.cost set. I'll add a note clarifying that mixed sources across spans are expected.

model/gen-ai/registry.yaml
elang2: Agreed. Collapsing both into `local`. If the instrumentor computed it, that's the source.

model/gen-ai/registry.yaml
elang2: Yeah, removing that section. The enum values on `cost.source` will be the single source of truth.

model/gen-ai/registry.yaml
elang2: Removing that header.

elang2: Exactly. Simplifying to `provider` (received from upstream, gateway included) and `local` (computed by the emitting instrumentor). From your side, the gateway is just another upstream.

model/gen-ai/registry.yaml
elang2: Same as above. Dropping `estimate` entirely.

model/gen-ai/registry.yaml
elang2: Think the right call is dropping "MUST NOT replace." If the operator has a better number locally, let it override. `cost.source` shows what won. Does that work for your use case?

model/gen-ai/spans.yaml
alexmojaki: > Kept the naming generic enough that a namespace move wouldn't require breaking changes.

i don't follow

model/gen-ai/registry.yaml
alexmojaki: i think allowing a choice in whether or not to override sounds reasonable. but i'm also wondering whether it should be possible to keep both. i guess probably not, we shouldn't complicate the common case, there should be a simple attribute which is a number, not an array or something.

model/gen-ai/registry.yaml
```diff
@@ -951,3 +951,97 @@ attributes:
...
+      returned by the provider, instrumentations MAY compute it from token
+      counts and a user-supplied pricing table.
+
+      Instrumentations MUST NOT hard-code pricing assumptions. If no
+      cost value is available from the provider or from user configuration,
+      this attribute MUST NOT be set.
```
alexmojaki: I don't understand this. Pydantic AI and Logfire instrumentation use the data in genai-prices. We don't expect the user to provide the costs of models, that would be an undue burden.

model/gen-ai/registry.yaml
```diff
@@ -951,3 +951,97 @@ attributes:
...
+      contracted price, effective price after discounts), this attribute
+      SHOULD record the effective cost, the amount actually charged after
+      any applicable discounts or commitments.
+
+      For parent spans that invoke sub-operations (e.g. agent or workflow
+      spans), this attribute records only the cost of the span's own
+      operation. Cost of child spans is NOT included. This matches the
+      aggregation rule for `gen_ai.usage.input_tokens` and
+      `gen_ai.usage.output_tokens` and ensures the attribute is safely
+      summable across a trace without double-counting.
```
alexmojaki, line 1068: I don't see anything suggesting that this is how tokens are currently treated.

model/gen-ai/spans.yaml
elang2: Apologies. I re-read #287 . The issue proposes **operation.cost** as a generic attribute. Logfire to ship that and seem like Red Hat  and Honeycomb like that approach too.  But if you look at **gen_ai.usage.cost** it has wider adoption in other tools like langfuse, litellm etc. 

I was thinking that gen_ai.usage.* made better sense given that we measure tokens and tool calls. 

@lmolkova , does keeping **gen_ai.usage.cost.***  make sense given the existing adoption in other tools or would you prefer a different direction.   If we go the **operation.cost** approach we likely need to add a new model/operation/ directory right ?

model/gen-ai/registry.yaml
elang2: Agreed. Single value, **cost.source** shows where it came from. I'II remove the "MUST NOT replace" from the estimate note . Override policy belongs to the pipeline, not the attribute definition.

model/gen-ai/registry.yaml
elang2: Agreed. How about this :

"When cost is not returned by the provider,  instrumentations MAY compute it from token counts and a known pricing table (e.g., maintained as a library dependency or provided as operator configuration) . Instrumentations SHOULD NOT inline price  constants in source code" 

Does this work ?

model/gen-ai/registry.yaml
elang2: I agree. I'll remove this para. For context, #460 also addresses this problem for tokens. Same principle for 
cost: record what you received or computed, don't aggregate children

docs/gen-ai/gen-ai-agent-spans.md
elang2: Correction. Reviewed #19  again and i see the double counting problem . Should I remove cost from agent span definition ( consistent with #460 for cache tokens) , or keep it only when provider returns single total for whole operation?

alexmojaki: are there services where an agent runs remotely and returns a total cost?

removing cost to start seems safest to me in terms of compat, we can add it later.

model/gen-ai/registry.yaml
alexmojaki: why specify anything about this? "inline price constants in source code" does sound bad architecturally, but a client instrumentation directly maintaining pricing data in a more robust way sounds ok.

docs/gen-ai/gen-ai-agent-spans.md
wolfgangcodes: I agree with @alexmojaki that removing seems safest for now.

model/gen-ai/registry.yaml
wolfgangcodes: Agreed. This seems like implementation guidance for instrumentation providers and out of scope fora definition.

model/gen-ai/spans.yaml
```diff
@@ -46,6 +46,13 @@ attribute_groups:
...
+        requirement_level: recommended
+      - ref: gen_ai.usage.cost.currency
+        requirement_level:
+          conditionally_required: If `gen_ai.usage.cost.amount` is set.
+      - ref: gen_ai.usage.cost.source
+        requirement_level: recommended
```
wolfgangcodes: I think this should be `conditionally_required`  when wouldn't this be available? 🤔

docs/gen-ai/anthropic.md
```diff
@@ -129,41 +132,99 @@ version identifier used by that system.
...
+   return cost directly in the response body or headers. Client-level
+   instrumentation CAN record this value when present.
+2. **Gateway, proxy, or router** (`cost.source=pricing_table`): LLM
+   gateways that terminate provider credentials typically own a pricing
+   table as configuration. They compute cost at response-completion
+   time from token counts in the response. This is the most common
+   production pattern.
+3. **Backend enrichment** (`cost.source=estimate`): A collector
+   processor or backend pipeline joins token counts against a pricing
+   table after the fact, filling the attribute when no in-path
```
wolfgangcodes: Overall I find the _values_ confusing. They do not map clearly to the categories since pricing tables are _also_ estimates.  Maybe `provider`, `gateway`, `ingest-pipeline`? 

enum values that more clearly maps to the categories:
* furnished by the provider directly
* returned in the response form a gateway/broker/harness
* Enriched in the ingest pipeline

model/gen-ai/registry.yaml
elang2: Yeah I think single value is the way to go for now. If someone ends up needing both a provider-reported and a locally-computed cost on the same span that's really two different attributes with different trust semantics, and I'm not sure there's a use case driving that yet. Keeping it to one number plus cost.source means the attribute stays summable without extra logic.

docs/gen-ai/gen-ai-agent-spans.md
elang2: Note added in 8580a27. Different spans in the same trace can have different cost.source values (one from the provider, another computed locally). Consumers summing cost across a trace should expect that mix.

model/gen-ai/registry.yaml
elang2: Ah sorry if that was confusing. That's exactly the `local` case. I see Logfire pulling from genai-prices and computing the amount, that maps straight to `cost.source=local`. Nobody needs to manually provide anything. "Local" just means the instrumentor did the calculation rather than receiving a pre-computed value from upstream.

model/gen-ai/spans.yaml
elang2: Good catch, yeah. Updated to `conditionally_required` when `cost.amount` is set. 8580a27

docs/gen-ai/anthropic.md
elang2: Yeah agreed, the old values were hard to reason about. I collapsed it down to two. `provider` (the value came from upstream, whether that's the actual model provider or a gateway in front of it) and `local` (your component computed it). A gateway returning cost in the response is just another upstream from the client's perspective, so that falls under `provider`. 8580a27

model/gen-ai/spans.yaml
elang2: My bad, worded that poorly. All I meant was the leaf names (cost.amount, cost.currency) don't have gen_ai in them, so if they ever moved to a broader namespace you'd only change the prefix. But that's hypothetical anyway. gen_ai.usage is where usage-adjacent stuff lives and it fits fine here.

elang2:
Looking at who's actually emitting cost today. LiteLLM already ships a `gen_ai.usage.cost` histogram and `gen_ai.cost.*` span attributes from their internal pricing tables. Pydantic AI/Logfire computes it via genai-prices and emits an `operation.cost` metric. Both are doing this under their own schemas already, so landing this convention prevents the namespace from forking further.

Rebased over #440, simplified the enum to `provider`/`local` per alexmojaki's feedback, made `cost.source` conditionally_required per wolfgangcodes, and addressed all inline threads. CI should be green shortly.

docs/gen-ai/gen-ai-agent-spans.md
elang2: Done. Split cost into its own group (`attributes.gen_ai.usage.cost`), only referenced by the inference span now. Agent spans get tokens but not cost. Thanks for flagging it.

elang2:
Merged from main and addressed the cost-on-agent-spans feedback inline.

For context on why this split matters beyond the design preference: I ran 8 frameworks against a deterministic mock with a request ledger (same 4 LLM calls, 3 tool calls, budget=3) and got 4 different consumed values across 5 distinct counting units:

| Framework | consumed | unit |
|-----------|----------|------|
| LangChain / CrewAI / SK | 3 | tool_cycles / auto_invoke_rounds |
| OpenAI Agents / LlamaIndex | 4 | llm_invocations |
| AutoGen | 5 | composite_messages |
| LangGraph | 8 | graph_nodes |

([full data](https://github.com/elang2/agent-budget-semantics/releases/tag/v0.3.0))

Cost inherits the same aggregation problem if it propagates to agent spans. Inference-only sidesteps it.

@lmolkova I responded to the collection-path question from Aug 10 in [this comment](https://github.com/open-telemetry/semantic-conventions-genai/pull/443#issuecomment-5248928932) and the [follow-up on Aug 17](https://github.com/open-telemetry/semantic-conventions-genai/pull/443#issuecomment-5311559526). Let me know if anything's still open. Ready for re-review whenever.

xrmx:
> Looking at who's actually emitting cost today. LiteLLM already ships a `gen_ai.usage.cost` histogram and `gen_ai.cost.*` span attributes from their internal pricing tables.

OpenRouter also sends these when you enable observability:

```
     -> gen_ai.usage.input_cost: Double(0)
     -> gen_ai.usage.output_cost: Double(0)
     -> gen_ai.usage.total_cost: Double(0)
```

model/gen-ai/registry.yaml
alexmojaki: but the lines remain

      Instrumentations MUST NOT hard-code pricing assumptions. If no
      cost value is available from the provider or from user configuration,
      this attribute MUST NOT be set.

model/gen-ai/registry.yaml
alexmojaki: are you still going to do this? the problem remains

model/gen-ai/registry.yaml
```diff
@@ -1025,3 +1025,98 @@ attributes:
...
+      ## Collection guidance
+
+      Cost is typically not available to client-library auto-instrumentations
+      because pricing knowledge does not exist at that layer. The expected
+      producers of this attribute are:
+
+      1. **Provider response** (`cost.source=provider`): Some providers
+         return cost directly in the response body or headers. Client-level
+         instrumentation CAN record this value when present.
+      2. **Gateway, proxy, or router** (`cost.source=local`): LLM
```
alexmojaki: this doesn't sound like it should be 'local'

model/gen-ai/registry.yaml
elang2: Deleted in `aeb9284`. It ruled out the genai-prices pattern.

model/gen-ai/registry.yaml
elang2: Deleted in `aeb9284`. It was a duplicate of the enum that had drifted. Sourcing lives on the members now; invariants on `.amount`.

model/gen-ai/registry.yaml
elang2: You're right. If cost came back in the response, it's `provider`, whoever answered. `local` now means just "the recorder computed it". Rewrote the member notes in `aeb9284`. Filing `pricing.source_ref` as a follow-up and tagging you.

elang2:
Re-requesting review on the Aug 10 collection question: enum members now carry the collection story. `provider` covers cost present in the response the client received (from the provider or a priced intermediary). `local` covers the emitter computing at recording time from pricing data it owns (configured table, or bundled pricing like genai-prices). Backend enrichment is out of scope. LiteLLM reference scenario emits both attrs and histogram (`source=local`: the SDK derives internally; the mock upstream returns no cost). Threads with @alexmojaki resolved. `pricing.source_ref` and per-token-class breakdown deferred.

On #439: cost is comparable across providers where iteration counts weren't. What diverges is attribute plumbing, which is what a convention fixes.

meshailabs:
@xrmx that's a real collision. `gen_ai.usage.total_cost` and `gen_ai.usage.cost.amount` are the same number under two names.

Does OpenRouter send a currency with those? I don't see one in your paste, and the current text says don't set the amount if you can't determine a currency. If there isn't one, renaming won't be enough to converge.

They also split input and output, while this PR defers the breakdown. Is that deferred as in never, or deferred as in later under `cost.*`? Anyone migrating off their shape needs to know which.

We price requests at our gateway from our own table, and we compute input and output separately before summing. Right now that split has nowhere to go.

elang2:
@meshailabs Yes, one name has to migrate, but fewer emitters than it looks. [LiteLLM's `semconv.py`](https://github.com/BerriAI/litellm/blob/e55dbaf34710783ce263a2d1a3039953c4a05148/litellm/integrations/otel/model/semconv.py#L338-L348) already sets `TOKEN_COST = "gen_ai.usage.cost"` with a comment describing it as "the name backends already query for spend." Langfuse's OTel processor [already reads that name](https://github.com/langfuse/langfuse/blob/471e150a87540ff8050b96f1233717f4c1767559/packages/shared/src/server/otel/OtelIngestionProcessor.ts#L2606-L2607) into `observation.totalCost`. The OTel Collector has a [`genainormalizerprocessor`](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/genainormalizerprocessor) waiting for a canonical cost name to exist, and neither the official OTel language contribs nor OpenLLMetry emit an OTel cost attribute today. This lands into open space more than into a rename.

On currency, there's no currency field in the export xrmx pasted, and [OpenRouter's policy](https://openrouter.ai/docs/faq) is USD-denominated credits, so any emitter turning that into an OTel attribute has already made an implicit assumption. The conditional-requirement rule surfaces it at the point of emission. Real migration cost since nobody emits currency today, but the alternative is aggregation that silently mislabels values once a provider ships regional pricing.

The breakdown is later under `cost.*`, not never. It would sit next to the existing `attributes.gen_ai.usage.breakdown` group in `spans.yaml`, with per-class amount attributes (input, output, cache read, cache write, reasoning, tool) sharing the parent's currency and source. [LiteLLM's `cost_breakdown` payload](https://github.com/BerriAI/litellm/blob/e55dbaf34710783ce263a2d1a3039953c4a05148/litellm/integrations/opentelemetry.py#L2383-L2390) already emits a `gen_ai.cost.{class}_cost` pattern. [OpenInference has an eleven-attribute breakdown](https://github.com/Arize-ai/openinference/blob/d773c1ac1c7bcfef565e6280932af68d388531e7/spec/semantic_conventions.md#L88-L98) under `llm.cost.prompt_details.*` and `llm.cost.completion_details.*`. [Langfuse #13807](https://github.com/langfuse/langfuse/issues/13807) shows what happens when a backend recomputes Anthropic prompt-caching costs without per-class data: the derived value lands at roughly 40% of true spend. I split v0.1 at the total because the per-class shape has open questions I did not want to freeze here. If your gateway computes input and output separately before summing, that is the shape the follow-up needs. Want to open the breakdown issue with your gateway's schema as the starting example?

model/gen-ai/registry.yaml
```diff
@@ -1025,3 +1025,77 @@ attributes:
...
+          stability: development
+          value: "local"
+          brief: The emitting component computed the value itself.
+          note: |
+            No cost was present in the response; the emitting component
+            computed the value at recording time from token counts and
+            pricing data it owns, such as a configured pricing table or
+            maintained pricing data bundled as a versioned dependency
+            (e.g. `genai-prices`). Client-library instrumentations are not
+            expected to emit cost unless it is present in the response.
```
alexmojaki, line 1109: `logfire.instrument_openai()` uses `genai-prices` to emit cost. Why wouldn't this be expected for client-library instrumentations?

alexmojaki:
> We price requests at our gateway from our own table, and we compute input and output separately before summing. Right now that split has nowhere to go.

How do you classify non-tokens costs such as builtin tool calls or request counts?

alexmojaki (approved):

meshailabs:
@alexmojaki we don't classify them today. The gateway prices input and output tokens from its table; builtin tool calls, per-request fees, and image inputs are invisible in the per-request figure and surface later as drift against the invoice.

Which is the argument for explicit classes: an emitter that cannot price a class should omit it visibly, not fold the gap into the total. A total-only attribute cannot distinguish "fully priced" from "tokens only".

elang2:
@lmolkova gentle nudge for a re-review when you have time. The changes since your Aug 10 review (commits 92931209, 0958ceed, c404e8f8) reframe cost.source as a receive-vs-derive test, which addresses the collection-stage question.

model/gen-ai/registry.yaml
elang2: You were right, and it's gone. `e31a9ad` dropped the sentence saying client-library instrumentations aren't expected to emit cost — it contradicted the enum member it was documenting, since `local` already describes exactly what Logfire does. That note now names maintained pricing data bundled as a versioned dependency, `genai-prices` included, as a first-class case rather than an exception.

The text that replaced it says cost is recorded by whichever component can determine it: any client instrumentation when the value is in the response, a component configured with pricing data when it isn't. An instrumentation that ships pricing data is squarely in scope. The only thing not expected is going and acquiring pricing data you don't have.

elang2:
@lmolkova consolidating my answers here:

| Your question | Where it's answered | What it says |
|---|---|---|
| where does it come from | `gen_ai.usage.cost.source` enum notes | `provider` — "present in the response the instrumented client received, whether from the model provider directly or from an intermediary (gateway, proxy, router) that priced the request". `local` — "the emitting component computed the value at recording time from token counts and pricing data it owns" |
| who collects it | `gen_ai.usage.cost.amount` note, new | "recorded by whichever component can determine it. Any instrumentation of the client can record it when the value is present in the response … When it is absent, only a component configured with pricing data can record it. A component with neither is **not expected to acquire pricing data in order to record cost**" |
| at which stage | `gen_ai.usage.cost.amount` note | "reflects the pricing data in force at recording time and MUST NOT be recomputed against later pricing data. Backend or collector-side enrichment that reconstructs cost after the fact is out of scope" |

An intermediate revision after your review picked up four normative changes nobody had asked for: the requirement-level escalation on `cost.amount`, an inference-only restriction on the metric, and two smaller ones. 738270b removes them per CONTRIBUTING.md, the PR description has a *Follow-ups* section covering where they go next, and the argument for them stays in 80325bb's message for whoever picks them up. What's left is the collection model above plus plain fixes.

So the one thing I'd like your read on: does that collection model answer the review? If it does, I'd appreciate another look when you have time.

PR #444, opened 2026-08-09
run conformance in strict mode, fix or declare violations

lmolkova:
some reference scenarios have conformance violations. Fix them or suppress violations. 

Most of the violations come from mock not returning properties, a few more from required things not being available for library, i.e. some semconv changes needed to relax requirements, will address these in a follow up.


reference/scenarios/aws-bedrock-agentcore/conformance.yaml
```diff
@@ -11,3 +11,23 @@ env:
 scenarios:
   reference:
     run: uv run --frozen --python 3.12 --project . python scenario.py
+    expected_violations:
+      # Memory operations are experimental and not yet in the semconv operation name enum
+      - id: genai_operation_name_unknown
```
lmolkova, line 16: todo - not right

JWinermaSplunk:
One overarching comment: for the updated `conformance.yaml `files, could we add a context to each `expected_violations` entry when applicable? Without a context, the conformance runner accepts any violation with the same ID, so a new unrelated violation can be masked by an existing expected violation.
For instance, if the current `genai_span_kind_unexpected` exception is intended only for an internal chat span, a later change that emits an internal embeddings span would produce the same violation ID and also pass today. Scoping the exception to `operation: chat` and `kind: internal` would allow the known chat violation while causing strict mode to fail for the new embeddings violation.

PR #445, opened 2026-08-09
Add agent lifecycle events for async and long-running executions

### LINKED ISSUE #159: Add async/long-running agent lifecycle events: gen_ai.agent.paused / .resumed / .checkpointed
## TL;DR

Propose adding standard span events for long-running async agent lifecycles: `gen_ai.agent.paused`, `.resumed`, `.checkpointed`, plus supporting `.started`, `.completed`, `.failed`. These complement #55's `gen_ai.task.status` attribute (which captures final state) by capturing **timing of mid-execution state transitions** — critical for HITL, scheduled, and durable agents.

Filed separately from #55 per @Cirilla-zmh's suggestion that this is beyond #55's scope. Full background in [the original gist](https://gist.github.com/meshailabs/51897e8efb49ef88a6f767112cdc325b).

## Why it's distinct from `gen_ai.task.status`

#55's `gen_ai.task.status` (values: `pending / in_progress / completed / failed`) is useful as a **final-state marker** on a task span. It doesn't capture **when** or **why** an agent transitioned mid-execution: when it paused for human input, when it persisted a checkpoint, when it resumed.

For short-lived workflow execution, the status attribute is enough. For long-running, human-in-the-loop, scheduled, or durable agents — increasingly common as agents move to async execution patterns (see [Zak Knill, "All Your Agents Are Going Async", 2026](https://zknill.io/posts/all-your-agents-are-going-async)) — timing matters too. An agent that paused for 3 hours waiting on human approval needs observability on *when* and *why* it paused, not just whether it ultimately completed.

## Adjacent issues that don't cover this

- **#94** (Generic Grouping Attributes) — logical grouping across spans, not lifecycle state transitions
- **#81** (ReAct Iterations Spans) — iteration counting within a reasoning loop, not pause/resume
- **#89** (`gen_ai.repair.*`) — post-detection failure remediation, not in-flight lifecycle

## Proposed events

Span events that long-running agents MAY emit, nested inside a `gen_ai.workflow`, `gen_ai.agent`, or `gen_ai.task` span:

| Event | Description | Stability |
|---|---|---|
| `gen_ai.agent.started` | Agent execution has begun; resources loaded | Development |
| `gen_ai.agent.paused` | Agent has paused execution awaiting external input | Development |
| `gen_ai.agent.resumed` | Agent has resumed execution after a pause | Development |
| `gen_ai.agent.checkpointed` | Agent state has been persisted for durability/recovery | Development |
| `gen_ai.agent.completed` | Agent execution ended successfully | Development |
| `gen_ai.agent.failed` | Agent execution ended in failure requiring attention | Development |

## Suggested event attributes

| Attribute | Description | Type |
|---|---|---|
| `gen_ai.agent.pause.reason` | Why execution paused (`human_input`, `external_system`, `rate_limit`, `scheduled_delay`) | string |
| `gen_ai.agent.checkpoint.id` | Identifier for the persisted checkpoint, for recovery correlation | string |
| `gen_ai.agent.execution.id` | Identifier for a single end-to-end execution run (can differ from `gen_ai.agent.id` when the same agent runs repeatedly) | string |

## Example: nightly reconciliation agent

```
span: gen_ai.agent invoke_agent reconciliation-agent [4h duration]
  event: gen_ai.agent.started                       (t+0s)
  span: execute_tool fetch_invoices                 (t+5s)
  span: chat claude-sonnet-4-7                      (t+12s)
  event: gen_ai.agent.checkpointed (checkpoint.id=ck_001)  (t+900s)
  event: gen_ai.agent.paused (reason=human_input)   (t+1800s)
  event: gen_ai.agent.resumed                       (t+14000s, 3h later)
  span: execute_tool post_journal_entry             (t+14010s)
  event: gen_ai.agent.completed                     (t+14200s)
```

Without these events, reconstructing "what was the agent doing from t+900s to t+14200s" requires custom instrumentation.

## Open question for the SIG

Should these be events on the existing `invoke_agent` span (simpler), or should long-running async agents get a new span kind entirely (more honest about the different execution model)?

## Implementation status

MeshAI Labs has implemented these as custom attributes in our platform. Happy to:
1. Refine into a formal spec PR once the SIG indicates direction
2. Contribute a reference implementation showing the attributes across CrewAI / AutoGen / LangGraph

cc @Cirilla-zmh per #55 redirect

meshailabs:
## Description

Adds lifecycle events for agent executions that suspend and resume: `gen_ai.agent.paused`, `gen_ai.agent.checkpointed`, `gen_ai.agent.resumed`, with `gen_ai.agent.execution.id`, `gen_ai.agent.pause.id`, `gen_ai.agent.pause.reason`, `gen_ai.agent.checkpoint.id`, and the `gen_ai.agent.resumed_from.type` / `.id` pair. Resolves #159.

Draft while the prototypes below settle the shape. Pause deadline, pause resolution, the resolved event, and the `rate_limit`, `scheduled_delay`, `expired` members were removed after review; nothing in this repo produces them.

## Motivation

Durable executions pause on human input or an external system, persist state, and continue later, often in another process. Incident triage and audit need to see where a run stopped, what was persisted, and which record a resumed segment continued from. Prior art: LangGraph `interrupt()` with checkpointers, CrewAI flow persistence and `HumanFeedbackPending`, ADK `ResumabilityConfig`, OpenAI Agents SDK `RunState`.

## Prototype

Real instrumentation rather than a scenario, per review: a branch of `opentelemetry-instrumentation-genai-langchain` emits the three events from LangGraph's public `on_interrupt` / `on_resume` lifecycle callbacks and the checkpointer, with a captured telemetry sample and design note: https://github.com/meshailabs/opentelemetry-python-genai/tree/feat/langgraph-lifecycle-events. CrewAI and ADK follow.

Known gaps from that prototype: LangGraph exposes no id spanning suspend and resume, so `execution.id` has no producer yet; `pause.reason` likewise; and the `pause` member of `resumed_from.type` has none, since LangGraph always resumes from a checkpoint. All three depend on the CrewAI and ADK prototypes. This PR carries no reference scenario: the earlier hand-emitted langchain scenario could not produce `execution.id` and was removed; the scenario returns once it can run the instrumentation as its emitter.

## Checklist

- [ ] Reference scenarios updated for affected libraries (none yet; pending the instrumentation)


meshailabs:
Updated since the initial draft; summarizing the changes for review:

- The reference scenario now instruments LangGraph's durable-interrupt API (`interrupt()` / `Command(resume=...)` with a checkpointer, `langgraph==1.2.10`) instead of a synthetic gate. Attributes come from runtime state: the pause id from the returned `Interrupt` object, the checkpoint id from `get_state`, the pending pause and its deadline read back from persisted state for the expiry sweep. The conformance runner passes and `uv.lock` plus `data.json` are committed.
- Working against a real framework surfaced one refinement to the shape #159 converged on, flagged in the body: delivering a refusal re-enters execution in LangGraph (the decline path runs and reaches END), so the `gen_ai.agent.resumed` note now separates outcome (the resolution record) from continuation (the resumed event), and consumers determine the outcome only from the resolution record. Calling it out explicitly since the thread had settled on the stricter no-resumed-after-refusal form.
- resumed_from linkage: a segment reconstructed from a persisted checkpoint names the checkpoint; the pause linkage is carried by the resolution event. This keeps correlation unambiguous when one execution holds several pauses and checkpoints.
- Known limits are declared under Capture notes in the body and the scenario README: resolution and reason semantics are gate-owned in LangGraph, the thread-id mapping assumes one logical execution per thread, and `gen_ai.agent.id` is absent per the registry's own hosted-agent guidance (the events carry `gen_ai.agent.name`, now declared). A second scenario on a framework whose public API owns the decision (OpenAI Agents SDK tool-approval) is offered as a follow-up if wanted in this PR.

Docs and reports are regenerated, so generated-docs and the agent-lifecycle scenario job should now pass.

StevenMih:
Hi. Same-day as promised. 

Read the full diff — registry, events, and the scenario. Its a faithful encoding of where #159 landed and I'm in support of it moving out of draft. 

Notes in order of how much I care:

The post-refusal resumed refinement: lgtm. Separating outcome (the resolution record) from continuation (the resumed event) is cleaner than the stricter no-resumed-after-refusal form the thread converged on, and it's the reality-tested one — LangGraph's decline path re-entering to reach END is exactly the case the stricter form couldn't express without lying. The guard sentence ("consumers MUST determine the outcome of a pause from its resolution record, never from the presence or absence of a resumed event") carries the invariant intact. One ordering nit worth a sentence in the gen_ai.agent.resumed note: when a decline-path segment does emit resumed, the resolution record SHOULD already exist or be emitted with it. The scenario does this naturally, but as written a consumer could observe a resumed-after-pause with the refusal record still in flight and have no way to tell it from an approved resume until the resolution lands. That's just the pending invariant applied to the gap; making the producer ordering explicit closes it.

gen_ai.agent.pause.id on paused: the conditional requirement is predicated on knowledge the producer doesn't have at emit time. "When a later gen_ai.agent.resumed or gen_ai.agent.pause.resolved event is expected to reference this boundary" asks the instrumentation to predict the future. Since pause.resolved requires pause.id, any pause that could be resolved needs the id — which is every pause. I'd make it required (every framework in the prior-art list exposes a pause/interrupt identifier), or, if there's a known producer that can't, recommended with a note stating the actual consequence: a pause emitted without an id can never be resolved or resumed-from.

The deadline question (@Rul1an's "more than one cause" case). The registry note is honest that MUST-NOT-without-deadline constrains producer behavior and isn't validatable from the record. It could be: an optional attribute on paused (say gen_ai.agent.pause.deadline, RFC 3339) carrying the configured deadline would let a consumer distinguish "pending, deadline ahead," "pending, deadline passed, expiry record in flight or producer dead," and "pending indefinitely, no deadline" — without touching the vocabulary or the silence-stays-consumer-policy rule. It gives the consumer's policy something to be a policy about. Not worth holding the PR on; fine as a follow-up if the SIG prefers minimal surface.

Scenario notes, all minor: the expiry sweep's clock is synthetic (now = deadline + timedelta(minutes=1)), so the now > deadline check is vacuously true — the honest part (pause id and deadline read back from persisted state) is real, but one README line saying the sweep time is simulated would keep anyone from citing the clock as runtime-derived too. In the approved flow, resumed is emitted before graph2.invoke(...) actually continues; if the invoke throws, a continuation is recorded that never happened — emitting at segment start is defensible, just noting the choice is implicit. The capture notes are exactly the disclosures I'd want, and the OpenAI Agents SDK second scenario would strengthen the case that pause.resolution is framework-derivable, but it doesn't need to be in this PR.

The flat-pair note reads faithful to @Rul1an 's reasoning to me, including the resolve-to-interpret point, though he's the authority there. And the resumed_from linkage decision — checkpoint-reconstructed segments name the checkpoint, pause linkage rides the resolution — is the right disambiguation for executions holding several pauses and checkpoints; keep it exactly as is.

meshailabs:
All applied, and thank you for the same-day turnaround and the quality of it. Point by point:

- Ordering: the `gen_ai.agent.resumed` note now states that a decline-path resumed SHOULD have its resolution record already present or emitted together with it. You are right that it is just the pending invariant applied to the gap.
- `gen_ai.agent.pause.id` is now required, with your rationale in the note: whether a pause will later be resolved is not knowable at emit time, and a pause recorded without an id can never be resolved or resumed from. The predict-the-future conditional is gone.
- Deadline: added now rather than deferred, as `gen_ai.agent.pause.deadline` (RFC 3339, recommended when configured). Two independent reviews converged on the same gap, and your three-state framing (deadline ahead; deadline passed with the record in flight or the producer dead; no deadline at all) is the attribute note nearly verbatim. The `expired` rule now points to it, and the scenario records it from the interrupt payload.
- Scenario: the README now says the expiry sweep's clock is simulated (the pause id and deadline are read from persisted state; the timing is staged), and the emit-before-invoke choice in the approved flow is an explicit comment.
- Flat pair and checkpoint linkage: untouched, as you suggested.

Moving this out of draft on the strength of your review. @Rul1an the flat-pair attribute note quotes your #159 reasoning nearly verbatim and StevenMih defers to you on it; your look remains welcome at any point in the review.

The OpenAI Agents SDK scenario stays a committed follow-up.

changelog.d/+.enhancement.md
```diff
@@ -0,0 +1,7 @@
+Add agent lifecycle events for asynchronous and long-running executions:
+`gen_ai.agent.paused`, `gen_ai.agent.checkpointed`, `gen_ai.agent.resumed`,
+and `gen_ai.agent.pause.resolved` (resolution: `approved` / `refused` /
+`expired`), with `gen_ai.agent.execution.id` correlating execution segments
+across suspend/resume boundaries and the flat
+`gen_ai.agent.resumed_from.type` / `.id` pair identifying the suspended
+boundary a resumed segment continues from.
```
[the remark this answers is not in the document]
meshailabs: Collapsed to a single logical line in ee42112 (and updated to mention `gen_ai.agent.pause.deadline`, added since the fragment was first written).

Rul1an:
Read the registry, the events and the scenario at `ee42112`. Taking you up on the look, since the flat pair started with me.

**On whether the note does the reasoning justice: mostly, with one substitution worth making.** The note gives the resolve-to-interpret argument, which is the right argument for *carrying the type at all*. It is not the argument for carrying it *flat*, and as written the `so` does not connect: a nested `resumed_from: {type, id}` would equally let a consumer read the type without resolving a counterpart. The reason flatness specifically was worth arguing is the one that did not ship, namely the authoring guidance to represent complex values as flat attributes wherever possible and to assume backends do not index their individual properties, with `gen_ai.tool.type` sitting next to `gen_ai.tool.call.id` as the precedent already in this registry, and grouping resumptions by boundary type staying an ordinary query. Both reasons are true and they justify different decisions. Swapping the `so` clause for the indexing one, or keeping both with the resolve-to-interpret half attached to the type existing rather than to its shape, would make the note say why the encoding is what it is.

`MUST be recorded together or not at all` is the sentence that makes the type load-bearing rather than decorative, since a half-populated pair reads as a typed reference and answers nothing. Worth keeping exactly as is.

**One thing I do not think the record supports yet, on the linkage decision I otherwise agree with.** Its stated justification is the execution holding several pauses and several checkpoints. Given that case, nothing joins a checkpoint to the pause it was taken at: `gen_ai.agent.checkpointed` carries `execution.id` and `checkpoint.id`, `gen_ai.agent.pause.resolved` carries `execution.id`, `pause.id` and `resolution`, and none of the eight new attributes bridges them. So a consumer holding `resumed_from.type=checkpoint` can reach the checkpoint but not the resolution that authorized continuing from it, except by ordering records that may be sampled out or in another trace, which is the reconstruction the pair exists to avoid. A `gen_ai.agent.pause.id` on `checkpointed`, recommended when the checkpoint is taken at a pause boundary, would close it. The scenario cannot show this either way, since each flow uses its own thread id and holds one pause.

Two small ones. The scenario only ever emits `resumed_from.type: checkpoint` (`scenario.py:154`), which follows from LangGraph checkpointing at every interrupt; the `pause` member's brief already settles precedence for a producer holding both, so nothing needs changing in the note, but a README line saying `pause` is unexercised here would stop the scenario reading as coverage of both members. And on the deadline, which you have already landed: separating an expiry in flight from a producer that died needs producer liveness, a lease or a heartbeat, which is a different signal from this vocabulary rather than a missing attribute in it. That is what makes the consumer-policy sentence a stopping point rather than a placeholder.

(AI-assisted; I ran the reads and am responsible for the claims.)

meshailabs:
Applied at 5426dfb, both substantive points:

- The `resumed_from.type` note now distributes the reasoning the way you argued it: resolve-to-interpret attached to carrying the type at all, flatness justified by the authoring guidance (flat attributes, backends assumed not to index structured-value properties) with `gen_ai.tool.type` alongside `gen_ai.tool.call.id` cited as the in-registry precedent, and grouping by boundary type as the query consequence. The MUST-record-together sentence is byte-identical.
- The checkpoint-to-pause join: `gen_ai.agent.pause.id` is now on `gen_ai.agent.checkpointed`, recommended when the checkpoint is taken at a pause boundary, with the note stating what it buys: a consumer reaching the checkpoint through `resumed_from` can also reach the resolution that authorized continuing from it, without reconstructing record order. The scenario emits it, since LangGraph checkpointing at an interrupt is exactly a pause-boundary checkpoint, so the record graph you described is now walkable end to end in the committed data.json.
- README now states that only `resumed_from.type=checkpoint` is exercised here and that the `pause` member is defined but not demonstrated by this scenario.

On liveness: agreed, and that framing (separating an expiry in flight from a dead producer needs a lease or heartbeat, a different signal) is the strongest defense of the consumer-policy sentence so far; it is what makes it a stopping point rather than a placeholder.

Thanks for the close read.

StevenMih:
Read 5426dfb and both points read as faithfully applied, and the checkpoint-to-pause join makes the graph walkable in exactly the way the thread wanted. The README's honesty about which resumed_from.type member is exercised looks good. From my side this is ready for further review.

reference/scenarios/agent-lifecycle/README.md
```diff
@@ -0,0 +1,67 @@
...
+reference litmus tests only framework-owned operations are emitted.
+
+## Capture notes
+
+Declared per the evaluation rubric, so review can weigh them explicitly:
+
+- **Resolution semantics are gate-level, not framework-level, in LangGraph.**
+  `interrupt()` payloads and `Command(resume=...)` values are
+  application-defined, so generic LangGraph instrumentation cannot know that
+  a given resume value means approved or refused, nor that a payload field
```
lmolkova: These semantics are app-defined, not LangGraph-defined. This does not demonstrate generic library capture for `gen_ai.agent.pause.reason` and `gen_ai.agent.pause.resolution`. Please add a framework-owned reference or narrow scope.

model/gen-ai/registry.yaml
```diff
@@ -1025,3 +1025,134 @@ attributes:
...
+  - key: gen_ai.agent.checkpoint.id
+    type: string
+    brief: Identifier of a persisted agent state checkpoint.
+    note: |
+      Used for durable-recovery correlation. When resumed work continues from a
+      checkpoint, this is the value carried in `gen_ai.agent.resumed_from.id`
+      with `gen_ai.agent.resumed_from.type` set to `checkpoint`.
+    examples: ["ck_001", "chk_8f3a2d"]
+    stability: development
+  - key: gen_ai.agent.resumed_from.type
```
lmolkova: `gen_ai.agent.resumed_from.type` adds `pause`, but this PR’s reference only exercises `checkpoint`. Please add a reference that emits `pause`, or defer that enum member until it is demonstrated.

reference/scenarios/agent-lifecycle/scenario.py
```diff
@@ -0,0 +1,253 @@
...
+    )
+
+
+def _invoke_agent_span():
+    span = _reference_tracer.start_as_current_span(f"invoke_agent {AGENT_NAME}", kind=SpanKind.INTERNAL)
+    return span
+
+
+def _pause_until_interrupt(graph, config, action: str):
+    """Run the graph to its interrupt and emit paused + checkpointed.
```
lmolkova: This does not demonstrate library-owned capture: LangGraph only transports the scenario's `reason`, `deadline`, and `approved` payloads, while `_emit` creates the lifecycle events outside a LangGraph lifecycle API. The reference rules require each value to come from a library-defined input, output, state, or error. Please drop this scenario as coverage for these conventions or use an API that owns these semantics.

lmolkova (changes requested):
there is no evidence that this data is collectable from auto-instrumentations. The scenario is just a demo app that emits some telemetry. Please read https://github.com/open-telemetry/semantic-conventions-genai/blob/main/.github/skills/reference/SKILL.md for more info on what counts as a reference scenario

meshailabs:
Thank you for the pointer to SKILL.md, it made the problem obvious: the old scenario handed its own payload to LangGraph and read it back, so it demonstrated nothing the library owns. It is deleted, and the conventions are now covered by three libraries that do own the semantics: LangGraph's `interrupt()` and checkpointer for the durable half, Agent Framework's `function_approval_request` and the `approved` field of its response for the decision half, and an OpenAI Assistants run held in `requires_action` for an external-system pause. `gen_ai.agent.pause.deadline` is the one field left uncovered: `Run.expires_at` is the source, but the pinned mock returns null for it, so it is recorded as `add reference for supporting library` rather than claimed as demonstrated.

reference/scenarios/agent-framework/scenario.py
```diff
@@ -177,12 +179,92 @@ async def run_agent_workflow():
...
+            "gen_ai.agent.name": agent.name,
+        },
+    )
+
+    # The decision travels back in the framework's own response content, whose
+    # approved field carries the outcome.
+    decision = approval_request.to_function_approval_response(approved=True)
+    await agent.run(Message(role="user", contents=[decision]), session=session)
+
+    reference_event_logger(_LIFECYCLE_LOGGER).emit(
```
lmolkova: These lifecycle events are emitted by scenario code after `agent.run`; Agent Framework does not emit them or expose an instrumentation hook here. This does not demonstrate library-owned capture under the reference rules. Please drop this as coverage or instrument a library API that owns pause/resume semantics.

reference/scenarios/openai-assistants/scenario.py
```diff
@@ -138,6 +140,25 @@ def get_weather(location):
...
+                # both come from the run object. The run also names what it
+                # waits for, and submitting tool outputs is the caller's job
+                # rather than a person's.
+                pause_reason = "external_system" if required_action.type == "submit_tool_outputs" else "human_input"
+                reference_event_logger(_LIFECYCLE_LOGGER).emit(
+                    event_name="gen_ai.agent.paused",
+                    body="paused",
+                    attributes={
+                        "gen_ai.agent.execution.id": run.id,
+                        "gen_ai.agent.pause.id": tool_calls[0].id,
```
lmolkova: `tool_calls[0].id` identifies the first tool call, not the suspension boundary. A run can pause for multiple tool calls, so this is not a stable `pause.id` for the pause event. Please omit this attribute unless the library exposes a pause/request id.

lmolkova (changes requested):

meshailabs:
Both dropped. The agent-framework scenario is a native-telemetry scenario, so events emitted around `agent.run` were the wrong shape for it, and you are right that the Assistants API gives no id for the suspension boundary itself, only for the calls it is waiting on.

That leaves `langchain` as the coverage, where LangGraph owns the pause, the checkpoint, and the ids for both. `gen_ai.agent.pause.reason`, `.deadline`, `.resolution` and the `gen_ai.agent.pause.resolved` event now have nothing emitting them. They are listed in the description as `add reference for supporting library`, with the current-call source each would need, rather than left implicit. If you would rather they not ship uncovered, say so and I will move them to a follow-up.

reference/scenarios/agent-lifecycle/scenario.py
meshailabs: Deleted. The values came from the scenario's own payload, so it demonstrated nothing the library owns, and the pointer to SKILL.md made that clear.

Coverage now lives in `reference/scenarios/langchain` on LangGraph's `interrupt()` with a checkpointer: the execution id from the thread id in `aget_state().config`, the pause id from the returned `Interrupt`, and the checkpoint id from the snapshot, emitted inline at the instrumentation site.

reference/scenarios/agent-framework/scenario.py
meshailabs: Dropped, and the file is back to its upstream state. Agent Framework's scenario is a native-telemetry scenario, so lifecycle events emitted around `agent.run` would not reflect what the framework itself produces. Thank you for catching it.

reference/scenarios/openai-assistants/scenario.py
meshailabs: Dropped, and the file is back to its upstream state. `required_action` exposes no id for the suspension itself and a run can wait on several tool calls, so there is no stable pause id to read here. The PR description records it under capture gaps as `add reference for supporting library`.

reference/scenarios/langchain/scenario.py
```diff
@@ -260,6 +266,85 @@ async def run_workflow_reference():
...
+    # First invocation runs the agent node and suspends at the gate. LangGraph
+    # reports the interrupts it raised under the __interrupt__ key.
+    suspended = await graph.ainvoke({"messages": [input_text]}, config=config)
+    (pause,) = suspended["__interrupt__"]
+
+    # The checkpointer persisted the suspended state. Both ids come from the
+    # snapshot LangGraph returns, not from the config passed in.
+    snapshot = await graph.aget_state(config)
+    execution_id = snapshot.config["configurable"]["thread_id"]
+    checkpoint_id = snapshot.config["configurable"]["checkpoint_id"]
```
lmolkova: `aget_state` is an application-facing read API - no instrumentation calls it, so reaching for it here hides where the data would actually come from. Generic instrumentation has two interception points and the scenario should emit from them: a wrapper around `Pregel.ainvoke`/`astream`, whose return value carries `__interrupt__` with `Interrupt.id`, and a wrapper around the checkpointer object passed to `compile()`, which LangGraph drives itself.

Emitting from the driver function leaves rule 6 unchecked - as written the scenario shows the events are emittable, not that they are instrumentable.

reference/scenarios/langchain/scenario.py
```diff
@@ -260,6 +266,85 @@ async def run_workflow_reference():
...
+
+    # First invocation runs the agent node and suspends at the gate. LangGraph
+    # reports the interrupts it raised under the __interrupt__ key.
+    suspended = await graph.ainvoke({"messages": [input_text]}, config=config)
+    (pause,) = suspended["__interrupt__"]
+
+    # The checkpointer persisted the suspended state. Both ids come from the
+    # snapshot LangGraph returns, not from the config passed in.
+    snapshot = await graph.aget_state(config)
+    execution_id = snapshot.config["configurable"]["thread_id"]
```
lmolkova: `thread_id` is a conversation thread reused across many runs, which contradicts `gen_ai.agent.execution.id`'s own definition of one end-to-end execution. `CheckpointMetadata.run_id` is per-run but changes on the resume invocation, so it doesn't span the suspend boundary either.

LangGraph has no id matching these semantics. CrewAI's `flow_id` and ADK's `invocation_id` do - either would make a better reference. If this scenario stays, declare the mismatch as a capture gap rather than aliasing `thread_id` silently.

reference/scenarios/langchain/scenario.py
```diff
@@ -260,6 +266,85 @@ async def run_workflow_reference():
...
+    from langgraph.graph import END, START, StateGraph
+    from langgraph.types import Command
+
+    builder = StateGraph(GraphState)
+    builder.add_node("agent", agent_node)
+    builder.add_node("approval_gate", approval_gate)
+    builder.add_edge(START, "agent")
+    builder.add_edge("agent", "approval_gate")
+    builder.add_edge("approval_gate", END)
+    graph = builder.compile(checkpointer=InMemorySaver())
```
lmolkova: This is where checkpoint instrumentation would attach: a `BaseCheckpointSaver` subclass wrapping `InMemorySaver` and emitting from `aput`, which LangGraph calls on its own and which hands back the `RunnableConfig` holding `checkpoint_id` plus `CheckpointMetadata` (`source`, `step`, `run_id`).

Writing it that way surfaces what the current hand-picked emit hides: `aput` fires at every superstep, not once at a pause, so real instrumentation emits many of these per run. The convention then has to decide - scope the event to suspension boundaries, or state that it is per-step. It also shows the join the event asks for: `pause.id` is not in scope inside `aput`, so the instrumentation must carry the pending interrupt across two hooks.

reference/scenarios/langchain/scenario.py
```diff
@@ -260,6 +266,85 @@ async def run_workflow_reference():
...
+
+    reference_event_logger(_LIFECYCLE_LOGGER).emit(
+        event_name="gen_ai.agent.paused",
+        body="paused",
+        attributes={
+            "gen_ai.agent.execution.id": execution_id,
+            "gen_ai.agent.pause.id": pause.id,
+            "gen_ai.agent.name": AGENT_NAME,
+        },
+    )
```
lmolkova: Suspension is detectable generically - `ainvoke` returned without reaching END and `__interrupt__` is present - so show the emit from that wrapper rather than from the driver.

Note also that the `interrupt()` payload is app-defined, so a generic instrumentation cannot classify `pause.reason` here, which is why neither it nor `pause.deadline` is recorded.

reference/scenarios/langchain/scenario.py
```diff
@@ -260,6 +266,85 @@ async def run_workflow_reference():
...
+        attributes={
+            "gen_ai.agent.execution.id": execution_id,
+            "gen_ai.agent.checkpoint.id": checkpoint_id,
+            "gen_ai.agent.pause.id": pause.id,
+            "gen_ai.agent.name": AGENT_NAME,
+        },
+    )
+
+    # A later invocation continues the run from the persisted checkpoint.
+    resumed = await graph.ainvoke(Command(resume="approved"), config=config)
```
lmolkova: `interrupt()` takes `Any` and `Command(resume=...)` is `Any`, so `"approved"` is a string this scenario invented - a wrapper can't derive anything from it, and the scenario as written proves the opposite of what it needs to.

Use LangGraph's typed HITL shape instead: `HumanInterrupt` in the interrupt payload and `HumanResponse` (`type: accept|ignore|response|edit`) as the resume value, both from `langgraph.prebuilt.interrupt`. That is a shape generic instrumentation can duck-type, and it makes `pause.reason: human_input` and `gen_ai.agent.pause.resolved` derivable rather than hand-written.

reference/scenarios/langchain/scenario.py
```diff
@@ -260,6 +266,85 @@ async def run_workflow_reference():
...
+    reference_event_logger(_LIFECYCLE_LOGGER).emit(
+        event_name="gen_ai.agent.resumed",
+        body="resumed",
+        attributes={
+            "gen_ai.agent.execution.id": resumed_snapshot.config["configurable"]["thread_id"],
+            "gen_ai.agent.resumed_from.type": "checkpoint",
+            "gen_ai.agent.resumed_from.id": checkpoint_id,
+            "gen_ai.agent.name": AGENT_NAME,
+        },
+    )
```
lmolkova: Two issues. This fires after `ainvoke` returned - after the segment it announces already ended - where consumers need it at segment start. And it is emitted with no ambient span, unlike `run_workflow_reference` which wraps its graph run in `invoke_workflow`, so the events land with no trace correlation. Emit inside the span for the resumed segment.

model/gen-ai/events.yaml
```diff
@@ -60,3 +60,133 @@ events:
...
+    stability: development
+    attributes:
+      - ref: gen_ai.agent.execution.id
+        requirement_level: required
+      - ref: gen_ai.agent.pause.reason
+        requirement_level: recommended
+      - ref: gen_ai.agent.pause.id
+        requirement_level: required
+      - ref: gen_ai.agent.pause.deadline
+        requirement_level:
```
lmolkova: Neither attribute is emitted by any scenario. `pause.reason` becomes derivable via the `HumanInterrupt` shape (see the scenario comment) and ADK distinguishes an auth or tool-confirmation pause from a long-running tool call, so `human_input` and `external_system` have producers - please cover it in a scenario.

`pause.deadline` has no source in any of the four durable-execution libraries. Please drop it from this PR until a scenario shows where a deadline is observable.

model/gen-ai/events.yaml
```diff
@@ -60,3 +60,133 @@ events:
...
+      - ref: gen_ai.agent.execution.id
+        requirement_level: required
+      - ref: gen_ai.agent.pause.id
+        requirement_level: required
+      - ref: gen_ai.agent.pause.resolution
+        requirement_level: required
+      - ref: gen_ai.agent.id
+        requirement_level: recommended
+      - ref: gen_ai.agent.name
+        requirement_level: recommended
```
lmolkova: No scenario emits this event or `gen_ai.agent.pause.resolution`, and the LangGraph resume path can't produce it - the resume payload is app-defined.

If the scenario moves to LangGraph's `HumanResponse` shape it becomes derivable, but the enum doesn't fit it: `accept` -> `approved` and `ignore` -> `refused` map cleanly, while `edit` and `response` are continuations with modified input and have no member. Either extend the enum or state how those map. If the scenario isn't reworked, please drop the event and the attribute from this PR (rule 5).

model/gen-ai/registry.yaml
```diff
@@ -1025,3 +1025,134 @@ attributes:
...
+      `gen_ai.agent.resumed_from.id`, rather than as a structured value, per
+      the authoring guidance to represent complex values as flat attributes
+      and to assume backends do not index their individual properties;
+      `gen_ai.tool.type` alongside `gen_ai.tool.call.id` is the existing
+      precedent, and grouping resumptions by boundary type stays an ordinary
+      query. Instrumentations MUST record
+      `gen_ai.agent.resumed_from.type` and `gen_ai.agent.resumed_from.id`
+      together or not at all.
+    examples: ["checkpoint", "pause"]
+    stability: development
```
lmolkova: This MUST-together pair cannot be filled in ADK, which supports durable execution but has no checkpoint identity - `run_async(invocation_id=...)` resumes an invocation from the last event in its session log. An ADK instrumentation can say a segment resumed and can name the execution, but has nothing to put in `resumed_from.id`. Either add a member for a resume with no boundary id, or make the pair optional for that case.

Separately, both CrewAI and Agent Framework model checkpoints as a tree - `parent_checkpoint_id` / `branch` / `parent_branch` in CrewAI's fork events, `previous_checkpoint_id` on `WorkflowCheckpoint`. A flat `resumed_from` cannot express that a resume forked a branch rather than continuing the trunk. Worth deciding whether lineage is in scope now or explicitly out.

model/gen-ai/registry.yaml
```diff
@@ -1025,3 +1025,134 @@ attributes:
...
+          value: "rate_limit"
+          brief: Paused due to provider or tool rate limiting.
+          stability: development
+        - id: scheduled_delay
+          value: "scheduled_delay"
+          brief: Paused until a scheduled time or deferred resumption.
+          stability: development
+    brief: The reason agent execution paused.
+    examples: ["human_input", "scheduled_delay"]
+    stability: development
```
lmolkova: `rate_limit` and `scheduled_delay` have no producer in any durable-execution library in this repo - LangGraph, CrewAI, ADK and Agent Framework all surface pauses as human-input or external-callback waits. Please drop both members until something can emit them.

model/gen-ai/registry.yaml
```diff
@@ -1025,3 +1025,134 @@ attributes:
...
+          brief: The pause was resolved affirmatively; execution is permitted to continue.
+          stability: development
+        - id: refused
+          value: "refused"
+          brief: A human or policy declined; execution ends at this boundary.
+          stability: development
+        - id: expired
+          value: "expired"
+          brief: A configured deadline passed with no decision; execution ends at this boundary.
+          stability: development
```
lmolkova: This enum fits none of the four libraries exactly: CrewAI's outcomes are an app-defined `emit` list of arbitrary strings, LangGraph's prebuilt shape is `accept|ignore|edit|response`, ADK's `ToolConfirmation.confirmed` is a bool, and Agent Framework takes an arbitrary response payload keyed by request id. Only approved/refused survive a cross-library mapping.

`expired` has no producer at all - no library in this repo emits a deadline-passed record, which matches `pause.deadline` having no source either. Please drop `expired` along with the deadline attribute.

reference/scenarios/langchain/scenario.py
lmolkova: `aget_state` is an application-facing read API - no instrumentation calls it, so reaching for it here hides where the data would actually come from. Generic instrumentation has two interception points and the scenario should emit from them: a wrapper around `Pregel.ainvoke`/`astream`, whose return value carries `__interrupt__` with `Interrupt.id`, and a wrapper around the checkpointer object passed to `compile()`, which LangGraph drives itself.

Emitting from the driver function leaves rule 6 unchecked - as written the scenario shows the events are emittable, not that they are instrumentable.

reference/scenarios/langchain/scenario.py
lmolkova: `thread_id` is a conversation thread reused across many runs, which contradicts `gen_ai.agent.execution.id`'s own definition of one end-to-end execution. `CheckpointMetadata.run_id` is per-run but changes on the resume invocation, so it doesn't span the suspend boundary either.

LangGraph has no id matching these semantics. CrewAI's `flow_id` and ADK's `invocation_id` do - either would make a better reference. If this scenario stays, declare the mismatch as a capture gap rather than aliasing `thread_id` silently.

reference/scenarios/langchain/scenario.py
lmolkova: This is where checkpoint instrumentation would attach: a `BaseCheckpointSaver` subclass wrapping `InMemorySaver` and emitting from `aput`, which LangGraph calls on its own and which hands back the `RunnableConfig` holding `checkpoint_id` plus `CheckpointMetadata` (`source`, `step`, `run_id`).

Writing it that way surfaces what the current hand-picked emit hides: `aput` fires at every superstep, not once at a pause, so real instrumentation emits many of these per run. The convention then has to decide - scope the event to suspension boundaries, or state that it is per-step. It also shows the join the event asks for: `pause.id` is not in scope inside `aput`, so the instrumentation must carry the pending interrupt across two hooks.

reference/scenarios/langchain/scenario.py
lmolkova: Suspension is detectable generically - `ainvoke` returned without reaching END and `__interrupt__` is present - so show the emit from that wrapper rather than from the driver.

Note also that the `interrupt()` payload is app-defined, so a generic instrumentation cannot classify `pause.reason` here.

reference/scenarios/langchain/scenario.py
lmolkova: `interrupt()` takes `Any` and `Command(resume=...)` is `Any`, so `"approved"` is a string this scenario invented - a wrapper can't derive anything from it, and the scenario as written proves the opposite of what it needs to.

Use LangGraph's typed HITL shape instead: `HumanInterrupt` in the interrupt payload and `HumanResponse` (`type: accept|ignore|response|edit`) as the resume value, both from `langgraph.prebuilt.interrupt`. That is a shape generic instrumentation can duck-type, and it makes `pause.reason: human_input` and `gen_ai.agent.pause.resolved` derivable rather than hand-written. CrewAI and the OpenAI Agents SDK are cleaner still - `HumanFeedbackReceivedEvent.outcome` against its `request_id`, and `RunState.approve()` / `reject()` over `get_interruptions()`, are unambiguous resolutions.

reference/scenarios/langchain/scenario.py
lmolkova: Two issues. This fires after `ainvoke` returned - after the segment it announces already ended - where consumers need it at segment start. And it is emitted with no ambient span, unlike `run_workflow_reference` which wraps its graph run in `invoke_workflow`, so the events land with no trace correlation. Emit inside the span for the resumed segment.

model/gen-ai/events.yaml
lmolkova: Neither attribute is emitted by any scenario in this PR. `pause.reason` is at least derivable elsewhere - LangGraph's `HumanInterrupt` shape carries it, and ADK distinguishes an auth or tool-confirmation pause from a long-running tool call, so `human_input` and `external_system` have producers. Please cover it in a scenario.

`pause.deadline` has no producer in any framework in this repo: none of CrewAI, Agent Framework, ADK, the OpenAI Agents SDK or LangGraph attaches a deadline to a suspension boundary. Please drop it from this PR and bring it back with a scenario that shows one (rule 5).

model/gen-ai/events.yaml
lmolkova: No scenario emits this event or `gen_ai.agent.pause.resolution`, and the LangGraph resume path can't produce it - the resume payload is app-defined.

The concept is well-founded: CrewAI carries `HumanFeedbackReceivedEvent.outcome` against the originating `request_id`, the OpenAI Agents SDK has `RunState.approve()` / `reject()`, ADK has `ToolConfirmation.confirmed`. So the fix is a scenario in a library that can express the outcome, not a rewrite of the event. As it stands please drop the event and the attribute from this PR and land them with that scenario (rule 5).

model/gen-ai/registry.yaml
lmolkova: This MUST-together pair cannot be filled by ADK, which supports resumable invocations but has no checkpoint identity - `run_async(invocation_id=...)` resumes from the last event in the session log. An ADK instrumentation can say a segment resumed and can name the execution, but has nothing to put in `resumed_from.id`. Either add a member for a resume with no boundary id, or make the pair optional in that case.

Separately, both CrewAI and Agent Framework model checkpoints as a tree - `parent_checkpoint_id` / `branch` / `parent_branch` in CrewAI's fork events, `previous_checkpoint_id` on `WorkflowCheckpoint`. A flat `resumed_from` cannot express that a resume forked a branch rather than continuing the trunk. Worth deciding whether lineage is in scope now or explicitly out.

model/gen-ai/registry.yaml
lmolkova: `rate_limit` and `scheduled_delay` have no producer in any framework in this repo - CrewAI, Agent Framework, ADK, the OpenAI Agents SDK and LangGraph all surface pauses as human-input or external-callback waits. Rate limiting appears as internal retry with backoff, not as a suspension boundary with an id. Please drop both members until something can emit them.

model/gen-ai/registry.yaml
lmolkova: `approved` and `refused` are well-grounded - they map onto `RunState.approve()` / `reject()` in the OpenAI Agents SDK and onto `ToolConfirmation.confirmed` in ADK. The problem is that this PR's scenario can't produce either, so the vocabulary ships unproven; see the events.yaml comment.

`expired` has no producer in any framework here, which matches `pause.deadline` having no source - no library in this repo records a deadline against a suspension boundary or emits a deadline-passed record. Please drop `expired` along with the deadline attribute.

One mapping to settle for the two that stay: LangGraph's prebuilt `HumanResponse` also has `edit` and `response`, continuations with modified input that are neither a plain approval nor a refusal.

reference/scenarios/langchain/scenario.py
lmolkova: `aget_state` is an application-facing read API - no instrumentation calls it. Everything here is emitted from the driver function with ids fetched on purpose for the emit, so the scenario shows these events are emittable, not that generic instrumentation can capture them (rule 6).

The interception points a real instrumentation would have are a wrapper around `Pregel.ainvoke`/`astream` and a wrapper around the checkpointer passed to `compile()`. Building it that way is what would surface whether the proposed shape holds.

reference/scenarios/langchain/scenario.py
lmolkova: `thread_id` is a conversation thread reused across many runs, which contradicts `gen_ai.agent.execution.id`'s own definition of one end-to-end execution. `CheckpointMetadata.run_id` is per-run but changes on the resume invocation, so it doesn't span the suspend boundary either. LangGraph has no id with these semantics, while CrewAI's `flow_id` and ADK's `invocation_id` do.

The attribute may well be right; this library just cannot demonstrate it.

reference/scenarios/langchain/scenario.py
lmolkova: Checkpoint instrumentation would attach here, by wrapping the checkpointer. That matters because `aput` fires every superstep, not once at a pause, so `gen_ai.agent.checkpointed` is a per-step record in LangGraph and a per-event one in CrewAI, whose `CheckpointConfig.on_events` accepts `["*"]`. The hand-picked emit below hides that volume question entirely.

reference/scenarios/langchain/scenario.py
lmolkova: `interrupt()` takes `Any` and `Command(resume=...)` is `Any`, so `"approved"` is a string this scenario invented. Nothing generic can read an outcome out of it.

This is the clearest case of the scenario not backing the vocabulary: the resolution attributes are demonstrated by an application choosing a string that happens to match an enum member.

model/gen-ai/events.yaml
lmolkova: Neither attribute is emitted by any scenario here. `pause.reason` has producers elsewhere - ADK distinguishes an auth or tool-confirmation pause from a long-running tool call. `pause.deadline` has none: no library in this repo attaches a deadline to a suspension boundary.

Please drop `pause.deadline` from this PR (rule 5).

model/gen-ai/events.yaml
lmolkova: No scenario emits this event or `gen_ai.agent.pause.resolution`, and LangGraph cannot produce either. The concept is real - CrewAI carries `HumanFeedbackReceivedEvent.outcome` against its `request_id`, the OpenAI Agents SDK has `RunState.approve()` / `reject()`, ADK has `ToolConfirmation.confirmed` - but nothing in this PR shows it.

Please drop the event and the attribute here and land them with a scenario or instrumentation that produces them (rule 5).

model/gen-ai/registry.yaml
lmolkova, line 1110: This MUST-together pair can only be filled by libraries with identified checkpoints. ADK resumes an invocation from the last event in its session log and has nothing to put in `resumed_from.id`; Pydantic AI's durability lives in Temporal, DBOS or Prefect, which own their own ids.

And where checkpoints do have ids they form a tree - `parent_id` / `branch` / `parent_branch` in CrewAI, `previous_checkpoint_id` on `WorkflowCheckpoint`, `fork_session` in the Claude Agent SDK. A flat pointer cannot say which branch a run continued on. This is the part I would most want settled by prototypes before it lands.

model/gen-ai/registry.yaml
lmolkova: `rate_limit` and `scheduled_delay` have no producer in any library in this repo - pauses everywhere are human-input or external-callback waits, and rate limiting appears as internal retry with backoff rather than a suspension boundary with an id. Please drop both members.

model/gen-ai/registry.yaml
lmolkova: `approved` and `refused` do map onto `RunState.approve()` / `reject()` in the OpenAI Agents SDK and `ToolConfirmation.confirmed` in ADK, though CrewAI's outcomes are an app-defined list of strings and Pydantic AI carries approvals per deferred tool call - the shape varies more than three members suggest.

`expired` has no producer at all, matching `pause.deadline` having no source. Please drop it along with the deadline attribute.

lmolkova (changes requested):

lmolkova (changes requested):
Reading this end to end, the goal looks like visibility into **durable execution**: runs that suspend, persist state, and continue later, possibly in another process. That is worth having, but it is a much bigger design space than one scenario suggests, and this repo already carries several libraries that do it in materially different ways.

A quick sketch of what they actually offer:

| library | how state survives | what suspends a run | ids you get |
|---|---|---|---|
| LangGraph | `BaseCheckpointSaver.aput()` every superstep | `interrupt()` in a node | `thread_id`, `checkpoint_id`, `Interrupt.id` |
| CrewAI | two mechanisms: `@persist` calls `FlowPersistence.save_state(flow_uuid, method_name, state)` per method; `CheckpointConfig(on_events=...)` writes identified checkpoints | `@human_feedback` method raises `HumanFeedbackPending` | `flow_id`, `request_id`, `checkpoint_id`, `parent_id`, `branch` |
| MS Agent Framework | `CheckpointStorage` holding `WorkflowCheckpoint` | run reaches `IDLE_WITH_PENDING_REQUESTS` | `checkpoint_id`, `previous_checkpoint_id`, request id from `get_request_info_events()` |
| Google ADK | session event log plus `EventActions.agent_state`, enabled by `ResumabilityConfig(is_resumable=True)` | long-running tool call, tool confirmation, auth request | `invocation_id`, `session_id`, `Event.id`, `long_running_tool_ids`. No checkpoint id |
| OpenAI Agents SDK | `RunState.to_string()` / `from_string()` | tool approval interruption | `ToolApprovalItem.call_id` |
| Pydantic AI | delegated to `durable_exec.temporal`, `.dbos` or `.prefect` | `DeferredToolRequests` (`calls`, `approvals`) | tool call id, plus the engine's own workflow and run ids |
| Claude Agent SDK | `ClaudeAgentOptions.session_store` | nothing suspends; a past session is resumed or forked | `session_id`, `resume`, `fork_session` |

To be clear about what that is: a quick AI sketch from reading the installed packages, not real research. It is enough to show the problem is bigger than the current shape assumes:

- Three different durability models coexist: identified checkpoints, an append-only event log with nothing to point at, and durability handed to an external workflow engine that owns its own ids. `resumed_from.type`/`.id` can only be filled by the first.
- Checkpoints form a tree in CrewAI and Agent Framework, and sessions fork in the Claude Agent SDK. A flat resume pointer cannot say which branch a run continued on.
- One library can have two durability mechanisms at once with different identity models, as CrewAI does.
- Checkpoint writes range from once, to per superstep, to per event, which decides whether a checkpoint record is rare or a firehose.
- The outcome of a human decision is a bool in one library, approve/reject in another, and an app-defined list of strings in a third.

Finding a shape that works across this needs human judgement about what to model now and what to leave out, and a review that does it justice will take a while. It would be great if you could bring this to the GenAI SIG call on Tuesday and present the research behind it - the schedule and calendar link are in the [community repo](https://github.com/open-telemetry/community#sig-genai-instrumentation).

I would also suggest breaking this down and starting from prototypes of real instrumentation rather than scenarios here. instrumentation over two or three of these libraries in [opentelemetry-python-genai](https://github.com/open-telemetry/opentelemetry-python-genai) producing coherent telemetry would help design it much better than scenarios here and would serve as a much better evidence.

The inline comments above are the concrete instances of this, not a request to polish the current scenario.

meshailabs:
Apologies, I cannot make Tuesday's SIG call. Rather than hold this thread until I can attend one, here is the response in writing, and I am happy to discuss async or on a future call.

**Accepting all five drops.** `gen_ai.agent.pause.deadline`, the `gen_ai.agent.pause.resolved` event with `pause.resolution`, and the `rate_limit`, `scheduled_delay`, `expired` members all come out. Each was already listed in the description as a capture gap, and rule 5 is the right bar: nothing in this repo produces them.

**Accepting the bigger point.** You are right that this PR models one durability pattern and calls it the design space. I am converting it to draft and taking the prototype path: instrumentation over three libraries chosen to span the three durability models your sketch identifies, LangGraph (identified checkpoints), CrewAI (dual mechanism, tree, app-defined outcomes), and ADK (append-only event log, nothing to point at). What holds across all three comes back here as evidence; what does not gets dropped or narrowed.

**On the research.** I verified your sketch against the library sources rather than presenting my own:

- Every API you name checks out at source: `HumanFeedbackPending`, `CheckpointConfig`, `ResumabilityConfig`, `ToolConfirmation`, `ToolApprovalItem`, `DeferredToolRequests`, `fork_session`, `WorkflowCheckpoint`.
- `interrupt(value: Any) -> Any` and `Command.resume` as `Any` confirmed in `langgraph/types.py`. Your rule 6 point stands: my scenario invented "approved", and `aget_state` is an application read, not an interception point.
- Per-superstep checkpointing is confirmed literally in a source comment: `_put_checkpoint is called once per superstep` in `pregel/_loop.py`. The volume question is real.
- One addition that strengthens your branch point: LangGraph checkpoints also form a tree. `CheckpointMetadata` carries `parents` (namespace to checkpoint id) and a `"fork"` source, so the flat `resumed_from` pointer fails in four libraries, not three.
- On `execution.id`: conceded that LangGraph cannot demonstrate it. `thread_id` is a conversation, `CheckpointMetadata.run_id` does not span the suspend boundary. CrewAI's `flow_id` and ADK's `invocation_id` are the demonstrating sources, so the attribute rides on the prototypes too.

I will bring the first prototype results back to this thread.

meshailabs:
The five drops are applied (afd68be, prose swept in 04db0ec): `pause.deadline`, `pause.resolution`, the `pause.resolved` event, and the `rate_limit`, `scheduled_delay`, `expired` members are gone. I said they would come out on 08-31 and only did it now, sorry for the lag.

Phase 1 of the prototypes, LangGraph first of three, is real instrumentation on a branch of the released `opentelemetry-instrumentation-genai-langchain` package: https://github.com/meshailabs/opentelemetry-python-genai/tree/feat/langgraph-lifecycle-events (design note and captured run under `examples/langgraph-lifecycle/`).

Two of my 08-31 claims were wrong. LangGraph 1.2 ships a public lifecycle API, `langgraph.callbacks.GraphCallbackHandler.on_interrupt` / `on_resume`, whose events carry the real `Interrupt` objects, the checkpoint id and the run id, and it already dispatches them to the released OTel handler, which currently raises `AttributeError` on every interrupt and resume. `on_chain_end` never sees `__interrupt__`, and `BaseCheckpointSaver.put` is abstract, so the saver instance passed to `compile()` is patched instead.

What the telemetry shows: per superstep is confirmed, 3 `checkpointed` on the pausing invoke and 2 on the resume; `paused` carries the real `Interrupt.id`; `resumed` carries `resumed_from = checkpoint` with the real checkpoint id.

Not demonstrable from LangGraph, so not emitted: `execution.id` (no id spans suspend and resume; the two invokes are unlinked traces), `pause.reason` (opaque payload), and the `pause` member of `resumed_from.type`. None of the three has a producer anywhere yet; they ride on CrewAI and ADK or they come out too.

Two things the flat model cannot express: nested subgraphs emit one `resumed` per level, and without a checkpointer `paused` still fires with a checkpoint id nothing persisted.

The hand-emitted langchain scenario is out of this PR; it could not produce `execution.id`. It comes back once it can run the instrumentation as its emitter.

CrewAI next.

socksninja:
I’m building SABLE, an independent reliability-evidence layer for real agent runs. I noticed this PR is moving toward lifecycle events for async/long-running executions, which is very close to a problem I’m testing from the evidence side: how do we preserve provenance strongly enough that an external evaluator can verify what actually happened across runtime boundaries?

I’m not looking to add another competing schema. I’d value your view on one network question: **who is the next runtime/observability maintainer you think I should talk to about validating this across a real agent execution?** If someone comes to mind, would you be willing to point me to them?

socksninja:
I’m testing an adjacent evidence question from SABLE: once agent runtime semantics are standardized, can a real execution be converted into a machine-readable Reliability Record whose provenance and externally observable state can be independently verified?

@StevenMih your review on this PR is exactly the kind of runtime-vs-semantics perspective I’m looking for. I’m not asking for adoption or endorsement. **Who is the single next person you’d point me to** — a runtime maintainer, observability/evidence engineer, or standards contributor — who could stress-test this on a real durable agent execution?

PR #447, opened 2026-08-10
Unified OpenTelemetry model for tracking agent-to-agent interactions

nikhilNava:
## Description

This PR defines two refinements for agent interactions without introducing new base span types:

| Runtime operation | Span refinement | Added context |
| --- | --- | --- |
| A tool transfers work or control | `gen_ai.execute_tool.transfer.internal` | `gen_ai.transfer.mode` and `gen_ai.transfer.target.*` |
| An API or protocol invokes a remote agent and exposes its logical caller | `gen_ai.invoke_agent.caller.client` | `gen_ai.caller.type` and `gen_ai.caller.name` |

On a tool transfer, `gen_ai.agent.*` continues to identify the agent executing the tool. The target is recorded separately, and `gen_ai.transfer.mode` distinguishes operations that return to the caller from those that pass control.

On a caller-aware remote invocation, `gen_ai.agent.*` identifies the invoked agent and `gen_ai.caller.*` identifies the immediate logical caller. Caller and transfer attributes are recorded only when the library or protocol exposes them. Instrumentation must not infer them from span hierarchy, names, timing, or application conventions.

## Motivation

During incident triage, operators need to identify who initiated an agent interaction, where work was sent, and whether the original agent resumes. Existing frameworks expose these facts through different APIs, including OpenAI Agents tools and handoffs, Google ADK agent tools and A2A agents, and LangGraph `Command` transfers.

The refinements preserve the meaning of existing attributes and spans. In particular, tool metrics remain attributed to the agent that executed the tool rather than to the transfer target.

## Prototype

The reference scenarios derive the new values from library-owned API inputs, results, or state:

- Google ADK and OpenAI Agents cover tool transfers that return to the caller.
- Google ADK and LangGraph cover tool transfers that pass control.
- Google ADK `RemoteA2aAgent` performs an A2A request and records the workflow exposed through ADK state as the logical caller of the remote invocation.

The generated reference reports track coverage separately for each refinement. A dedicated in-process transfer that exposes no caller-owned tool or protocol operation does not produce an inferred transfer span; observable source and target executions may each produce their own `invoke_agent` INTERNAL span.

## Checklist

See [CONTRIBUTING.md]


nikhilNava:
Pushed `7f32bac` — two follow-up fixes to the openai-agents reference scenario
(no model or docs change):

1. **The native-handoff target execution started its own trace.**
   `invoke_agent billing-agent` was started from whatever context was current
   between SDK turns — empty once the previous execution span's scope had exited
   — so it became a root span disconnected from the triage execution and the
   transfer that caused it. The caller-owned
   `execute_tool transfer_to_billing_agent` span context is now captured at the
   handoff seam and used as the parent of the target's execution span, so the
   run is one causal chain in one trace under the public `Runner.run`:

   ```
   invoke_agent triage-agent
     execute_tool transfer_to_billing_agent   interaction.type=handoff, agent.name=billing-agent
       invoke_agent billing-agent
   ```

   An active-agent change with no captured transfer falls back to the run's own
   context, so it still joins that trace. The holder is per run, so nothing leaks
   into the next `Runner.run`.

2. **`gen_ai.execute_tool.duration` was missing for the ordinary `get_weather`
   tool.** It is now recorded there too, timed in `finally` so a failed execution
   is still measured, dimensioned by the agent executing the tool and with no
   `gen_ai.agent.interaction.type` — an ordinary tool call is not an agent
   interaction. All three openai-agents tool operations now record exactly one
   point each, in a single stream:

   ```
   point count=1 {"gen_ai.agent.name": "test-agent", "gen_ai.tool.name": "get_weather", "gen_ai.tool.type": "function"}
   point count=1 {"gen_ai.agent.interaction.type": "delegation", "gen_ai.agent.name": "weather-specialist", "gen_ai.tool.name": "weather-specialist", "gen_ai.tool.type": "function"}
   point count=1 {"gen_ai.agent.interaction.type": "handoff", "gen_ai.agent.name": "billing-agent", "gen_ai.tool.name": "transfer_to_billing_agent", "gen_ai.tool.type": "function"}
   ```

Both were reproduced first as failing in-memory exporter / metric-reader
assertions and then proven fixed. Validation on the new head: `run-scenario
openai-agents` and `run-scenario google-adk` exit 0 with byte-identical
`data.json` (only pre-existing WARN classes), `ruff check` / `ruff format
--check` clean, `weaver registry check` with the pinned policy pack exits 0, and
`git diff --exit-code` over `docs model AGENTS.md` and over
`reference/README.md reference/reports reference/scenarios` both return 0.

The PR description has been updated with the corrected handoff span shape and the
duration coverage.

reference/scenarios/langchain/scenario.py
```diff
@@ -211,6 +254,338 @@ def run_execute_tool_reference():
...
+        tool_span_attributes = {
+            "gen_ai.operation.name": "execute_tool",
+            "gen_ai.tool.name": specialist_name,
+            "gen_ai.tool.type": "function",
+        }
+        with _reference_tracer.start_as_current_span(
+            f"execute_tool {specialist_name}",
+            attributes=tool_span_attributes,
+        ) as tool_span:
+            tool_span.set_attribute("gen_ai.agent.interaction.type", "delegation")
```
trask: Under the reference-review rules in #455, this does not demonstrate the field. `call_weather_specialist` is an application-defined ordinary `@tool`; LangChain exposes no API value that marks it as an agent delegation or associates it with `specialist`. Hardcoding `delegation` here only emits the desired string.

reference/scenarios/langchain/scenario.py
```diff
@@ -211,6 +254,338 @@ def run_execute_tool_reference():
...
+        tool_span_attributes = {
+            "gen_ai.operation.name": "execute_tool",
+            "gen_ai.tool.name": "transfer_to_weather_agent",
+            "gen_ai.tool.type": "function",
+        }
+        with _reference_tracer.start_as_current_span(
+            "execute_tool transfer_to_weather_agent",
+            attributes=tool_span_attributes,
+        ) as tool_span:
+            tool_span.set_attribute("gen_ai.agent.interaction.type", "handoff")
```
trask: `handoff` is derivable from the actual LangGraph result, but this sets it before the `Command` exists. Please construct or receive the command first, then emit this only after verifying `isinstance(command, Command)` and `command.graph == Command.PARENT`; read the target from `command.goto`. This makes the values come from library-owned state rather than the scenario's assertion.

model/gen-ai/registry.yaml
```diff
@@ -383,6 +383,29 @@ attributes:
...
+      members:
+        - id: delegation
+          value: "delegation"
+          brief: The source agent expects control or a result to return.
+          stability: development
+        - id: handoff
+          value: "handoff"
+          brief: The source agent transfers ownership of the remaining work.
+          stability: development
+    brief: The explicitly observed interaction type between two GenAI agents.
```
trask: The reference set is not repository-wide yet. Existing scenarios cover libraries with formal APIs that define both the interaction and target: Google ADK `TransferToAgentTool`, CrewAI `DelegateWorkTool`, AutoGen `Handoff`/`Swarm`, LlamaIndex `AgentWorkflow` handoff, and Claude Agent SDK `Task` subagents. Agent Framework `HandoffBuilder` should also be checked against the pinned core version. Please add references for the supported APIs instead of treating them as capture gaps.

model/gen-ai/metrics.yaml
```diff
@@ -196,6 +201,11 @@ metrics:
 
       This metric SHOULD be emitted together with the
       `gen_ai.invoke_agent.internal` span for the same invocation.
+
+      This metric counts the agent's own execution. A caller-owned
+      `invoke_agent` operation that only directs work to another agent — the
+      delegation or handoff operation carrying `gen_ai.agent.interaction.type` —
+      MUST NOT record it, so one agent invocation yields one data point.
```
trask: This carve-out is not demonstrated for `gen_ai.invoke_agent.inference_calls` or `gen_ai.invoke_agent.tool_calls`. The OpenAI handoff scenario only emits `gen_ai.invoke_agent.duration`, so the absence of extra points for these metrics could simply mean they are unimplemented. Please emit both execution metrics and assert one point per real agent execution and none for the caller-owned handoff operation.

lmolkova (commented):
Thanks for working on this! A few conceptual considerations on the proposed model:

1. **`gen_ai.agent.name` should not change meaning on tool calls**:
   On `execute_tool`, `gen_ai.agent.name` currently represents the agent executing the tool. Overwriting it with the target agent's identity when a transfer occurs drops the caller identity from the tool span and from `gen_ai.execute_tool.duration`. Caller tool metrics become misattributed to the target. The target should be captured in a dedicated attribute instead.

2. **"Handoff" and "delegation" are jargon and not expressive**:
   Terms like "handoff" vs "delegation" are confusing, especially for non-native English speakers. They do not clearly convey the operational behavior. What telemetry consumers actually need to know is:
   - Where did execution go (target)?
   - Is the source waiting for a result, or is control passed over?

3. **Both sides of the transfer do not need to be agents**:
   Transfers in real architectures frequently involve non-agent entities:
   - Initiator can be a non-agent (e.g. a deterministic workflow router node in LangGraph, an API gateway, or a human triage operator).
   - Target can be a non-agent (e.g. human-in-the-loop escalation / support queue, or a workflow step).
   The model should be consistent whether transferring between agent and agent, agent and human, or router node and agent.

---

### Alternative Proposal (Draft / Sketch)

*Note: This proposal was created with AI and is a sketch for discussion, so it should be polished and carefully considered.*

#### Attributes
- **`gen_ai.transfer.mode`**:
  - `return_to_caller`: The initiator waits for a result and resumes execution (sub-task / delegation).
  - `pass_control`: Control is handed over; the initiator does not wait for a return (handoff).
- **`gen_ai.transfer.target.name`**: Name/identifier of the recipient (target agent, human queue/role, or graph node).
  - Alternatively, target name could be named after the target entity directly (e.g. `gen_ai.transfer.target.agent.name`).
  - We can also incrementally add a target type attribute (`gen_ai.transfer.target.type`: `agent` | `step` | `human` | etc.) later as needed.
- **`gen_ai.agent.name`**: Retains its consistent meaning as the executing agent (caller on `execute_tool`, callee on `invoke_agent`).

#### Span and Metric Rules
- **Tool transfers**: Recorded on `execute_tool`:
  - `gen_ai.agent.name`: Source agent executing the tool.
  - `gen_ai.transfer.target.name`: Recipient.
  - `gen_ai.transfer.mode`: `return_to_caller` or `pass_control`.
  - `gen_ai.execute_tool.duration`: Preserves caller tool metrics.
- **Target execution**: When the target is an agent, it records its own `invoke_agent` span and metrics cleanly, without intermediate duplicate in-process spans or metric suppression rules.
- **Non-agent initiators**: A router node or workflow step records `gen_ai.transfer.mode` and `gen_ai.transfer.target.name` on its step span without requiring an agent name.

#### Framework Mapping

| Framework | Mechanism | Operation | `gen_ai.agent.name` | `gen_ai.transfer.target.name` | `gen_ai.transfer.mode` |
| :--- | :--- | :--- | :--- | :--- | :--- |
| **OpenAI Agents** | `agent.as_tool()` | `execute_tool` | Caller agent | Specialist agent | `return_to_caller` |
| **OpenAI Agents** | `handoff(target)` | `execute_tool` (`transfer_to_<target>`) | Caller agent | Target agent | `pass_control` |
| **Google ADK 2** | `AgentTool` | `execute_tool` | Caller agent | Specialist agent | `return_to_caller` |
| **Google ADK 2** | `TransferToAgentTool` | `execute_tool` (`transfer_to_agent`) | Caller agent | Target agent | `pass_control` |
| **Google ADK 2** | `adk_request_input` (HITL) | `execute_tool` (`adk_request_input`) | Caller agent | `human` / user | `return_to_caller` |
| **LangGraph** | Sub-agent as tool | `execute_tool` | Caller agent | Sub-agent | `return_to_caller` |
| **LangGraph** | `Command(goto="billing")` | `execute_tool` | Caller agent | `billing` | `pass_control` |
| **LangGraph** | `Command(goto="human_gate")` | `execute_tool` | Caller agent | `human_gate` | `pass_control` |
| **LangGraph** | Conditional router node | Node span (`invoke_workflow` step) | *(none)* | Target agent or node | `pass_control` |
| **CrewAI** | `Delegate work to coworker` | `execute_tool` | Manager agent | Coworker agent | `return_to_caller` |
| **Pydantic AI** | Tool calling `other_agent.run()` | `execute_tool` | Caller agent | Delegate agent | `return_to_caller` |

nikhilNava:
> Thanks for working on this! A few conceptual considerations on the proposed model:
> 
> 1. **`gen_ai.agent.name` should not change meaning on tool calls**:
>    On `execute_tool`, `gen_ai.agent.name` currently represents the agent executing the tool. Overwriting it with the target agent's identity when a transfer occurs drops the caller identity from the tool span and from `gen_ai.execute_tool.duration`. Caller tool metrics become misattributed to the target. The target should be captured in a dedicated attribute instead.
> 2. **"Handoff" and "delegation" are jargon and not expressive**:
>    Terms like "handoff" vs "delegation" are confusing, especially for non-native English speakers. They do not clearly convey the operational behavior. What telemetry consumers actually need to know is:
>    
>    * Where did execution go (target)?
>    * Is the source waiting for a result, or is control passed over?
> 3. **Both sides of the transfer do not need to be agents**:
>    Transfers in real architectures frequently involve non-agent entities:
>    
>    * Initiator can be a non-agent (e.g. a deterministic workflow router node in LangGraph, an API gateway, or a human triage operator).
>    * Target can be a non-agent (e.g. human-in-the-loop escalation / support queue, or a workflow step).
>      The model should be consistent whether transferring between agent and agent, agent and human, or router node and agent.
> 
> ### Alternative Proposal (Draft / Sketch)
> _Note: This proposal was created with AI and is a sketch for discussion, so it should be polished and carefully considered._
> 
> #### Attributes
> * **`gen_ai.transfer.mode`**:
>   
>   * `return_to_caller`: The initiator waits for a result and resumes execution (sub-task / delegation).
>   * `pass_control`: Control is handed over; the initiator does not wait for a return (handoff).
> * **`gen_ai.transfer.target.name`**: Name/identifier of the recipient (target agent, human queue/role, or graph node).
>   
>   * Alternatively, target name could be named after the target entity directly (e.g. `gen_ai.transfer.target.agent.name`).
>   * We can also incrementally add a target type attribute (`gen_ai.transfer.target.type`: `agent` | `step` | `human` | etc.) later as needed.
> * **`gen_ai.agent.name`**: Retains its consistent meaning as the executing agent (caller on `execute_tool`, callee on `invoke_agent`).
> 
> #### Span and Metric Rules
> * **Tool transfers**: Recorded on `execute_tool`:
>   
>   * `gen_ai.agent.name`: Source agent executing the tool.
>   * `gen_ai.transfer.target.name`: Recipient.
>   * `gen_ai.transfer.mode`: `return_to_caller` or `pass_control`.
>   * `gen_ai.execute_tool.duration`: Preserves caller tool metrics.
> * **Target execution**: When the target is an agent, it records its own `invoke_agent` span and metrics cleanly, without intermediate duplicate in-process spans or metric suppression rules.
> * **Non-agent initiators**: A router node or workflow step records `gen_ai.transfer.mode` and `gen_ai.transfer.target.name` on its step span without requiring an agent name.
> 
> #### Framework Mapping
> Framework	Mechanism	Operation	`gen_ai.agent.name`	`gen_ai.transfer.target.name`	`gen_ai.transfer.mode`
> **OpenAI Agents**	`agent.as_tool()`	`execute_tool`	Caller agent	Specialist agent	`return_to_caller`
> **OpenAI Agents**	`handoff(target)`	`execute_tool` (`transfer_to_<target>`)	Caller agent	Target agent	`pass_control`
> **Google ADK 2**	`AgentTool`	`execute_tool`	Caller agent	Specialist agent	`return_to_caller`
> **Google ADK 2**	`TransferToAgentTool`	`execute_tool` (`transfer_to_agent`)	Caller agent	Target agent	`pass_control`
> **Google ADK 2**	`adk_request_input` (HITL)	`execute_tool` (`adk_request_input`)	Caller agent	`human` / user	`return_to_caller`
> **LangGraph**	Sub-agent as tool	`execute_tool`	Caller agent	Sub-agent	`return_to_caller`
> **LangGraph**	`Command(goto="billing")`	`execute_tool`	Caller agent	`billing`	`pass_control`
> **LangGraph**	`Command(goto="human_gate")`	`execute_tool`	Caller agent	`human_gate`	`pass_control`
> **LangGraph**	Conditional router node	Node span (`invoke_workflow` step)	_(none)_	Target agent or node	`pass_control`
> **CrewAI**	`Delegate work to coworker`	`execute_tool`	Manager agent	Coworker agent	`return_to_caller`
> **Pydantic AI**	Tool calling `other_agent.run()`	`execute_tool`	Caller agent	Delegate agent	`return_to_caller`

Thanks @lmolkova for the detailed review.
I can incorporate the attribute suggestions for execute tool.

**For non tool based transfer:**
OpenAi today records a sperate Handoff span and not execute tool span. 
For such scenarios where there is non tool transfer to agent we can use the invoke agent client span, right?

Below is the actual span tree generated by open ai. 
image

We are recommending only tool based transfer to be captured via execute tool and non tool based transfer to be captured via invoke agent?

reference/scenarios/langchain/scenario.py
nikhilNava: Addressed. I removed the ordinary application-defined delegation example and now demonstrate the field through LangGraph's `Command` API. The scenario records a transfer only from a validated `Command.PARENT` result, derives the target from `command.goto`, and leaves `gen_ai.transfer.target.type` unset because LangGraph does not expose the target kind.

reference/scenarios/langchain/scenario.py
nikhilNava: Addressed in `06745ee`. The scenario now constructs the `Command` first, verifies `isinstance(command, Command)` and `command.graph == Command.PARENT`, validates that `command.goto` is a single named target, and only then records `gen_ai.transfer.mode=pass_control` and `gen_ai.transfer.target.name` from the command.

Krishnachaitanyakc:
@nikhilNava handoff is a function tool call in the SDK, so execute_tool covers it and it is not a capture gap. HandoffCallItem extends RunItemBase[ResponseFunctionToolCall] ([items.py#L266](https://github.com/openai/openai-agents-python/blob/v0.17.3/src/agents/items.py#L266)); convert_handoff_tool returns "type": "function" ([L887](https://github.com/openai/openai-agents-python/blob/v0.17.3/src/agents/models/chatcmpl_converter.py#L887)). #98 emits this today, happy to fold it in here under gen_ai.transfer.*.

nikhilNava:
> @nikhilNava handoff is a function tool call in the SDK, so execute_tool covers it and it is not a capture gap. HandoffCallItem extends RunItemBase[ResponseFunctionToolCall] ([items.py#L266](https://github.com/openai/openai-agents-python/blob/v0.17.3/src/agents/items.py#L266)); convert_handoff_tool returns "type": "function" ([L887](https://github.com/openai/openai-agents-python/blob/v0.17.3/src/agents/models/chatcmpl_converter.py#L887)). #98 emits this today, happy to fold it in here under gen_ai.transfer.*.

thanks for the clarification

alexlu4250 (approved):

nikhilNava:
The current branch incorporates both review threads:

- `gen_ai.agent.name` keeps its existing meaning on `execute_tool`: it identifies the agent executing the tool. The transfer target is recorded separately in `gen_ai.transfer.target.*`.
- `gen_ai.transfer.mode` uses the behavioral values `return_to_caller` and `pass_control`; the model no longer uses `handoff` or `delegation` as telemetry values.
- `gen_ai.transfer.target.type` supports `agent`, `workflow`, and `human`, so the target is not restricted to an agent.
- OpenAI Agents handoffs are modeled as the SDK's function-tool call: an `execute_tool` transfer refinement with `gen_ai.transfer.mode=pass_control`, while the target agent may emit its own `invoke_agent` execution span.

The relevant model and reference updates are in `233bb54`, `35a8917`, `30bf035`, and the later reference refinements. The latest head `137b089` also clears the CI failures; all required checks now pass.

nikhilNava:
@lmolkova @Krishnachaitanyakc the review items are addressed in the current branch, all review threads are resolved, and every required check passes on `137b089`. Ready for re-review.

reference/src/semconv_genai/refinement_coverage.py
```diff
@@ -0,0 +1,51 @@
...
+    for path in sorted(report_dir.glob("*.json")):
+        report = json.loads(path.read_text(encoding="utf-8"))
+        for sample in report.get("samples", []):
+            span = sample.get("span")
+            if not isinstance(span, dict):
+                continue
+            attributes = {
+                attribute["name"]: attribute.get("value")
+                for attribute in span.get("attributes", [])
+                if isinstance(attribute, dict) and isinstance(attribute.get("name"), str)
```
girishkvs: Can we exclude attributes Weaver marked `type_mismatch`, as the base coverage reducer does?

— girishkvs's Copilot 🤖

reference/src/semconv_genai/run_scenario.py
```diff
@@ -107,6 +108,8 @@ def main(argv: list[str] | None = None) -> int:
                 report_only=not args.strict,
                 extra_args=extra,
             )
+            if exit_code == 0:
+                update_span_refinement_coverage(reference_project_dir(library))
```
girishkvs: Can we update the selected output paths only after a complete run, respecting `--report-dir`, `--data-file` and `--scenario`?

— girishkvs's Copilot 🤖

reference/scenarios/google-adk/scenario.py
```diff
@@ -380,15 +497,276 @@ async def _run():
...
+                    client_span.set_attribute("gen_ai.agent.description", agent_card.description)
+                if request.context_id:
+                    client_span.set_attribute("gen_ai.conversation.id", request.context_id)
+                input_text = get_message_text(request, delimiter=" ")
+                if input_text:
+                    client_span.set_attribute(
+                        "gen_ai.input.messages",
+                        json.dumps([{"role": "user", "parts": [{"type": "text", "content": input_text}]}]),
+                    )
+                async for response in original_send_message(
```
girishkvs: Can we set `error.type` when this call raises? The current A2A report already has an `ERROR` client span without it.

— girishkvs's Copilot 🤖

reference/scenarios/langchain/scenario.py
```diff
@@ -211,6 +245,206 @@ def run_execute_tool_reference():
...
+
+    builder = StateGraph(AgentState)
+    builder.add_node(source_name, call_source)
+    builder.add_node(target_name, call_target)
+    builder.add_edge(START, source_name)
+    builder.add_edge(target_name, END)
+    graph = builder.compile()
+
+    input_text = "What's the weather in Seattle?"
+    result = await graph.ainvoke({"messages": [{"role": "user", "content": input_text}]})
```
girishkvs: Can we wrap this graph run in an `invoke_workflow` span so the source and target stay in the same trace?

— girishkvs's Copilot 🤖

reference/scenarios/google-adk/scenario.py
```diff
@@ -380,15 +497,276 @@ async def _run():
...
+            target_url = urlparse(agent_card.supported_interfaces[0].url)
+            client_attributes = {
+                "gen_ai.operation.name": "invoke_agent",
+                "gen_ai.agent.name": agent_card.name,
+                "gen_ai.caller.type": "workflow",
+                "gen_ai.caller.name": caller.name,
+                "gen_ai.provider.name": agent_card.provider.organization,
+                "server.address": target_url.hostname or "localhost",
+                "server.port": target_url.port or 443,
+            }
```
girishkvs, line 671: Can we record `agent_card.version` as `gen_ai.agent.version`? It's available on this card and conditionally required.

— girishkvs's Copilot 🤖

reference/scenarios/google-adk/scenario.py
```diff
@@ -380,15 +497,276 @@ async def _run():
...
+            with _reference_tracer.start_as_current_span(
+                f"execute_tool {agent_tool.name} {specialist.name}", attributes=tool_span_attributes
+            ) as tool_span:
+                if tool_context.function_call_id:
+                    tool_span.set_attribute("gen_ai.tool.call.id", tool_context.function_call_id)
+                tool_span.set_attribute("gen_ai.tool.call.arguments", json.dumps(args))
+                sub_agent_span_attributes = {
+                    "gen_ai.operation.name": "invoke_agent",
+                    "gen_ai.request.model": request_model,
+                    "gen_ai.agent.name": specialist.name,
```
girishkvs, line 564: Can we record `gen_ai.agent.description` on both delegation agent spans? Both descriptions are available and required by the base contract.

— girishkvs's Copilot 🤖

reference/src/semconv_genai/refinement_coverage.py
nikhilNava: System.Collections.Hashtable.body

reference/src/semconv_genai/run_scenario.py
nikhilNava: System.Collections.Hashtable.body

reference/scenarios/google-adk/scenario.py
nikhilNava: System.Collections.Hashtable.body

reference/scenarios/langchain/scenario.py
nikhilNava: System.Collections.Hashtable.body

reference/scenarios/google-adk/scenario.py
nikhilNava: System.Collections.Hashtable.body

reference/scenarios/google-adk/scenario.py
nikhilNava: System.Collections.Hashtable.body

girishkvs (approved):
Thanks — the five fixes look good on the new commit head.

One still open: the two delegation spans set `gen_ai.agent.name` but not `gen_ai.agent.description`, and `gen_ai.invoke_agent.internal` in the regenerated `data.json` still does not list it. Not blocking.

nit: Your six replies above posted as the literal text `System.Collections.Hashtable.body`, so the text did not make it — looks like an issue with pwsh script where `"$x.body"` was used instead of `"$($x.body)"`.

— girishkvs’s Copilot 🤖

neil-the-nowledgeable (commented):
Seems well built and consistent, thanks for the contribution

PR #453, opened 2026-08-16
Fix MCP session-duration metric attributes

### LINKED ISSUE #314: Questions about mcp.server.session.duration metric
### Area(s)

_No response_

### What's the problem?

After reading the doc about https://opentelemetry.io/docs/specs/semconv/gen-ai/mcp/#metric-mcpserversessionduration, I have several questions:

First of all, the doc says:

> This metric SHOULD be specified with [ExplicitBucketBoundaries](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.50.0/specification/metrics/api.md#instrument-advisory-parameters) of [ 0.01, 0.02, 0.05, 0.1, 0.2, 0.5, 1, 2, 5, 10, 30, 60, 120, 300 ]

So the max session duration is 300s. However, an MCP session can be preserved for hours or even for days. In the discussion https://github.com/modelcontextprotocol/modelcontextprotocol/discussions/1972, someone said they had a 30-day TTL for a session.

Secondly, there are several `network.*` attributes for `mcp.server.session.duration`. I guess it is a mirror of the `mcp.server.operation.duration`.
 But an MCP session, which is identified by MCP session ID, can contain several operations. One operation can go via TCP, another one can use QUIC. Therefore, adding `network.*` attributes to `mcp.server.session.duration` is not so reasonable.

_No response_

### Tip

danielpolimac:
## Motivation

MCP sessions can contain operations over more than one transport, so a session-duration metric cannot accurately carry per-operation transport details.

## Changes

- Remove `network.transport` and `network.protocol.version` from MCP client and server session-duration metrics.
- Retain session-stable protocol attributes and client endpoint attributes.
- Keep MCP-specific `network.protocol.name` examples for the session metrics.
- Keep transport attributes on per-operation metrics, where they describe the observed operation.

## Validation

- `make generate-all`
- `make check-policies`

Fixes #314


model/mcp/metrics.yaml
```diff
@@ -11,7 +11,6 @@ attribute_groups:
     visibility: internal
     attributes:
       - ref: mcp.protocol.version
-      - ref_group: mcp.network.protocol.attributes
       - ref: error.type
         requirement_level:
           conditionally_required: If and only if session ends with an error.
```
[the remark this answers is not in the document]
danielpolimac: Addressed in 547e96c. Removed the remaining client endpoint dimensions so both session-duration metrics have only session-level attributes.

docs/gen-ai/mcp.md
```diff
@@ -931,6 +931,7 @@ Examples:
 This metric SHOULD be specified with
 [`ExplicitBucketBoundaries`](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.55.0/specification/metrics/api.md#instrument-advisory-parameters)
 of `[ 0.01, 0.02, 0.05, 0.1, 0.2, 0.5, 1, 2, 5, 10, 30, 60, 120, 300 ]`.
+Measurements above 300 seconds are captured in the `+Inf` bucket.
```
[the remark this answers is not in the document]
danielpolimac: Addressed in 547e96c. Rephrased the note to refer to the exporter-neutral implicit overflow bucket.

changelog.d/+.bugfix.1.md
```diff
@@ -0,0 +1 @@
+Remove transport attributes from MCP session-duration metrics.
```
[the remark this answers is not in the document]
danielpolimac: Addressed in 547e96c. The changelog now describes removal of network and protocol attributes from the session-duration metrics.

model/mcp/metrics.yaml
```diff
@@ -11,7 +11,6 @@ attribute_groups:
     visibility: internal
     attributes:
       - ref: mcp.protocol.version
-      - ref_group: mcp.network.protocol.attributes
```
lmolkova, line 14: Dropping the whole `mcp.network.protocol.attributes` group removes more than what actually varies within a session.

`jsonrpc.protocol.version` and `network.protocol.name` are fixed for the lifetime of a session, and the transport *family* (stdio vs HTTP) is too - only `tcp` vs `quic` and the HTTP version can change between operations. Losing the stdio-vs-HTTP split takes away what is probably the most useful dimension on session duration, since the two have completely different lifetime distributions.

Suggest keeping the session-stable attributes and narrowing the removal to `network.transport` + `network.protocol.version`

model/mcp/metrics.yaml
```diff
@@ -50,7 +49,6 @@ metrics:
     stability: development
     attributes:
       - ref_group: mcp.session.metrics.attributes
-      - ref_group: mcp.client.address_and_port.attributes
```
lmolkova, line 53: Removing `server.address`/`server.port` is not supported by the motivation - the MCP endpoint is fixed for a session; a client cannot move a session to another server.

Without them the session-duration metric cannot be broken down per server, and cannot be joined with the client operation metrics and spans that keep these attributes.

docs/gen-ai/mcp.md
```diff
@@ -931,6 +931,7 @@ Examples:
 This metric SHOULD be specified with
 [`ExplicitBucketBoundaries`](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.55.0/specification/metrics/api.md#instrument-advisory-parameters)
 of `[ 0.01, 0.02, 0.05, 0.1, 0.2, 0.5, 1, 2, 5, 10, 30, 60, 120, 300 ]`.
+Measurements above 300 seconds are captured in the implicit overflow bucket.
```
lmolkova: This is standard OTel histogram behavior - measurements above the last boundary always fall into the last (`+Inf`) bucket - so it does not need to be restated on individual metrics. Please drop the sentence here and at line 997.

Also, "overflow bucket" is not the right term: in OTel, overflow refers to the attribute set produced when a cardinality limit is hit.

model/mcp/metrics.yaml
danielpolimac: Addressed in d8929b0. Retained network.protocol.name and jsonrpc.protocol.version, and removed only network.transport and network.protocol.version.

model/mcp/metrics.yaml
danielpolimac: Addressed in d8929b0. Retained the client endpoint attribute group.

docs/gen-ai/mcp.md
danielpolimac: Addressed in d8929b0. Removed the per-metric histogram sentence.

model/mcp/metrics.yaml
```diff
@@ -11,7 +11,12 @@ attribute_groups:
     visibility: internal
     attributes:
       - ref: mcp.protocol.version
-      - ref_group: mcp.network.protocol.attributes
+      - ref: network.protocol.name
+        requirement_level:
+          recommended: When applicable.
```
trask: Inlining the ref drops the `examples: ["http", "websocket"]` override from `mcp.network.protocol.attributes`, so both session-duration tables now advertise the generic registry examples `amqp`; `http`; `mqtt` instead of the MCP-specific ones every other MCP signal shows (docs/gen-ai/mcp.md lines 953 and 1025). Re-run `make generate-all` after fixing.

```suggestion
      - ref: network.protocol.name
        requirement_level:
          recommended: When applicable.
        examples:
          - "http"
          - "websocket"
```

changelog.d/+.bugfix.1.md
```diff
@@ -0,0 +1 @@
+Remove per-operation transport details from MCP session-duration metrics.
```
trask, line 1: The PR description still describes the first commit. It says JSON-RPC, network, and endpoint attributes are removed, but this head retains `jsonrpc.protocol.version`, `network.protocol.name`, `server.address`, and `server.port`, and the `+Inf` bucket clarification it lists is no longer in the diff. Only `network.transport` and `network.protocol.version` are removed now.

trask (commented):
AI-generated feedback

model/mcp/metrics.yaml
danielpolimac: Addressed in 85a5f78. Restored the MCP-specific http and websocket examples and regenerated the documentation.

changelog.d/+.bugfix.1.md
danielpolimac: Addressed. Updated the PR description to match the final scope: only network.transport and network.protocol.version are removed.

model/mcp/metrics.yaml
```diff
@@ -11,7 +11,15 @@ attribute_groups:
     visibility: internal
     attributes:
       - ref: mcp.protocol.version
-      - ref_group: mcp.network.protocol.attributes
+      - ref: network.protocol.name
```
lmolkova: `network.protocol.name` and `jsonrpc.protocol.version` are now defined twice - here and in `mcp.network.protocol.attributes` - with identical overrides that have to be kept in sync by hand. The `examples` override already drifted once in this PR.

Split the shared group instead: a session-stable group holding `network.protocol.name` + `jsonrpc.protocol.version`, referenced both here and from `mcp.network.protocol.attributes`, which keeps only the per-operation `network.transport` + `network.protocol.version`.

model/mcp/metrics.yaml
```diff
@@ -11,7 +11,15 @@ attribute_groups:
...
+        requirement_level:
+          recommended: When applicable.
+        examples:
+          - "http"
+          - "websocket"
+      - ref: jsonrpc.protocol.version
```
lmolkova: `network.transport` in `model/mcp/common.yaml` still carries the brief "The transport protocol used for the MCP session." If transport is per-operation, that brief is wrong on the signals that keep the attribute - it describes the transport of the observed operation, not of the session.

model/mcp/metrics.yaml
danielpolimac: Addressed in b55ea0c. Split the session-stable dimensions into `mcp.network.protocol.session.attributes`, shared by session metrics and the operation group; `network.transport` and `network.protocol.version` remain per-operation.

model/mcp/metrics.yaml
danielpolimac: Addressed in b55ea0c. Corrected the brief to refer to the observed MCP operation.

model/mcp/common.yaml
```diff
@@ -82,28 +82,33 @@ attribute_groups:
...
+  - id: mcp.network.protocol.attributes
+    visibility: internal
+    attributes:
+      - ref: network.transport
+        brief: The transport protocol used for the observed MCP operation.
+        note: |
+          This attribute SHOULD be set to `tcp` or `quic` if the transport protocol
+          is HTTP. It SHOULD be set to `pipe` if the transport is stdio.
+        examples: ["tcp", "quic", "pipe"]
+      - ref_group: mcp.network.protocol.session.attributes
```
lmolkova: instead of building deep hierarchies of groups, flatten them down and reference on the actual signals rather than other groups unless there is a good reason to.

`- ref_group: mcp.network.protocol.session.attributes` is not a big duplication, but makes yaml much more readable on the signal level

lmolkova (approved):

model/mcp/common.yaml
danielpolimac: Addressed in 9272e96. Removed the nested session group; session metrics now list their stable protocol attributes directly, and the operation group contains direct attribute references only.

lmolkova (approved):

PR #454, opened 2026-08-16
Disallow null optional GenAI JSON fields

### LINKED ISSUE #53: GenAI json schemas consider making optional fields not required instead of allowing null
### Area(s)

area:gen-ai

### What's missing?

Fields that we consider "optional" in the GenAI JSON schemas, like the tool call ID and mime type, are both not required and allow null. That allows both of these variants

```jsonc
{
  "type": "tool_call",
  "name": "mytool",
  "id": null,
}

// or optional
{
  "type": "tool_call",
  "name": "mytool",
}
```

Raised by @Kludex in https://github.com/open-telemetry/semantic-conventions/pull/2754/files#r2416223197

We could disallow nulls type while keeping them not required instead.

> Would it be possible to represent this as `NotRequired` (using `TypedDict` and `TypeAdapter`)/`MISSING` instead of `nullable`? 
>
> https://pydantic.dev/articles/pydantic-v2-12-release#missing-sentinel

This makes the JSON a bit more strict.

### Tip

danielpolimac:
## Motivation

Generated GenAI JSON schemas currently accept both a missing optional field and an explicit `null`. Consumers cannot distinguish omission from an explicit null value.

## Changes

- Keep optional fields optional while making their declared JSON type non-nullable.
- Regenerate the input/output message, tool-definition, retrieval-document, and memory-record schemas.
- Add a bugfix changelog fragment.

## Validation

- `make generate-all`
- `make check-policies`
- Verified an omitted tool-call `id` is accepted.
- Verified an explicit null tool-call `id` is rejected.

Fixes #53

### DIFF
--- a/changelog.d/+.bugfix.2.md
+++ b/changelog.d/+.bugfix.2.md
@@ -1,1 +1,1 @@
-Disallow null values for optional fields in generated GenAI JSON schemas.
+Distinguish omitted and null values in generated GenAI JSON schemas.
--- a/changelog.d/+.bugfix.2.md
+++ b/changelog.d/+.bugfix.2.md
@@ -1,0 +1,1 @@
+Disallow null values for optional fields in generated GenAI JSON schemas.
--- a/docs/gen-ai/non-normative/models.py
+++ b/docs/gen-ai/non-normative/models.py
@@ -37,6 +37,16 @@ from pydantic import (
 from pydantic_core import core_schema
 
 
+def omittable_field(**kwargs: Any):
+    """Define an optional, non-nullable JSON property."""
+
+    return Field(
+        default=None,
+        json_schema_extra=lambda schema: schema.pop("default", None),
+        **kwargs,
+    )
+
+
 # --------------------------------------------------------------------------
 # Common message-part models
 # --------------------------------------------------------------------------
@@ -63,11 +73,11 @@ class ToolCallRequestPart(BaseModel):
     type: Literal["tool_call"] = Field(
         description="The type of the content captured in this part."
     )
-    id: str = Field(
-        default=None, description="Unique identifier for the tool call."
+    id: str = omittable_field(
+        description="Unique identifier for the tool call."
     )
     name: str = Field(description="Name of the tool.")
-    arguments: Any = Field(default=None, description="Arguments for the tool call.")
+    arguments: Any = omittable_field(description="Arguments for the tool call.")
 
     model_config = ConfigDict(extra="allow")
 
@@ -80,7 +90,7 @@ class ToolCallResponsePart(BaseModel):
     type: Literal["tool_call_response"] = Field(
         description="The type of the content captured in this part."
     )
-    id: str = Field(default=None, description="Unique tool call identifier.")
+    id: str = omittable_field(description="Unique tool call identifier.")
     response: Any = Field(description="Tool call response.")
 
     model_config = ConfigDict(extra="allow")
@@ -122,8 +132,8 @@ class ServerToolCallPart(BaseModel):
     type: Literal["server_tool_call"] = Field(
         description="The type of the content captured in this part."
     )
-    id: str = Field(
-        default=None, description="Unique identifier for the server tool call."
+    id: str = omittable_field(
+        description="Unique identifier for the server tool call."
     )
     name: str = Field(description="Name of the server tool.")
     server_tool_call: ServerToolCall = Field(
@@ -139,8 +149,7 @@ class ServerToolCallResponsePart(BaseModel):
     type: Literal["server_tool_call_response"] = Field(
         description="The type of the content captured in this part."
     )
-    id: str = Field(
-        default=None,
+    id: str = omittable_field(
         description="Unique server tool call identifier matching the original call.",
     )
     server_tool_call_response: ServerToolCallResponse = Field(
@@ -180,12 +189,10 @@ class CompactionPart(BaseModel):
     type: Literal["compaction"] = Field(
         description="The type of the content captured in this part."
     )
-    id: str = Field(
-        default=None,
+    id: str = omittable_field(
         description="Provider-assigned identifier for the compaction item or block.",
     )
-    content: str = Field(
-        default=None,
+    content: str = omittable_field(
         description="The unencrypted compacted conversation summary, when available.",
     )
 
@@ -198,8 +205,8 @@ class BlobPart(BaseModel):
     type: Literal["blob"] = Field(
         description="The type of the content captured in this part."
     )
-    mime_type: str = Field(
-        default=None, description="The IANA MIME type of the attached data."
+    mime_type: str = omittable_field(
+        description="The IANA MIME type of the attached data."
     )
     modality: Union[Modality, str] = Field(
         description="The general modality of the data if it is known. Instrumentations SHOULD also set the mimeType field if the specific type is known."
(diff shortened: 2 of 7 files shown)


model/gen-ai/gen-ai-tool-definitions.json
```diff
@@ -16,25 +16,15 @@
...
                 "parameters": {
                     "anyOf": [
                         {
                             "$ref": "http://json-schema.org/draft-07/schema#"
-                        },
-                        {
-                            "type": "null"
                         }
                     ],
                     "default": null,
```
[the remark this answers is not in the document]
danielpolimac: Addressed in 136ff7e. Optional non-nullable fields now omit the generated `default: null` annotation.

model/gen-ai/gen-ai-retrieval-documents.json
```diff
@@ -5,30 +5,16 @@
...
-                        },
-                        {
-                            "type": "null"
-                        }
-                    ],
                     "default": null,
                     "description": "The relevance score of the document.",
-                    "title": "Score"
+                    "title": "Score",
+                    "type": "number"
```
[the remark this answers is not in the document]
danielpolimac: Addressed in 136ff7e. The generated retrieval-document fields are omittable, non-nullable, and no longer advertise a null default.

model/gen-ai/gen-ai-memory-records.json
```diff
@@ -9,44 +9,23 @@
...
-                        },
-                        {
-                            "type": "null"
-                        }
-                    ],
                     "default": null,
                     "description": "The relevance score of the memory record when populated on search results.",
-                    "title": "Score"
+                    "title": "Score",
+                    "type": "number"
```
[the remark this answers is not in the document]
danielpolimac: Addressed in 136ff7e. The generated memory-record fields are omittable, non-nullable, and no longer advertise a null default.

docs/gen-ai/non-normative/models.py
```diff
@@ -417,15 +417,15 @@ class FunctionToolDefinition(GenericToolDefinition):
...
         description=(
             "The description of the tool. "
             "Since this attribute could be large, it's NOT RECOMMENDED to be populated by default. "
             "Instrumentations MAY provide a way to enable populating this property."
         ),
     )
-    parameters: Optional[JsonSchemaDraft7Dict] = Field(
+    parameters: JsonSchemaDraft7Dict = Field(
         default=None,
         description=(
```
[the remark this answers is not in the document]
danielpolimac: Addressed in 136ff7e. Added a shared `omittable_field` helper that preserves the runtime default while omitting `default: null` from the generated schema, keeping the omittable/non-nullable contract consistent.

model/gen-ai/gen-ai-input-messages.json
```diff
@@ -420,25 +364,16 @@
...
-                        },
-                        {
-                            "type": "null"
-                        }
-                    ],
-                    "default": null,
                     "description": "Unique identifier for the tool call.",
-                    "title": "Id"
+                    "title": "Id",
+                    "type": "string"
```
lmolkova: This doesn't reject anything for message parts. `parts` items are an `anyOf` ending in `GenericPart`, which accepts any object with a string `type` plus `additionalProperties: true`, so the exact payload from #53 still validates:

```json
[{"role": "assistant", "parts": [{"type": "tool_call", "name": "mytool", "id": null}]}]
```

Same for `gen-ai-tool-definitions.json` via `GenericToolDefinition` (`{"type": "function", "name": "x", "description": null}` still passes). Only retrieval-documents and memory-records actually get stricter, and memory-records isn't in the conformance policy's schema map, so nothing enforces it.

Closing #53 needs the catch-all variants to stop swallowing known types - e.g. make the union discriminate on `type` so a known `type` value has to match its specific schema.

docs/gen-ai/non-normative/models.py
```diff
@@ -37,6 +37,16 @@
 from pydantic_core import core_schema
 
 
+def omittable_field(**kwargs: Any):
```
lmolkova, line 41: Have you explored `MISSING` / `NotRequired` as suggested in #53? That is the part of the problem this helper leaves open: it changes what the schema says without changing what the models do, so the annotation is now false and the round trip is broken.

```python
ToolCallRequestPart(type="tool_call", name="x").model_dump_json()
# {"type":"tool_call","id":null,"name":"x","arguments":null}  <- rejected by the new schema
```

`MISSING` (pydantic 2.13) covers serialization and deserialization together - omitted on dump, rejected on parse, and the generated schema is the one this PR wants:

```python
from pydantic.experimental.missing_sentinel import MISSING

id: str | MISSING = Field(MISSING, description="Unique identifier for the tool call.")
```

docs/gen-ai/non-normative/models.py
```diff
@@ -417,16 +424,14 @@ class FunctionToolDefinition(GenericToolDefinition):
     """
 
     type: Literal["function"] = Field(description="The type of the tool.")
-    description: Optional[str] = Field(
-        default=None,
+    description: str = omittable_field(
```
lmolkova: Too strict here. Omission of `description`/`parameters` already has a defined meaning in this convention - capture is opt-in, so absent means "not collected". A provider-reported null means something different: the tool has no description / takes no parameters. Banning null collapses the two.

It also pushes work onto every producer: SDK objects carry these as optional and serialize to null. The repo's own scenario does it - `reference/scenarios/openai-agents/scenario.py:80` emits `"description": t.description`, which is `None` for a tool without a docstring.

Either keep null allowed on these two and document what it means, or state that instrumentations MUST drop null before emitting.

docs/gen-ai/non-normative/models.py
```diff
@@ -63,11 +73,11 @@ class ToolCallRequestPart(BaseModel):
...
         description="The type of the content captured in this part."
     )
-    id: Optional[str] = Field(
-        default=None, description="Unique identifier for the tool call."
+    id: str = omittable_field(
+        description="Unique identifier for the tool call."
     )
     name: str = Field(description="Name of the tool.")
-    arguments: Any = Field(default=None, description="Arguments for the tool call.")
+    arguments: Any = omittable_field(description="Arguments for the tool call.")
```
lmolkova: `arguments` is `Any`, so `"arguments": null` stays valid while `"id": null` becomes invalid - and this is the field where the distinction actually carries information (arguments not captured vs. the call genuinely had null arguments). Same for `response` and memory `content`.

The PR makes omission normative without saying what it means. Spell out, for each of these fields, what absent means and what null means, otherwise consumers still can't rely on either.

docs/gen-ai/non-normative/models.py
danielpolimac: Addressed in de056bd. The models now use Pydantic MISSING so unavailable fields are omitted during serialization, while explicit null values remain representable where captured.

docs/gen-ai/non-normative/models.py
danielpolimac: Addressed in de056bd. Tool descriptions and parameters are now typed as value, null, or MISSING; their descriptions state the distinct capture semantics.

docs/gen-ai/non-normative/models.py
danielpolimac: Addressed in de056bd. Arguments uses MISSING, and the model descriptions distinguish omission from captured null for arguments, tool responses, and memory content.

changelog.d/+.bugfix.2.md
```diff
@@ -0,0 +1 @@
+Distinguish omitted and null values in generated GenAI JSON schemas.
```
trask: This rejects telemetry that used to validate: `"name": null` on a message, and `"id": null`, `"score": null` or `"metadata": null` on a retrieval document or memory record. That makes it a breaking change, so rename this to `454.breaking.md` and say that null is no longer accepted.

PR #469, opened 2026-08-23
Remove cache token usage attributes from internal agent spans

### LINKED ISSUE #460: Remove cache token usage attributes from internal agent spans
We have general and cache usage tokens on internal spans https://github.com/open-telemetry/semantic-conventions-genai/blob/main/docs/gen-ai/gen-ai-agent-spans.md#invoke-agent-internal-span

It creates issues like https://github.com/open-telemetry/semantic-conventions-genai/issues/19 for which we don't have a good answer for.

And while general token usage is still useful on agent spans, cache breakdown (along with other breakdown usage attributes being introduced in https://github.com/open-telemetry/semantic-conventions-genai/pull/440) would be misleading: they need to be aggregated across different models and multiple inference calls.  Since we're talking about aggregating anyway, it can be done by the backends over traces and will eventually be done in some way via metrics (https://github.com/open-telemetry/semantic-conventions-genai/pull/374 and https://github.com/open-telemetry/semantic-conventions-genai/pull/197).

This issue tracks removing cache usage tokens from internal spans, leaving general usage to #19

lmolkova:
## Description

Fixes #460

Removes `gen_ai.usage.cache_read.input_tokens` and `gen_ai.usage.cache_write.input_tokens` from the internal `invoke_agent` span. The inference span and the client `invoke_agent` span are unaffected.

## Motivation

An internal agent span reports usage accumulated over its whole call tree, which can span multiple models and inference calls. A cache breakdown at that level mixes models with different cache semantics and pricing, so the number is not interpretable on its own.

**User journey:** cost allocation and cache-efficiency analysis. Both need the breakdown attributed to a model, which an internal agent span cannot provide. Consumers can aggregate from the `gen_ai.inference.client` spans in the tree, and metrics will cover it (#374, #197).

**Prior art:** provider APIs (OpenAI, Anthropic, Gemini) report cache token counts per inference response only - no agent framework exposes a cache breakdown at agent scope. The same argument already kept modality breakdown attributes (#440) off this span.

Total token usage stays on the span; whether that is the right shape is tracked separately in #19.

## Prototype

N/A - Agent Framework instrumentation is the only one that emits `cache_read` on the internal agent span, and it's native.

## Checklist


JWinermaSplunk (approved):

changelog.d/469.breaking.md
```diff
@@ -0,0 +1 @@
+Remove `gen_ai.usage.cache_read.input_tokens` and `gen_ai.usage.cache_write.input_tokens` from the internal `invoke_agent` span. Cache breakdowns aggregate across models and inference calls, which makes them misleading on agent spans; consumers should aggregate them from `gen_ai.inference.client` spans instead.
```
trask: The client `invoke_agent` span keeps both attributes, so "on agent spans" overstates the removal.

```suggestion
Remove `gen_ai.usage.cache_read.input_tokens` and `gen_ai.usage.cache_write.input_tokens` from the internal `invoke_agent` span. Cache breakdowns on that span aggregate across models and inference calls, which makes them misleading; consumers should aggregate them from `gen_ai.inference.client` spans instead.
```

trask (approved):

PR #475, opened 2026-08-25
Add guidance on how to avoid duplicate `inference` spans

DylanRussell:
## Description

Add guidance that instrumentations should look for the `inference` span/event within the context under a key (location / value TBD.. for now just saying a language-defined context key), if it exists they should suppress their instrumentation of the inference span/event (AND associated metrics) OR optionally modify the span/event with additional `attributes` if they so wish (server address/port are example attributes that downstream instrumentations may have but upstream may not..). If the span/event doen't exist in the context under the key, the instrumentation should add it.

If multiple instrumentations set an attribute on a span/event, the last instrumentation to set it wins out.. In general the instrumentation setting the span / event in the context is responsible for most everything, setting the right attributes, emitting the span / event etc.

## Motivation

- **User journey:** Applications frequently layer prompt compilers (such as DSPy), gateways (such as Portkey), agent frameworks, and provider client SDKs. Without context coordination, multiple instrumentations emit duplicate inference spans and double-count client duration and token metrics for a single logical model call.
- **Prior art:** OpenTelemetry HTTP client instrumentations suppress nested calls via context keys. `opentelemetry-python-genai#479` implements this pattern for GenAI inference invocations.

## Prototype

Implemented and tested in [open-telemetry/opentelemetry-python-genai#479](https://github.com/open-telemetry/opentelemetry-python-genai/pull/479), where inference invocations store the active span under `opentelemetry.genai.inference_span` in the OpenTelemetry context so downstream instrumentations can inspect, enrich, or suppress tracing.

## Checklist


docs/gen-ai/gen-ai-spans.md
```diff
@@ -408,6 +416,21 @@ and SHOULD be provided **at span creation time** (if provided at all):
...
+- **Setting the inference span in context**: The library that creates the inference span (the first layer in the call stack to start the span for that model call) SHOULD store the active `Span` in the OpenTelemetry `Context` under the key `opentelemetry.genai.inference_span`. When the inference span finishes (or ends upon error or cancellation), the context SHOULD be restored.
+- **Checking for an existing inference span**: Any instrumentation that is about to create an inference span—if it is not certain whether it is at the top of the call stack or if an upstream library has already started one—SHOULD check whether an active inference span is already present in the current context under `opentelemetry.genai.inference_span`.
+- **Enriching vs. suppressing**: If an active inference span is already present in the context:
+  - Downstream instrumentations MAY enrich the existing span by setting attributes that were not available to the upstream caller (for example, low-level connection attributes such as `server.address` and `server.port`, or provider-specific metadata).
```
RKest: If upstream instrumentor already set some attributes, who's set of attributes wins out?

I.e. could a downstream instrumentor overwrite some attributes?

docs/gen-ai/gen-ai-spans.md
```diff
@@ -408,6 +416,21 @@ and SHOULD be provided **at span creation time** (if provided at all):
...
+  - Downstream instrumentations SHOULD NOT create a duplicate inference span for the same logical model call.
+  - Downstream instrumentations SHOULD NOT record duplicate client metrics (such as `gen_ai.client.operation.duration` or `gen_ai.client.token.usage`) for the call if the existing span records them.
```
RKest: What about logs?

In ADK, the main use-case for OTel instrumentor, is the GCS object reference and these references can be stored in logs or spans.

How about in addition to the span-in-context, we can also have a `ReadWriteLogRecord` in context, which would allow downstream instrumentations to enrich the [gen_ai.client.inference.operation.details](https://github.com/open-telemetry/semantic-conventions-genai/blob/main/docs/gen-ai/gen-ai-events.md#event-gen_aiclientinferenceoperationdetails) log.
Then, the instrumentation which put the log-record there, would be responsible for the final emission of it.

This will take care of a pretty big feature gap in ADK, where the `opentelemetry-instrumentation-google-genai` emitted logs, don't hold `user.id`, which I have been trying to close for a while.

What I tried previously, and both seems like an overkill:
- https://github.com/open-telemetry/opentelemetry-python-genai/pull/337
- https://github.com/open-telemetry/opentelemetry-specification/issues/5247

DylanRussell: That's a good point, we could expand it to logs.. in your example ADK would put it on the context (in the request path), `opentelemetry-instrumentation-google-genai` would populate it with `input_messages/output_messages/system_messages` etc. , stamp the GCS refs on it, then ADK would would stamp `user.id` on it and emit the log (on the response path)... ?

If the instrumentation library adds the log to the context and emits it, that would bypass ADK entirely..

docs/gen-ai/gen-ai-spans.md
DylanRussell: Yeah I think last instrumentation to set the attribute wins out. On the request path that'll be the library closest to the model call, and on the response path that'll be the one furthest away..

DylanRussell: Added language to clarify it..

docs/gen-ai/gen-ai-spans.md
```diff
@@ -408,6 +417,21 @@ and SHOULD be provided **at span creation time** (if provided at all):
 <!-- END AUTOGENERATED TEXT -->
 <!-- endweaver -->
 
+#### Handling existing inference spans
+
+An inference call can be initiated or wrapped by various libraries in the call stack, including prompt programming or optimization frameworks (such as [DSPy](https://github.com/stanfordnlp/dspy)), AI gateways, routers, or proxies (such as [Portkey](https://portkey.ai/) or LiteLLM), agent frameworks, workflow orchestrators, or direct provider client SDKs (such as OpenAI, Anthropic, or Google GenAI).
+
+Because multiple libraries participating in the same model call may be instrumented simultaneously:
+
+- **Context key**: Instrumentations SHOULD use the well-known context key `opentelemetry.genai.inference_span` to store and access the active inference span in the OpenTelemetry `Context`.
```
aabmass: Chatted in the SIG but the context keys are supposed to be opaque, so specifying a string won't work. Can we make it more vague (something like "languages should expose a context key per span type") and leave it up to language implementations to implement?

aabmass (commented):
I think this is a great idea, glad to see it written down

docs/gen-ai/gen-ai-spans.md
DylanRussell: Ack switched to using that language..

docs/gen-ai/gen-ai-spans.md
RKest: Yes, exactly!

Instrumentation library should read the log in-context, and only enrich it if present.

Instrumentation lib putting the log in-context wouldn't make much sense, because AFAIU it's always the last piece in the chain.

DylanRussell: Ack.. I don't see any reason not to expand it to the inference event.. i updated the PR to include it now

docs/gen-ai/gen-ai-events.md
```diff
@@ -31,7 +31,13 @@ The event name MUST be `gen_ai.client.inference.operation.details`.
...
 Describes the details of a GenAI completion request including chat history and parameters.
 
-This event could be used to store input and output details independently from traces.
+This event could be used to store input and output details
+independently from traces.
+
+When capturing this event, instrumentation SHOULD store it in the OpenTelemetry
+context on the request path using a language-defined context key for inference events.
+This allows instrumentations further down the call stack to modify or enrich the event
+if they wish, and avoids emitting duplicate events for the same call.
```
lmolkova: I'm not sure it's implementable in general case.

I'd mean we need to create event when operation starts, put it on the context, then update to add attributes - OTel Logs API does not assume attributes are readable or updateable.

docs/gen-ai/gen-ai-spans.md
```diff
@@ -422,6 +432,23 @@ and SHOULD be provided **at span creation time** (if provided at all):
 <!-- END AUTOGENERATED TEXT -->
 <!-- endweaver -->
 
+#### Handling existing inference spans
```
lmolkova, line 435: Do we have a prototype for it in python repo? I consider semconv guidance to be nice to have, suppression is an implementation detail and I'd rather start with python implementation than here.

Do we need mechanism to be specific to inference? We can make it generic and based on span type. I.e. when instrumentations creates a span type X, check if it's already on the context. If no, start a span, put it on context under key dedicated to that span type. If yes, just enrich spans (we need to figure out what to do with events). Don't report metrics.

I also feel this text can be made much more concise and actionable, all the examples, justification and context can stay in the PR description.

docs/gen-ai/gen-ai-events.md
DylanRussell: The only issue with events that I see is that they are often created / emitted in one step, but that isn't true for this event. I don't see any issue with reading / overwriting attributes on a log.. Although attributes on a span are not readable.

docs/gen-ai/gen-ai-spans.md
DylanRussell: Ok I've created a prototype here that was a good idea and makes the proposal more clear: https://github.com/open-telemetry/opentelemetry-python-genai/pull/479

I don't think the mechanism should be specific to inference (check out the implementation I made) but I don't know if it makes sense for any of the other span types, and I would start out just doing it for inference

docs/gen-ai/gen-ai-spans.md
DylanRussell: I think what you want will be hard actually.. I'm now proposing that if the instrumentation finds the event/span on the context, they assume that most of the attributes are set by the upstream instrumentation, and only set attributes on the span / event that they know the upstream instrumentation will not be able to

docs/gen-ai/gen-ai-spans.md
DylanRussell: I shortened the language used in the PR too.

docs/gen-ai/gen-ai-events.md
lmolkova: >  I don't see any issue with reading / overwriting attributes on a log.

Not in python. But there is no spec language that guarantees it's possible. Try doing it in Java or .NET

DylanRussell: i agree with you.. it's probably not possible to say something like "instrumentations should check if a span/log attribute is set, and if it is don't set it.." currently I have the following language in this sem conv "Downstream instrumentation MAY enrich the existing span and event with additional attributes. When attributes conflict, the last write wins."

felixfbecker:
How are you thinking about this if the duplicate spans come from a different service? E.g. as mentioned in the user journey in the PR description, there is a gateway service that emits duplicate spans, which wouldn't have access to in-process context.

In a way I also feel like we don't actually should try to prevent "duplicate" spans — if both the client and the server emit a span, you want both spans because they can have timing differences due to network latency etc. If a gateway or a server modifies the prompt, why can't they just both emit their version of the prompt, and the consumer of the traces filters/defines preference based on `service.name` etc? Curious to understand the context of this more

DylanRussell:
The problem we are try to solve is de-duplication of the same client side span. Also enhancement of the span attributes. 

It's common for several instrumented client-side libraries to be installed together, all of which have reasons for emitting the same span (ex: `adk` and `google-genai` SDK or `agno` / `dspy` and `litellm`), also libraries may have additional information the other library doesnt have that they want to add to it..

PR #479, opened 2026-08-29
Add GenAI client metrics to the groq reference scenario

tishachawla-jg:
## Description

Adds `gen_ai.client.token.usage` and `gen_ai.client.operation.duration` to the `groq` reference scenario. No changes to `model/` or `docs/` — this is reference-scenario coverage only.

Refs #281.

## Motivation

**User journey:** cost allocation and capacity planning. Per-span usage attributes answer "what did this one call cost"; the client metrics answer "what is token consumption doing across all calls", which is what dashboards and alerts are built on. Both are already defined in the conventions.

The gap is demonstration, not definition. 13 scenarios emit inference spans, but only 2 of 28 (`anthropic`, `agent-framework`) emit either client metric — and both get them indirectly, `anthropic` via the `util-genai` handler and `agent-framework` via the library's built-in instrumentation. Nothing demonstrated them being recorded explicitly against a plain vendor SDK, which is the shape most instrumentation authors are actually writing.

**Prior art:** `anthropic` and `agent-framework` already produce both metrics; this PR reuses their attribute sets rather than introducing a new shape.

## Prototype

`reference/scenarios/groq/scenario.py` records both metrics alongside the existing spans, using `reference_meter()` from `reference_shared` (previously unused by this scenario).

Resulting `data.json` attribute sets are identical to those `anthropic` already produces:

| Metric | Attributes |
|---|---|
| `gen_ai.client.operation.duration` | `gen_ai.operation.name`, `gen_ai.provider.name`, `gen_ai.request.model`, `gen_ai.response.model`, `server.address`, `server.port` |
| `gen_ai.client.token.usage` | the same, plus `gen_ai.token.type` |

Token usage is recorded once per `gen_ai.token.type`.

**Streaming records duration only.** The stream in this scenario carries no usage block, and the convention states instrumentation "MUST NOT report usage metric" when it cannot obtain token counts. Happy to add `stream_options={"include_usage": True}` and report usage there too if you'd prefer the scenario exercise that path — it changes what the scenario asks of the mock server, so I left it out rather than decide unilaterally.

**No new violations.** `uv run run-scenario groq` reports `WARN` with five findings. All five are present on `main` without this change (verified by running the scenario with the change stashed): the `gen_ai.tool.definitions` schema finding, two missing usage attributes on the streaming span, missing `server.address`, and missing `gen_ai.provider.name`. This PR neither adds nor fixes any of them.

Follow-ups, if this shape looks right: the same treatment for the other ~10 scenarios that emit inference spans but no metrics, and `gen_ai.client.operation.time_to_first_chunk` for the streaming paths.

## Checklist

- [ ] Towncrier fragment — not added: this changes no conventions, only reference coverage. Followed the precedent of #283, which added metrics to the `anthropic` scenario without a fragment. Happy to add one if you'd rather it be noted.


tishachawla-jg:
C:/Program Files/Git/easycla

lmolkova (approved):
Please address copilot comment

tishachawla-jg:
Addressed both: added the advised explicit bucket boundaries to each instrument, and inlined the attribute dictionaries and token observations at each call site so the provenance of every value stays visible next to the request arguments and response fields it comes from. Regenerating the scenario produces an unchanged `data.json`.

tishachawla-jg:
Hey @lmolkova, worked on the changes. Can you please merge this?

lmolkova (approved):

Issue #468, opened 2026-08-23
Agent spans: no attribute for tools offered to the model vs. tools actually called


sauravbhattacharya001:
### What I was doing

Hand-instrumenting a third-party agent (smolagents `ToolCallingAgent`) to the GenAI agent span conventions — manual spans, no instrumentation library — to find where the conventions did and didn't fit a real agent loop.

Working rule: anything the conventions don't define gets an `app.*` prefix rather than an invented `gen_ai.*` name.

### The gap

The conventions capture tools **called** (`execute_tool` spans, `gen_ai.tool.name`). I couldn't find anything for the tools **offered** to the model at a given step — the inventory it was choosing from.

Without it, a `chat` span that made no tool call is ambiguous: the model may have declined an available tool, or may have had none offered. Those are different behaviours, and tool-selection evaluation depends on telling them apart.

I emitted `app.tools_offered_count` on the `chat` span. A name and a shape in the spec would obviously be better.

### Three related things I also couldn't place

- how many tool calls a single completion emitted
- a tool that returned successfully but returned nothing usable — distinct from span status `ERROR`
- an agent loop that exhausted its step budget and produced an answer anyway (5 of 10 runs in my sample, indistinguishable from clean runs at the outcome level)

Happy to open these separately if they're better as distinct issues.

### Context

Traces, harness and writeup: https://github.com/sauravbhattacharya001/otel-genai-agent-eval

Is any of this already covered somewhere I've missed?

dgenio:
A second framework gives the same gap and suggests the recording boundary should be the **individual model invocation/turn**, not a longer-lived agent span.

In the OpenAI Agents SDK, the runtime resolves the effective tools and handoffs immediately before each model call, after dynamic enablement/filtering and name-collision handling. That set can change between turns (`FunctionTool.is_enabled`, `Handoff.is_enabled`, MCP `tool_filter`, workflow state, feature flags, etc.). The SDK currently writes the resolved names onto a longer-lived agent span, so later turns can overwrite earlier capability sets; I opened openai/openai-agents-python#4626 for that SDK-specific tracing gap.

The useful semantic distinction seems to be:

```text
offered / model-visible capabilities
        ↓
selected / requested tool calls
        ↓
actually executed tools
```

Those are three different facts. In particular, `execute_tool` spans only prove the last one.

A minimal conformance scenario would be:

```text
same agent
turn 1: offered [draft_invoice] -> model calls draft_invoice
workflow state changes
[...]

PR #481, opened 2026-08-30
Clarify the scope of `gen_ai.tool.definitions`

sauravbhattacharya001:
## Description

`gen_ai.tool.definitions` is referenced by inference spans, agent spans, and the fetch-response span, and its brief covers all of them — so a consumer cannot tell whether it holds the set of tools provided to one model request or the agent's whole configured catalog.

Adds a note per reference saying what the recorded set means for that signal. Note-only: brief, type, requirement level, and the JSON schema are unchanged, and the attribute still resolves onto exactly the same spans.

Related to #468.

## Motivation

**User journey:** triaging agent runs. "Which tools could the model have chosen from at this step?" cannot be answered today, because the same key means different things on different spans.

**Prior art:** two frameworks hit this independently in #468 — smolagents, where the available set mutates between steps, and the OpenAI Agents SDK, where tools are resolved before each model call but written to a longer-lived agent span ([openai/openai-agents-python#4626](https://github.com/openai/openai-agents-python/issues/4626)).

## Prototype

No reference scenario changes: no attribute added, no emitted value changed, no schema touched.

## Checklist

- [x] Reference scenarios updated for affected libraries — n/a, no emitted telemetry changes


model/gen-ai/registry.yaml
```diff
@@ -540,6 +540,20 @@ attributes:
...
+        effective set of tools provided to that request, recorded after any runtime
+        filtering, resolution, or name-collision handling has been applied.
+      - On spans that cover an agent invocation spanning multiple model requests,
+        this is the set of tools configured for the agent. When the effective set
+        varies between requests, this attribute does not necessarily reflect the set
+        provided to any individual request.
+
+      Consumers that depend on the exact set of tools provided to a specific model
+      request SHOULD read this attribute from the corresponding inference span or
+      event.
```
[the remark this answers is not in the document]
sauravbhattacharya001: @copilot fix this issue

model/gen-ai/registry.yaml
```diff
@@ -540,6 +540,20 @@ attributes:
       > [!WARNING]
       > This attribute may contain sensitive information.
 
+      The scope of the recorded set depends on the signal it is reported on:
+
+      - On spans and events that describe a single model request, this is the
+        effective set of tools provided to that request, recorded after any runtime
+        filtering, resolution, or name-collision handling has been applied.
```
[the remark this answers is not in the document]
sauravbhattacharya001: Added a fetch-response bullet to the note ("the set of tools carried on the fetched response", matching the existing `gen_ai.system_instructions` note on that span) and regenerated the docs in 7e468e5.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

model/gen-ai/registry.yaml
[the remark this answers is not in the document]
sauravbhattacharya001: Added a fetch-response bullet to the note ("the set of tools carried on the fetched response", matching the existing `gen_ai.system_instructions` note on that span) and regenerated the docs in 7e468e5.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

model/gen-ai/registry.yaml
```diff
@@ -540,6 +540,20 @@ attributes:
...
+      - On spans that cover an agent invocation spanning multiple model requests,
+        this is the set of tools configured for the agent. When the effective set
+        varies between requests, this attribute does not necessarily reflect the set
+        provided to any individual request.
```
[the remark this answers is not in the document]
sauravbhattacharya001: Agreed — the agent bullet defined every agent-span value as the configured catalog, which would have made the OpenAI Agents SDK behavior cited in the motivation nonconforming. The bullet now describes both recording behaviors (configured catalog, or an effective set resolved for one constituent request) in 426ad5f, and the PR description was updated to match. The guidance stays non-normative, so neither producer is invalidated.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

model/gen-ai/registry.yaml
```diff
@@ -540,6 +540,23 @@ attributes:
       > [!WARNING]
       > This attribute may contain sensitive information.
 
+      The scope of the recorded set depends on the signal it is reported on:
```
lmolkova: this belongs in the note where this attribute is referenced. this way it'd be focused and much easier to find. Also then it won't be accidentally overridden when something else is specialized on attribute refence

model/gen-ai/registry.yaml
```diff
@@ -540,6 +540,23 @@ attributes:
...
+
+      - On spans and events that describe a single model request, this is the
+        effective set of tools provided to that request, recorded after any runtime
+        filtering, resolution, or name-collision handling has been applied.
+      - On spans that fetch a previously generated model response, this is the set
+        of tools carried on the fetched response.
+      - On spans that cover an agent invocation spanning multiple model requests,
+        this is the set of tools configured for the agent; some instrumentations
+        instead record an effective set resolved for one of the constituent
+        requests. When the effective set varies between requests, this attribute
```
lmolkova: > some instrumentations
        instead record an effective set resolved for one of the constituent
        requests

we should be prescriptive - maybe it depends on the provider/framework, but should be the same across different instrumentations for that provider/technology.

lmolkova:
@sauravbhattacharya001 please consider shortening PR description. it has a lot of low level details and bigger than the change itself.

model/gen-ai/registry.yaml
sauravbhattacharya001: Moved in 0b6d556. Confirmed the override risk is real: on the fetch-response span, which specializes the note on `gen_ai.system_instructions`, the registry note is dropped entirely from the generated docs. Removed `gen_ai.tool.definitions` from the shared `attributes.gen_ai.content` group and referenced it explicitly on the inference span, the shared invoke-agent group, and the fetch-response span, each with the note for that signal. The resolved attribute set is unchanged — every generated table carries it on exactly the spans it did before.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

model/gen-ai/registry.yaml
sauravbhattacharya001: Made prescriptive in 0b6d556: the agent-span note now says the choice between the configured catalog and a per-request effective set may depend on the provider or framework, but instrumentations of the same provider or framework SHOULD record it consistently.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

model/gen-ai/registry.yaml
sauravbhattacharya001: Follow-up on the move (25db64f): because a ref-level note replaces the registry note rather than adding to it, moving the scope guidance to the references also dropped the sensitive-information warning and the payload-size guidance from all seven generated docs. I now repeat both in each reference note, so the output matches main.

That does mean the warning and size paragraph are duplicated across three reference notes. Happy to restructure if you would prefer a different split. Worth noting the same override already affects `gen_ai.system_instructions` on `gen_ai.fetch_response.client` on main, where its registry warning does not reach the generated docs.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

sauravbhattacharya001:
> @sauravbhattacharya001 please consider shortening PR description. it has a lot of low level details and bigger than the change itself.

done, thanks, please have a look @lmolkova

model/gen-ai/spans.yaml
```diff
@@ -117,6 +115,28 @@ attribute_groups:
...
+
+          Whether this is the agent's configured catalog or the effective set resolved
+          for one of the constituent requests depends on the provider or framework.
+          Instrumentations of the same provider or framework SHOULD record it
+          consistently.
+
+          When the effective set varies between requests, this attribute does not
+          necessarily reflect the set provided to any individual request. Consumers
+          that depend on the exact set of tools provided to a specific model request
+          SHOULD read this attribute from the corresponding inference span or event.
```
lmolkova, line 132:
```suggestion
        brief: The tool set associated with the agent.
        note: |
          It may differ from the tools provided to individual inference operations performed by the agent. 
```

I think this conveys the same level of details.

model/gen-ai/spans.yaml
```diff
@@ -384,6 +417,15 @@ spans:
...
         requirement_level: opt_in
+        note: |
+          The set of tools carried on the fetched response.
+
+          > [!WARNING]
+          > This attribute may contain sensitive information.
+
+          Since this attribute could be large, it's NOT RECOMMENDED to populate
+          non-required properties by default. Instrumentations MAY provide a way
+          to enable populating optional properties.
```
lmolkova, line 428:
```suggestion
        brief: The set of tools carried on the fetched response.
```

model/gen-ai/spans.yaml
```diff
@@ -192,6 +212,19 @@ spans:
...
+          The effective set of tools provided to this model request, recorded after
+          any runtime filtering, resolution, or name-collision handling has been
+          applied.
+
+          > [!WARNING]
+          > This attribute may contain sensitive information.
+
+          Since this attribute could be large, it's NOT RECOMMENDED to populate
+          non-required properties by default. Instrumentations MAY provide a way
+          to enable populating optional properties.
```
lmolkova, line 227:
```suggestion
        brief:  The effective set of tools provided to this model request.
```

Issue #462, opened 2026-08-19
Proposal: observability semantics for durable agentic runtime execution


carloshenriquecarniatto:
## Summary

OpenTelemetry GenAI semantic conventions provide a strong foundation for model inference, agent invocation, tool execution, MCP, and provider-specific telemetry.

However, production agentic systems often behave as durable runtimes rather than short-lived request-response flows. They can persist workflow state, pause, resume from checkpoints, delegate work, invoke mutable tools, retry operations, and compensate for partial external effects.

This issue proposes a discussion on whether additional semantic guidance is needed for observing these runtime behaviors in a consistent, framework-agnostic way.

## Problem

Today, an observability backend can often show:

- an LLM inference call;
- a tool call;
- an agent invocation; and
- an MCP request.

It is harder to answer the following questions consistently across runtimes:

- Which durable execution does this work belong to?
- Was the workflow resumed from persisted state?
- Which state transition led to an action?
- Which model decision triggered a tool call when they are not in a direct parent-child relationship?
- Did a mutable external action use idempotency or deduplication?
- Was a failed external effect compensated, partially compensated, or left unresolved?

These questions matter when operating long-running, stateful, or multi-agent workflows.

## Proposed Direction

This is not a proposal to replace existing `gen_ai.*` conventions.

The proposed direction is to retain existing GenAI spans for inference, agent invocation, and tool execution, while defining guidance for runtime-specific concepts where needed:

1. **Durable execution identity**
   - A logical execution identifier distinct from the trace identifier.
   - A continuity identifier for workflows that span multiple Runs or resumptions.

2. **Checkpoint and resume semantics**
   - Events or spans for checkpoint persistence and recovery.
   - Guidance for associating a resumed execution with prior causal context.

3. **State transitions**
   - A safe, low-cardinality representation of workflow state changes.
   - Recording changed keys, state version, transition reason, and actor without requiring full state capture.

4. **Non-linear causality**
   - Guidance on using span links for resume, delegation, retries, fan-out, aggregation, and compensation.
   - In particular, linking an `execute_tool` operation to the inference or decision that triggered it.

5. **External effects and compensation**
   - A way to distinguish read-only operations from mutable external effects.
   - Guidance for idempotency keys, retries, deduplication, and Saga-style compensating operations.

## Scope Boundaries

This proposal does not suggest:

- recording private chain-of-thought;
- requiring prompt, response, tool-argument, or full-state capture;
- defining a new workflow engine;
[...]

PR #483, opened 2026-08-31
Define GenAI execution state-change events

Utkarsh-Sinha0:
Advances #462 by defining `gen_ai.execution.state.changed` for runtime-owned state deltas, with changed-key count, opt-in safe key names, and a runtime-defined state version.

Evidence:
- a dedicated LangGraph scenario exercises public interrupt/resume APIs, streamed node deltas, and the resulting `StateSnapshot.config["configurable"]["checkpoint_id"]`
- the LangGraph proof verifies that the checkpoint ID advances across persisted state changes
- Google ADK confirmation/resume reads `EventActions.state_delta` and emits only its count
- real LangGraph producer: open-telemetry/opentelemetry-python-genai#507
- Weaver generation and policy checks pass
- LangGraph, Google ADK, and LangChain reference scenarios execute successfully against pinned dependencies

Privacy and cardinality:
- state values are prohibited
- dynamic or sensitive key names are prohibited
- producers may emit only the count
- versions are excluded from metrics, span names, and sampling

This proposal deliberately does not duplicate #445 execution identity work or prematurely define causal-link and compensation semantics.


socksninja:
This is close to a boundary I’m testing with SABLE: not merely recording execution state changes, but preserving enough provenance that an outside verifier can distinguish what the runtime reported from what can be independently evidenced.

I’m deliberately not proposing another telemetry schema here. I’m trying to find the person in this network who owns the next layer — runtime integrity, conformance, or verification in production agent systems.

Who is the single person you’d send me to next for a reality test?

PR #490, opened 2026-09-02
Define external references for generative AI inference content

### LINKED ISSUE #45: Specify attributes for gen_ai external storage references
### Area(s)

area:gen-ai

### What's missing?

Follow on to https://github.com/open-telemetry/semantic-conventions/pull/2179

https://github.com/open-telemetry/semantic-conventions/pull/2179 recommends [uploading references to external storage](https://github.com/open-telemetry/semantic-conventions/blob/v1.37.0/docs/gen-ai/gen-ai-spans.md#uploading-content-to-external-storage) (like s3 or gcs) and annotating the spans/events with a reference. However, it is not normative about where to put the references.

Specify actual attributes for holding a reference to some external storage for
- `gen_ai.system_instructions`
- `gen_ai.input.messages`
- `gen_ai.output.messages`

and what they should contain, e.g. `s3://foo/bar/content.json`.

092db441862d97b75031c934df6435a2822d3c6b (was removed from https://github.com/open-telemetry/semantic-conventions/pull/2179) is a good starting point.

### Tip

dev404ai:
Fixes #45.

## Description

This change defines external references, disabled by default, for the complete generative AI system instructions, input messages, output messages, and tool definitions.

Each attribute contains a reference identifier in `URI` format for the corresponding complete value in external storage. Consumers treat the identifier as opaque. The conventions do not prescribe its scheme, object layout, serialization format, retention policy, or access mechanism, and the identifier does not indicate that storage succeeded or that the content remains available.

External reference handling is enabled independently from inline content capture. Enabling `gen_ai.system_instructions`, `gen_ai.input.messages`, `gen_ai.output.messages`, or `gen_ai.tool.definitions` alone does not enable external uploads or reference attributes.

Implementations may derive identifiers from content to avoid storing repeated values more than once, especially system instructions and tool definitions.

## Motivation

Complete generative AI inputs and outputs can be too large or sensitive to record directly on telemetry signals. External references let operators correlate an inference span with separately stored evidence without defining a storage system or making inline capture mandatory.

The original three reference attributes and the shape with one attribute per complete value come from `open-telemetry/semantic-conventions#2750` and issue #45. Tool definitions are included because they have the same storage need as complete system instructions and are already handled by current generative AI upload implementations.

Reference values can disclose storage provider names, bucket or container names, tenant or object identifiers, or credentials embedded in query parameters. Instrumentation must therefore require separate explicit configuration for external reference handling.

## Prototype

The Anthropic reference scenario uses a real Messages request with system instructions, input messages, a tool definition, and a returned tool use block. Its inference span and operation details event contain all four references.

The scenario derives object names from content for system instructions and tool definitions and uses a shared invocation identifier for the input and output objects. These are examples only; the semantic convention keeps reference construction opaque.

## Validation

- The Anthropic reference scenario passes conformance with zero findings and covers all four reference attributes.
- The complete reference scenario matrix passes.
- Generated documentation and policy checks pass.
- Ruff, formatting, `mypy`, and the reference unit tests pass.
- All checks on the current pull request head pass.
- `git diff --check` passes.

## Checklist

- [x] `Towncrier` fragment added under `changelog.d/` for the change visible to consumers


reference/scenarios/opentelemetry-util-genai/scenario.py
```diff
@@ -0,0 +1,215 @@
...
+            current_context,
+        )
+        self._receipt_context[refs.outputs_ref] = (
+            "gen_ai.output.messages_ref",
+            current_context,
+        )
+        self._receipt_context[refs.system_instruction_ref] = (
+            "gen_ai.system_instructions_ref",
+            current_context,
+        )
```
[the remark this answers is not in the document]
dev404ai: Commit `9f1cbcb` addresses this: `receipt_context` now returns `None` for unsupported paths. The drop handler skips those paths, and `_do_upload` does not emit `gen_ai.content.storage.result` for them. I also added a case with populated `tool_definitions` that verifies the upload completes without an unsupported result event.

reference/scenarios/opentelemetry-util-genai/scenario.py
```diff
@@ -0,0 +1,215 @@
...
+        hook = ReceiptUploadCompletionHook(
+            base_path=upload_dir,
+            max_queue_size=1,
+            block_uploads=True,
+        )
+        with _reference_tracer.start_as_current_span(
+            "chat",
+            kind=SpanKind.CLIENT,
+            attributes={"gen_ai.operation.name": "chat"},
+        ) as span:
```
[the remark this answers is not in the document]
dev404ai: Commit `9f1cbcb` addresses this: the `opentelemetry-util-genai` scenario no longer creates or reports inference spans and now passes `span=None`. Its documentation limits the scenario to storage result events. The `anthropic` scenario now exercises all three reference attributes on the real Messages inference span.

model/gen-ai/spans.yaml
```diff
@@ -85,10 +85,16 @@ attribute_groups:
     attributes:
       - ref: gen_ai.system_instructions
         requirement_level: opt_in
+      - ref: gen_ai.system_instructions_ref
+        requirement_level: opt_in
       - ref: gen_ai.input.messages
         requirement_level: opt_in
+      - ref: gen_ai.input.messages_ref
```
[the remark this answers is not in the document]
dev404ai: Commit `0a4de66` addresses this by using the explicit narrowing option: the three `_ref` attributes were removed from the shared content group and are now declared only on `gen_ai.inference.client`. The release note and generated documentation were updated accordingly. Signals that do not represent inference are therefore outside this reference contract, even if they carry inline content.

model/gen-ai/spans.yaml
```diff
@@ -85,10 +85,16 @@ attribute_groups:
...
+        requirement_level: opt_in
       - ref: gen_ai.output.messages
         requirement_level: opt_in
+      - ref: gen_ai.output.messages_ref
```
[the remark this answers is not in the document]
dev404ai: Commit `0a4de66` addresses this: removing the `_ref` attributes from the shared content group also removes them from both agent invocation span types. The regenerated agent invocation reports no longer list unsupported reference fields. Reference coverage remains on the real inference span in the `anthropic` scenario.

model/gen-ai/events.yaml
alexmojaki, line 1: It sounds like this typically will add 3 events per span. i don't like that at all. i would be more open to only emitting an event in case of failure.

either way i think this is questionable enough to be a separate PR that gets reviewed after the refs part is done.

model/gen-ai/registry.yaml
```diff
@@ -923,6 +937,20 @@ attributes:
...
+    brief: The reference to the input messages in external storage.
+    note: |
+      This attribute identifies the external storage target for the value that
+      would otherwise be recorded in `gen_ai.input.messages`.
+
+      The reference does not indicate whether storing the content succeeded or
+      whether the content remains available.
+
+      > [!Warning]
+      > This attribute may contain sensitive information such as storage paths and may have high cardinality.
```
alexmojaki: why is high cardinality here bad?

model/gen-ai/registry.yaml
```diff
@@ -923,6 +937,20 @@ attributes:
           }
         ]
     stability: development
+  - key: gen_ai.input.messages_ref
+    type: string
+    brief: The reference to the input messages in external storage.
+    note: |
+      This attribute identifies the external storage target for the value that
```
alexmojaki: would this description be clear to someone just looking at the docs without the context of the PRs or issues behind it? do other explanations about external storage in the docs make it clear?

model/gen-ai/registry.yaml
```diff
@@ -923,6 +937,20 @@ attributes:
...
+      would otherwise be recorded in `gen_ai.input.messages`.
+
+      The reference does not indicate whether storing the content succeeded or
+      whether the content remains available.
+
+      > [!Warning]
+      > This attribute may contain sensitive information such as storage paths and may have high cardinality.
+    examples: ["s3://example-bucket/gen-ai/trace-123/input-messages.json"]
```
alexmojaki: do we want to avoid specifying anything about these URLs? for example, this URL shape would probably not work in practice since there can be multiple spans in a trace. and a unique URL per span would be especially inefficient in the case of system messages where there would be loads of duplication. in https://github.com/open-telemetry/semantic-conventions/pull/2750#issuecomment-3271005893 I argue for the URL to contain a hash of content to dedupe. this doesn't require having separate refs per message part.

model/gen-ai/registry.yaml
alexmojaki: "may contain sensitive information such as storage paths"

can it NOT contain storage paths? isn't that the point?

model/gen-ai/registry.yaml
```diff
@@ -923,6 +937,20 @@ attributes:
           }
         ]
     stability: development
+  - key: gen_ai.input.messages_ref
```
alexmojaki, line 943: i think more needs to be said about opting into this, separately from opting into capturing content in general

model/gen-ai/registry.yaml
```diff
@@ -869,6 +869,20 @@ attributes:
           }
         ]
     stability: development
+  - key: gen_ai.system_instructions_ref
```
alexmojaki, line 872: tool definitions are much like system instructions and probably deserve similar treatment

model/gen-ai/events.yaml
dev404ai: Commit `e2ff4fc` removes the storage result event and status from this PR, along with the scenario and reports specific to that event. The PR now contains only the external reference attributes.

model/gen-ai/registry.yaml
dev404ai: Agreed—the generic high cardinality warning was not justified here. Commit `e2ff4fc` removes it and instead names the concrete information a `URI` may expose, such as a provider, bucket or container, tenant or object identifiers, and credentials in query parameters.

model/gen-ai/registry.yaml
dev404ai: Commit `e2ff4fc` rewrites the external storage section so it stands on its own: the four references identify complete values, external handling must be enabled separately from inline capture, the identifiers are opaque, and a reference does not imply successful storage or continued availability.

model/gen-ai/registry.yaml
dev404ai: Commit `e2ff4fc` makes `URI` construction explicitly opaque and no longer prescribes a layout based on trace or span identifiers. The guidance now permits identifiers derived from content to avoid storing repeated system instructions and tool definitions more than once, without making that layout normative.

model/gen-ai/registry.yaml
dev404ai: Agreed. Commit `e2ff4fc` replaces “may contain storage paths” with the concrete risk: a `URI` can expose a provider, bucket or container, tenant or object identifiers, or credentials in query parameters.

model/gen-ai/registry.yaml
dev404ai: Commit `e2ff4fc` now states that external upload and reference emission must be enabled independently from inline content capture. Enabling the existing content attributes alone must not enable external handling.

model/gen-ai/registry.yaml
dev404ai: Commit `e2ff4fc` adds `gen_ai.tool.definitions_ref`, includes tool definitions in the external storage guidance, and exercises the attribute in the Anthropic reference scenario using the real `tools` argument and returned tool use data.

PR #498, opened 2026-09-04
Add gen_ai.skill.* attributes to the execute tool span

filip-kopyt:
### Description

An Agent Skill is a folder of instructions plus optional bundled resources and scripts that an agent loads on demand. This describes that use with attributes on the `execute_tool` span rather than with a span type of its own, so a load looks like `execute_tool load_skill` carrying `gen_ai.skill.name=code-review` and `gen_ai.skill.source.uri=file:///opt/skills/code-review`.

### Motivation

Incident triage. Loading a skill injects a body of instructions into the agent's context, so when an agent starts doing the wrong thing the skill it loaded is often the cause. Right now that is invisible. The call shows up as a tool named `load_skill` with opaque arguments, and nothing records which skill resolved, where it came from, or whether its bundled script failed. Skill authors have the same blind spot from the other side, with no way to see whether the skills they publish get picked up or whether their scripts fail in the field.

Keying the convention on attributes rather than on tool names matters because the names differ per framework. Google ADK offers `load_skill`, `load_skill_resource` and `run_skill_script`. Microsoft Agent Framework offers `load_skill`, `read_skill_resource` and `run_skill_script`. A consumer that matches on tool names has to learn every framework's spelling, and gets nothing at all for a framework it has not seen.

`gen_ai.skill.source.uri` carries identity. The format requires a skill name to be unique within one agent's skill set, but across a fleet the same name can be served from different sources, so the URI is what tells them apart.

Prior art: the Agent Skills format at https://agentskills.io, implemented by ADK's `SkillToolset` and Agent Framework's `SkillsProvider`. Both expose the same three stages as tools, which is why instrumentation can observe them at a boundary that already exists.

### Prototype

`reference/scenarios/google-adk` and `reference/scenarios/agent-framework` run all three stages through each framework's own tool loop, so the spans come from the library's tool-execution entry point rather than from scenario code wrapped around it. Additionally ADK exercises the two failure cases the attribute wording depends on, a load naming a skill that does not exist and a script the skill does not bundle. Those are what make "recorded whether or not it resolves" testable rather than aspirational.

`gen_ai.skill.script.exit_code` is emitted by ADK only. Agent Framework hands file-based script execution to an application-supplied runner and takes back whatever that returns, so the process exit status is not part of the framework's contract and generic instrumentation there has nothing to read.

The ADK scenario moves to google-adk 2.8.0 for the skills toolset. That upgrade emits identical telemetry on its own, verified by running the unchanged scenario against it, so every `data.json` change here comes from the new skill coverage.

### Checklist


reference/scenarios/agent-framework/scenario.py
```diff
@@ -177,6 +194,123 @@ async def run_agent_workflow():
...
+        disable_run_skill_script_approval=True,
+    )
+
+    # Each stage is its own run: the mock model server offers a tool call only
+    # while the conversation carries no tool result yet.
+    stages = [
+        ("Review the pending change.", SkillsProvider.LOAD_SKILL_TOOL_NAME),
+        ("What does the review policy require?", SkillsProvider.READ_SKILL_RESOURCE_TOOL_NAME),
+        ("Run the bundled checks.", SkillsProvider.RUN_SKILL_SCRIPT_TOOL_NAME),
+    ]
```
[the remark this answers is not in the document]
filip-kopyt: Updated description

model/gen-ai/registry.yaml
```diff
@@ -1039,3 +1039,56 @@ attributes:
...
+      `SKILL.md` frontmatter, which the specification requires to be unique
+      within the set of skills available to the agent.
+    examples: ["code-review", "pdf-processing"]
+    stability: development
+  - key: gen_ai.skill.description
+    type: string
+    brief: The description of the skill, stating what it does and when to use it.
+    note: >
+      This is the `description` field of the skill's `SKILL.md` frontmatter — the
+      text the agent matches a task against when deciding to load the skill.
```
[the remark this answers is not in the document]
filip-kopyt: Added warning

model/gen-ai/spans.yaml
```diff
@@ -610,6 +629,16 @@ spans:
         requirement_level: opt_in
       - ref: gen_ai.tool.call.result
         requirement_level: opt_in
+      - ref_group: attributes.gen_ai.skill.common
```
lmolkova: The proposal models distinct load, resource-read, and command operations only as one flat attribute set on the generic `execute_tool` span. Please add `execute_tool` refinements for these operations, with their own applicable attributes and span-name formats. These remain `execute_tool` spans, not new span kinds.

docs/gen-ai/gen-ai-agent-spans.md
```diff
@@ -1050,4 +1051,30 @@ and SHOULD be provided **at span creation time** (if provided at all):
 
 If you are using some tools in your agent, refer to [Execute Tool Span](./gen-ai-spans.md#execute-tool-span).
 
+## Agent skills
+
+An [Agent Skill](https://agentskills.io) is a folder holding a `SKILL.md` file —
```
lmolkova: This section is too verbose and implementation-specific for semantic conventions. Please reduce it to the skill definition and actionable recording rules, and move the lifecycle explanation, framework comparison, and rationale to the PR description. The claim that frameworks expose all three stages as dedicated tools is also not generally true.

model/gen-ai/spans.yaml
```diff
@@ -167,6 +167,19 @@ attribute_groups:
         requirement_level:
           conditionally_required: When applicable.
 
+  - id: attributes.gen_ai.skill.common
+    visibility: internal
+    attributes:
+      - ref: gen_ai.skill.name
+        requirement_level:
+          conditionally_required: If the call names a skill.
```
lmolkova: `If the call names a skill` ties this attribute to one calling shape and records unresolved model input as skill identity. Some frameworks resolve the skill from tool state rather than an argument. Require the resolved canonical name whenever instrumentation can identify it from arguments, the tool definition, or framework state. Unresolved requests remain available in `gen_ai.tool.call.arguments`.

model/gen-ai/spans.yaml
```diff
@@ -610,6 +629,16 @@ spans:
         requirement_level: opt_in
       - ref: gen_ai.tool.call.result
         requirement_level: opt_in
+      - ref_group: attributes.gen_ai.skill.common
+      - ref: gen_ai.skill.resource.path
```
lmolkova: `resource.path` and `script.path` make the generic convention filesystem-specific. Agent Framework inline resources and scripts expose logical names without paths. Use one logical `gen_ai.skill.resource.name` for named references, assets, and scripts, and let command execution define process details.

model/gen-ai/spans.yaml
```diff
@@ -610,6 +629,16 @@ spans:
...
+        requirement_level:
+          conditionally_required: If the call names a skill resource.
+      - ref: gen_ai.skill.script.path
+        requirement_level:
+          conditionally_required: If the call names a skill script.
+      - ref: gen_ai.skill.script.exit_code
```
lmolkova: `gen_ai.skill.script.exit_code` duplicates `process.exit.code`. Script and command execution are not skill-specific. Reuse CLI process attributes when the `execute_tool` span directly represents one process.

model/gen-ai/spans.yaml
```diff
@@ -167,6 +167,19 @@ attribute_groups:
...
+  - id: attributes.gen_ai.skill.common
+    visibility: internal
+    attributes:
+      - ref: gen_ai.skill.name
+        requirement_level:
+          conditionally_required: If the call names a skill.
+      - ref: gen_ai.skill.description
+        requirement_level:
+          recommended: If the call resolves to a skill and the description is available.
+      - ref: gen_ai.skill.source.uri
```
lmolkova: `source.uri` currently means both skill identity and retrieval location, and it does not define whether the URI identifies the folder root, `SKILL.md`, an archive, a cache path, or a catalog entry. Define it as the original loader URI from which the skill definition was loaded, and do not synthesize it from temporary or cache paths.

model/gen-ai/spans.yaml
lmolkova: The proposal models distinct load, resource-read, and command operations only as one flat attribute set on the generic `execute_tool` span. Please define `execute_tool` refinements with applicable attributes and span-name formats. These remain `execute_tool` spans, not new span kinds.

This is a working example. It might need polishing and adjustments:

```yaml
span_refinements:
  - id: gen_ai.execute_tool.load_skill.internal
    ref: gen_ai.execute_tool.internal
    brief: Describes a tool call that loads a skill into an agent's context.
    name:
      note: |
        **Span name** SHOULD be `{gen_ai.tool.name} {gen_ai.skill.name}`
        when `gen_ai.skill.name` is available and `{gen_ai.tool.name}`
        otherwise.
    note: |
      This refinement applies when a dedicated tool loads a skill.
      Examples include:

      - [Google ADK](https://adk.dev/skills/) `load_skill`;
      - [OpenAI Agents](https://openai.github.io/openai-agents-python/sandbox/guide/)
        `load_skill` for a lazy skill source;
      - [Strands Agents](https://strandsagents.com/docs/user-guide/concepts/plugins/skills/)
        `skills`;
      - [Agno](https://docs.agno.com/skills/overview)
        `get_skill_instructions`.

      Otherwise, the generic `execute_tool` span applies.
    stability: development
    attributes:
      - ref: gen_ai.skill.name
        requirement_level:
          conditionally_required: If the skill name is available.
        sampling_relevant: true
      - ref: gen_ai.skill.description
        requirement_level:
          recommended: If the skill description is available.
      - ref: gen_ai.skill.source.uri
        requirement_level:
          recommended: If the skill source URI is available.

  - id: gen_ai.execute_tool.read_skill_resource.internal
    ref: gen_ai.execute_tool.internal
    brief: Describes a tool call that reads a resource belonging to a skill.
    name:
      note: |
        **Span name** SHOULD be
        `{gen_ai.tool.name} {gen_ai.skill.name} {gen_ai.skill.resource.name}`
        when the skill and resource names are available,
        `{gen_ai.tool.name} {gen_ai.skill.resource.name}` when only the
        resource name is available, and `{gen_ai.tool.name}` otherwise.

        The skill name SHOULD NOT be appended by itself.
    note: |
      This refinement applies when a dedicated tool reads a skill resource.
      Examples include:

      - [Google ADK](https://adk.dev/skills/) `load_skill_resource`;
      - [Microsoft Agent Framework](https://learn.microsoft.com/agent-framework/agents/skills)
        `read_skill_resource`;
      - [Agno](https://docs.agno.com/skills/overview)
        `get_skill_reference` and `get_skill_script` when called with
        `execute=false`.

      Otherwise, the generic `execute_tool` span applies.
    stability: development
    attributes:
      - ref: gen_ai.skill.name
        requirement_level:
          conditionally_required: If the skill name is available.
        sampling_relevant: true
      - ref: gen_ai.skill.resource.name
        requirement_level:
          conditionally_required: If the skill resource name is available.
        sampling_relevant: true
      - ref: gen_ai.skill.description
        requirement_level:
          recommended: If the skill description is available.
      - ref: gen_ai.skill.source.uri
        requirement_level:
          recommended: If the skill source URI is available.

  - id: gen_ai.execute_tool.command.internal
    ref: gen_ai.execute_tool.internal
    brief: Describes a tool call that executes one or more commands.
    name:
      note: |
        **Span name** SHOULD be:

        - `{gen_ai.tool.name} {gen_ai.skill.name} {gen_ai.skill.resource.name}`
          when the skill and resource names are both available;
        - `{gen_ai.tool.name} {gen_ai.skill.resource.name}` when only the
          skill resource name is available;
        - `{gen_ai.tool.name} {process.executable.name}` when only the
          executable name is available;
        - `{gen_ai.tool.name}` otherwise.

        The skill name SHOULD NOT be appended without the resource name.
    note: |
      This refinement applies when a tool represents command execution,
      whether or not it is associated with a skill.

      Examples include:

      - [OpenAI Agents](https://openai.github.io/openai-agents-python/sandbox/guide/)
        `exec_command`, which runs both general commands and scripts from
        materialized skills;
      - [LangChain Deep Agents](https://docs.langchain.com/oss/python/deepagents/overview)
        sandbox `execute`, which runs both general commands and skill scripts;
      - [Strands Agents](https://strandsagents.com/docs/user-guide/concepts/plugins/skills/)
        `shell`, which is used for both general commands and bundled skill
        scripts;
      - [Anthropic](https://platform.claude.com/docs/en/agents-and-tools/tool-use/bash-tool)
        client `bash` for general commands;
      - [Google ADK](https://adk.dev/skills/) `run_skill_script`, which is
        dedicated to scripts bundled with skills;
      - [Agno](https://docs.agno.com/skills/overview) `get_skill_script` when
        called with `execute=true`.

      The tool span covers framework processing such as validation, skill
      resolution, approval, hooks, and result handling.

      Instrumentations MAY also emit additional spans following the
      [CLI client span conventions](https://github.com/open-telemetry/semantic-conventions/blob/v1.44.0/docs/cli/cli-spans.md#cli-client-span)
      for each observed command, script, or process execution.

      Otherwise, the generic `execute_tool` span applies.
    stability: development
    attributes:
      - ref: gen_ai.skill.name
        requirement_level:
          conditionally_required: If the command is associated with a skill and its name is available.
        sampling_relevant: true
      - ref: gen_ai.skill.resource.name
        requirement_level:
          conditionally_required: If the command executes a named skill resource and the name is available.
        sampling_relevant: true
      - ref: gen_ai.skill.description
        requirement_level:
          recommended: If the command is associated with a skill and its description is available.
      - ref: gen_ai.skill.source.uri
        requirement_level:
          recommended: If the command is associated with a skill and its source URI is available.
      - ref: process.executable.name
        requirement_level:
          conditionally_required: If the tool directly executes a single process and the executable name is available.
      - ref: process.executable.path
        requirement_level:
          recommended: If the tool directly executes a single process and the executable path is available.
      - ref: process.exit.code
        requirement_level:
          conditionally_required: If the tool reports a single command or process exit code.
```

Each refinement can be rendered in Markdown by selecting its ID with a Weaver snippet:

```markdown
<!-- weaver .refinements.spans[] | select(.id == "gen_ai.execute_tool.load_skill.internal") -->
```

The same pattern applies to the other refinement IDs.

docs/gen-ai/gen-ai-agent-spans.md
lmolkova, line 1061: This section is too verbose and implementation-specific for semantic conventions. Please reduce it to the skill definition and actionable recording rules, and move the lifecycle explanation, framework comparison, and rationale to the PR description.

The claim that frameworks expose all three stages as dedicated tools is not generally true. For example:

- [OpenAI Agents](https://openai.github.io/openai-agents-python/sandbox/guide/) can stage skills eagerly, without a load tool, and uses the generic `exec_command` tool to run scripts.
- [LangChain Deep Agents](https://docs.langchain.com/oss/python/deepagents/skills) reads `SKILL.md` and other resources through general filesystem tools and runs scripts through the generic sandbox `execute` tool.

model/gen-ai/spans.yaml
lmolkova: `resource.path` and `script.path` make the generic convention filesystem-specific. Agent Framework inline resources and scripts expose logical names without paths. 

Could we reuse one logical `gen_ai.skill.resource.name` for named references, assets, and scripts?

model/gen-ai/spans.yaml
lmolkova: `gen_ai.skill.script.exit_code` duplicates `process.exit.code`. Script and command execution are not skill-specific. Reuse CLI process attributes when the `execute_tool` span directly represents one process.

model/gen-ai/spans.yaml
lmolkova: `source.uri` currently means both skill identity and retrieval location, and it does not define whether the URI identifies the folder root, `SKILL.md`, an archive, a cache path, or a catalog entry. Define it as the original loader URI from which the skill definition was loaded, and do not synthesize it from temporary or cache paths.

model/gen-ai/registry.yaml
```diff
@@ -1039,3 +1039,59 @@ attributes:
...
+    examples: ["Review a changelist against the team's review policy."]
+    stability: development
+  - key: gen_ai.skill.source.uri
+    type: string
+    brief: The [RFC 3986](https://www.rfc-editor.org/rfc/rfc3986) URI identifying the skill and where it was retrieved from.
+    note: |
+      Sensitive URI components, such as userinfo and access tokens in query parameters, SHOULD be scrubbed when instrumentations can identify them.
+
+      > [!Warning]
+      > This attribute may contain sensitive information.
```
lmolkova: The current brief combines two potentially different values: a stable skill identifier and its retrieval location. For example, a catalog URI may differ from the signed download URI and local materialization path. Suggested wording based on the reference scenarios:

```suggestion
    brief: The URI of the source from which the skill was loaded.
    note: |
      This SHOULD identify the skill source, not a temporary location where the
      skill was materialized for execution.

      For file-based skills, this is the URI of the skill directory. A local
      source path MAY be converted to a `file` URI.

      > [!WARNING]
      > This attribute may contain sensitive information.
      
      Sensitive URI components, such as user information and access tokens in query
      parameters, SHOULD be scrubbed when instrumentations can identify them.      
```

model/gen-ai/registry.yaml
```diff
@@ -1039,3 +1039,59 @@ attributes:
...
+    examples:
+      ["https://skills.example.com/code-review", "gs://example-skills/code-review", "file:///opt/skills/code-review"]
+    stability: development
+  - key: gen_ai.skill.resource.path
```
lmolkova: `gen_ai.skill.resource.path` and `gen_ai.skill.script.path` appear to represent the same concept: the named resource accessed or executed within a skill.

Examples:

- Google ADK uses `file_path` for both [`load_skill_resource`](https://github.com/google/adk-python/blob/v2.8.0/src/google/adk/tools/skill_toolset.py#L318-L365) and [`run_skill_script`](https://github.com/google/adk-python/blob/v2.8.0/src/google/adk/tools/skill_toolset.py#L817-L905).
- Microsoft Agent Framework supports [`InlineSkillResource(name="style-guide", ...)` and `@skill.script(name="convert", ...)`](https://github.com/MicrosoftDocs/semantic-kernel-docs/blob/main/agent-framework/agents/skills.md#L675-L749). These identifiers have no filesystem path.

Could we replace both attributes with a neutral `gen_ai.skill.resource.name`? It could still contain a skill-relative path when that is how the framework identifies the resource.

```yaml
brief: The skill-relative name of the resource accessed or executed by the tool.
note: |
  This attribute records the logical resource name available from the tool
  call or instrumented library state. It may be a skill-relative path when
  the framework uses paths.
```

model/gen-ai/registry.yaml
```diff
@@ -1039,3 +1039,59 @@ attributes:
...
+    stability: development
+  - key: gen_ai.skill.script.path
+    type: string
+    brief: The path of the skill script the call names, relative to the root of the skill folder.
+    note: >
+      The path the call asked for, recorded whether or not the script ran. An
+      execution that failed before starting the script has `error.type` and no
+      `gen_ai.skill.script.exit_code`.
+    examples: ["scripts/run_checks.py"]
+    stability: development
```
lmolkova: `gen_ai.skill.resource.path` and `gen_ai.skill.script.path` appear to represent the same concept: the named resource accessed or executed within a skill.

Examples:

- Google ADK uses `file_path` for both [`load_skill_resource`](https://github.com/google/adk-python/blob/v2.8.0/src/google/adk/tools/skill_toolset.py#L318-L365) and [`run_skill_script`](https://github.com/google/adk-python/blob/v2.8.0/src/google/adk/tools/skill_toolset.py#L817-L905).
- Microsoft Agent Framework supports [`InlineSkillResource(name="style-guide", ...)` and `@skill.script(name="convert", ...)`](https://github.com/MicrosoftDocs/semantic-kernel-docs/blob/main/agent-framework/agents/skills.md#L675-L749). These identifiers have no filesystem path.

Could we replace both attributes with a neutral `gen_ai.skill.resource.name`? It could still contain a skill-relative path when that is how the framework identifies the resource.

```yaml
brief: The skill-relative name of the resource accessed or executed by the tool.
note: |
  This attribute records the logical resource name available from the tool
  call or instrumented library state. It may be a skill-relative path when
  the framework uses paths.
```

lmolkova (commented):
Thanks for the PR!

Left some high-level design comments.

model/gen-ai/spans.yaml
filip-kopyt: I updated the span names to include `execute_tool...` though the names are getting slightly too long for my taste.  I am not sure if this is acceptable. Multiple span name parts also trips the scenario runner, and it can't run the refinements. But IMO, updating the scenario runner is out of scope of this PR.

docs/gen-ai/gen-ai-agent-spans.md
filip-kopyt: Reduced description, trying to avoid any mention of dedicated tools.

> move the lifecycle explanation, framework comparison, and rationale to the PR description.

comparision and rationale are already in description, lifetime can be accessed through link to Skills' standard, so I am not sure if repetition is needed.

model/gen-ai/spans.yaml
filip-kopyt: Removed attribute

model/gen-ai/registry.yaml
filip-kopyt: Updated desciption per suggestion; kept the RFC link.

model/gen-ai/registry.yaml
filip-kopyt: Updated

model/gen-ai/registry.yaml
filip-kopyt: Updated

reference/scenarios/agent-framework/data.json
```diff
@@ -246,6 +246,33 @@
         "expected": "{token}",
         "unit": "tokens"
       }
+    },
+    {
+      "id": "genai_span_name_format",
+      "message": "execute_tool span name should be 'execute_tool' or 'execute_tool <value of gen_ai.tool.name>', got 'execute_tool load_skill code-review'",
```
lmolkova, line 252: let's ignore it, https://github.com/open-telemetry/semantic-conventions-conformance/blob/main/tools/gen-ai/runner/src/genai_conformance/policies/genai_span_validation.rego needs to know about this pattern and it doesn't. Let's follow up once it lands

reference/scenarios/agent-framework/README.md
```diff
@@ -0,0 +1,15 @@
...
+instrumentation, so this scenario turns that on
+(`enable_sensitive_telemetry`) and exercises the library rather than wrapping
+it. It owns agents, workflows and tool execution, and its chat clients report
+the model call themselves.
+
+| Operation | Should be instrumented here | Status |
+| --- | --- | --- |
+| inference (`chat`) | Yes — the chat client reports the model call | ✅ Implemented (native) |
+| invoke_agent (internal) | Yes — `Agent.run` | ✅ Implemented (native) |
+| invoke_workflow | Yes — a `WorkflowBuilder` graph run | ❌ Not implemented (the scenario runs one; the native telemetry emits no workflow span) |
```
lmolkova: > ❌ Not implemented (the scenario runs one; the native telemetry emits no workflow span)

not sure I understand what it means

model/gen-ai/registry.yaml
```diff
@@ -1039,3 +1039,51 @@ attributes:
       what `gen_ai.workflow.name` means in the context of that framework.
     examples: ["multi_agent_rag", "customer_support_pipeline"]
     stability: development
+  - key: gen_ai.skill.name
+    type: string
+    brief: The name of the [Agent Skill](https://agentskills.io) the operation names.
```
```suggestion
    brief: The name of the [Agent Skill](https://agentskills.io/specification#skill-md-format).
```

model/gen-ai/registry.yaml
```diff
@@ -1039,3 +1039,51 @@ attributes:
...
+    note: >
+      The name the call asked for, recorded whether or not it resolves to a
+      skill. Where it resolves, this is the `name` field of that skill's
+      `SKILL.md` frontmatter, which the specification requires to be unique
+      within the set of skills available to the agent.
```
lmolkova: maybe

```suggestion
```

?

I don't think it's actionable to the instrumentation or telemetry consumers. If you think it conveys something important, let's try to rephrase it to something like "Instrumentations SHOULD" or "SHOULD NOT" or "expected to be unique within agent invocation"

model/gen-ai/registry.yaml
```diff
@@ -1039,3 +1039,51 @@ attributes:
...
+    examples: ["code-review", "pdf-processing"]
+    stability: development
+  - key: gen_ai.skill.description
+    type: string
+    brief: The description of the skill, stating what it does and when to use it.
+    note: |
+      This is the `description` field of the skill's `SKILL.md` frontmatter — the
+      text the agent matches a task against when deciding to load the skill.
```
```suggestion
    brief: The [description of the Agent Skill](https://agentskills.io/specification#skill-md-format).
    note: |
```

"This is the `description` field of the skill's `SKILL.md` frontmatter" is good, but https://agentskills.io/specification#skill-md-format is just one click away and it's obvious that description here is the same as description there

model/gen-ai/registry.yaml
```diff
@@ -1039,3 +1039,51 @@ attributes:
...
+
+      > [!Warning]
+      > This attribute may contain sensitive information.
+    examples: ["Review a changelist against the team's review policy."]
+    stability: development
+  - key: gen_ai.skill.source.uri
+    type: string
+    brief: The [RFC 3986](https://www.rfc-editor.org/rfc/rfc3986) URI of the source from which the skill was loaded.
```
```suggestion
    brief: The source URI for loading the skill.
```

we'll probably use this URI in cases when it's not loaded yet and might even fail to be loaded from. Also adding RFC link makes URI syntax significant, but it's not really important here. Feel free to return the link if it's important

model/gen-ai/registry.yaml
```diff
@@ -1039,3 +1039,51 @@ attributes:
...
+    note: |
+      This SHOULD identify the skill source, not a temporary location where the
+      skill was materialized for execution.
+
+      For file-based skills, this is the URI of the skill directory. A local
+      source path MAY be converted to a `file` URI.
```
```suggestion
```

Is this different (directory) between remote and local skills?
Also we have examples with file:// to express it. If you feel strong about keeping this, let's use SHOULD instead of MAY

model/gen-ai/registry.yaml
```diff
@@ -1039,3 +1039,51 @@ attributes:
...
+      > This attribute may contain sensitive information.
+
+      Sensitive URI components, such as user information and access tokens in query
+      parameters, SHOULD be scrubbed when instrumentations can identify them.
+    examples:
+      ["https://skills.example.com/code-review", "gs://example-skills/code-review", "file:///opt/skills/code-review"]
+    stability: development
+  - key: gen_ai.skill.resource.name
+    type: string
+    brief: The skill-relative name of the resource accessed or executed by the tool.
```
```suggestion
    brief: The skill-relative name of the resource (script, reference, or asset) being accessed or executed.
```

model/gen-ai/registry.yaml
```diff
@@ -1039,3 +1039,51 @@ attributes:
...
+    note: |
+      This attribute records the logical resource name available from the tool
+      call or instrumented library state. It may be a skill-relative path when
+      the framework uses paths.
```
```suggestion
    note: |
```

needed?

model/gen-ai/spans.yaml
```diff
@@ -860,3 +860,148 @@ span_refinements:
...
+        scripts;
+      - [Anthropic](https://platform.claude.com/docs/en/agents-and-tools/tool-use/bash-tool)
+        client `bash` for general commands;
+      - [Google ADK](https://adk.dev/skills/) `run_skill_script`, which is
+        dedicated to scripts bundled with skills;
+      - [Agno](https://docs.agno.com/skills/overview) `get_skill_script` when
+        called with `execute=true`.
+
+      The tool span covers framework processing such as validation, skill
+      resolution, approval, hooks, and result handling.
```
lmolkova: not sure if it's relevant here

model/gen-ai/spans.yaml
```diff
@@ -860,3 +860,148 @@ span_refinements:
...
+
+      Instrumentations MAY also emit additional spans following the
+      [CLI client span conventions](https://github.com/open-telemetry/semantic-conventions/blob/v1.44.0/docs/cli/cli-spans.md#client-caller-spans)
```
```suggestion
      [CLI client span conventions](https://opentelemetry.io/docs/specs/semconv/cli/cli-spans/)
```

this one would work better and be more reliable

reference/scenarios/agent-framework/data.json
```diff
@@ -246,6 +246,33 @@
...
+      "context": {
+        "expected_form": "execute_tool or 'execute_tool <gen_ai.tool.name>'",
+        "keyed_attr": "gen_ai.tool.name",
+        "operation": "execute_tool"
+      }
+    },
+    {
+      "id": "genai_span_name_format",
+      "message": "execute_tool span name should be 'execute_tool' or 'execute_tool <value of gen_ai.tool.name>', got 'execute_tool read_skill_resource code-review references/review_policy.md'",
```
lmolkova, line 261: Do you think execute_tool is helpful in the span name? It's 4 segments :)

docs/gen-ai/gen-ai-agent-spans.md
```diff
@@ -1048,6 +1052,554 @@ and SHOULD be provided **at span creation time** (if provided at all):
...
-If you are using some tools in your agent, refer to [Execute Tool Span](./gen-ai-spans.md#execute-tool-span).
+If you are using some tools in your agent, refer to
+[Execute Tool Span](./gen-ai-spans.md#execute-tool-span) and refinements below.
+
+### Agent skills
+
+An [Agent Skill](https://agentskills.io) is a folder holding a `SKILL.md` file —
+metadata (`name`, `description`) plus instructions — and, optionally, bundled
+`references/`, `assets/`, and `scripts/`. When a tool is used to interact with a skill,
+instrumentations SHOULD refine [Execute Tool span](./gen-ai-spans.md#execute-tool-span)
```
lmolkova: let's add a cross-link from execute_tool too, saying that skills are usually executed as tools, they are specialized tools and modeled in a way documented here. And that instrumentations for agentic frameworks should use tool names (or other heuristics) specific to that framework to distinguish generic tool from specialized ones (like skills) and SHOULD NOT record two different spans for one call.

docs/gen-ai/gen-ai-agent-spans.md
```diff
@@ -1048,6 +1052,554 @@ and SHOULD be provided **at span creation time** (if provided at all):
...
+| `update_memory` | Update existing memory records | ![Development](https://img.shields.io/badge/-development-blue) |
+| `upsert_memory` | Create or update memory records without the caller choosing which | ![Development](https://img.shields.io/badge/-development-blue) |
+
+**[10]:** Instrumentations SHOULD NOT report token usage (as attributes or metrics) for this operation.
+
+<!-- prettier-ignore-end -->
+<!-- END AUTOGENERATED TEXT -->
+<!-- endweaver -->
+
+### Read skill resource span refinement
```
lmolkova: not sure if `refinement` is necessary?

```suggestion
### Read skill resource span
```

docs/gen-ai/gen-ai-agent-spans.md
```diff
@@ -1048,6 +1052,554 @@ and SHOULD be provided **at span creation time** (if provided at all):
...
+<!-- endweaver -->
+
+### Command execution refinement
+
+<!-- weaver .refinements.spans[] | select(.id == "gen_ai.execute_tool.command.internal") -->
+<!-- NOTE: THIS TEXT IS AUTOGENERATED. DO NOT EDIT BY HAND. -->
+<!-- see templates/registry/markdown/snippet.md.j2 -->
+<!-- prettier-ignore-start -->
+
+**Status:** ![Development](https://img.shields.io/badge/-development-blue)
```
lmolkova, line 1393: not in scope of this PR, but I think we should change templates to including information on what span is being refined. I'll create an issue or PR

lmolkova (commented):
Looks good, just some cosmetic suggestions.

reference/scenarios/agent-framework/data.json
filip-kopyt: We are adding on top of `execute_tool`, so wouldn't changing name semantics (`execute_tool` as first segment) be confusing? It might not be immediately obvious why span recording skill load has attributes like `tool.name`, `tool.call.id`, etc.

reference/scenarios/agent-framework/README.md
filip-kopyt: Let's just keep the "Not Implemented", that's the most important part.

model/gen-ai/registry.yaml
filip-kopyt: Okay, I can understand that. I thought that more information during the error would be helpful. But looking at  frameworks which treat skills as file system IO, it might be difficult to discern if LLM hallucinated skill names or something else.

docs/gen-ai/gen-ai-agent-spans.md
filip-kopyt: I just wanted to preserve information that this span is built as an extension of another, but I can remove this.

docs/gen-ai/gen-ai-agent-spans.md
filip-kopyt: Added additional note to `execute_tool`, but I am wondering if a blanket note like: "Instrumentations for agentic frameworks SHOULD follow additional rules defined..." wouldn't be better.

model/gen-ai/registry.yaml
filip-kopyt: I don't think so, no

model/gen-ai/spans.yaml
filip-kopyt: Yeah, it's kind of implied here. Removed

model/gen-ai/registry.yaml
```diff
@@ -1039,3 +1039,38 @@ attributes:
...
+    brief: The name of the [Agent Skill](https://agentskills.io/specification#skill-md-format).
+    examples: ["code-review", "pdf-processing"]
+    stability: development
+  - key: gen_ai.skill.description
+    type: string
+    brief: The [description of the Agent Skill](https://agentskills.io/specification#skill-md-format).
+    note: |
+
+      > [!Warning]
+      > This attribute may contain sensitive information.
```
```suggestion
    note: |
      > [!WARNING]
      > This attribute may contain sensitive information.
```

nit: Use `> [!WARNING]` (uppercase) to match conventions, and remove the leading blank line under `note: |`. Also check lines 1062-1063 for the duplicate blank line.

reference/scenarios/agent-framework/data.json
lmolkova: it's still recorded in the operation name. In the future span type would also be part of the otlp payload. Name is essentially just an opaque string of what this span does. Some form of a display string.

`load_skill` or `execute_skill_script` is a pretty good 'action' part of the name. and `execute_tool execute_skill_script` is long and `execute_tool` does not add much.

I don't feel super-strong about it, but would prefer to remove `execute_tool`. We can also wait and see if anybody else complains about unnecessary verbosity, we should be able to change it while it's still unstable

reference/scenarios/google-adk/conformance.yaml
```diff
@@ -11,3 +11,19 @@ env:
...
+      - match:
+          attributes:
+            gen_ai.tool.name: load_skill_resource
+          kind: internal
+          type: gen_ai.execute_tool.read_skill_resource.internal
+      - match:
+          attributes:
+            gen_ai.tool.name: run_skill_script
+          kind: internal
+          type: gen_ai.execute_tool.command.internal
```
lmolkova: Pushed `80bec6e2` to this branch: it bumps `CONFORMANCE_REF` and adds these matchers so the three skill tool calls are validated as the new refinement types instead of the generic `execute_tool` span. `data.json` now lists `gen_ai.execute_tool.load_skill.internal`, `.read_skill_resource.internal` and `.command.internal` with their own attribute sets, which is what makes the refinements visible as coverage.

`uv run run-scenario google-adk` reproduces the committed `data.json` exactly. The `genai_span_name_format` findings stay until the rego in the conformance repo learns refinement span names - that is the follow-up we agreed on, not a blocker.

agent-framework needs the same matchers; I left it alone since its three tools mirror these.

model/gen-ai/registry.yaml
lmolkova, line 1053:
```suggestion
    note: |
      > [!WARNING]
      > This attribute may contain sensitive information.
```

nit: Use `> [!WARNING]` (uppercase) to match conventions, and remove the leading blank line under `note: |`. Also check lines 1062-1063 for the duplicate blank line.

docs/gen-ai/gen-ai-agent-spans.md
lmolkova: my motivation was to say that there are special tools that should be treated differently. Known cases of such tools are skills, handoffs, and MCP. MCP, which we don't yet model as tool execution refinement (bug rather than intentional) is not necessarily agentic.  

But this goes beyond the scope of this PR, so "Instrumentations for agentic frameworks SHOULD follow additional rules defined..." sounds good

model/gen-ai/spans.yaml
```diff
@@ -589,6 +589,13 @@ spans:
       are encouraged to follow this semantic convention for tools invoked by their
       own code and to manually instrument any tool calls that automatic
       instrumentations do not cover.
+
+      Though [AgentSkills](https://agentskills.io) are usually executed as tools,
+      due to their specialized nature, instrumentations for agentic frameworks
+      SHOULD follow additional [conventions and models for GenAI agents](/docs/gen-ai/gen-ai-agent-spans.md#execute-tool-span).
+      The instrumentations SHOULD distinguish generic tool from specialized ones,
+      using framework specific tool names or heuristics, and
+      they SHOULD NOT record two different spans for one call.
```
lmolkova, line 598: maybe?

```suggestion
      Some tools are specialized, such as [Agent Skills](https://agentskills.io)
      that agentic frameworks expose as tools. Instrumentations SHOULD distinguish
      generic tools from specialized ones using framework-specific tool names or
      other heuristics and record the applicable
      [refinement](/docs/gen-ai/gen-ai-agent-spans.md#agent-skills).
      They SHOULD NOT record two different spans for one call.
```

model/gen-ai/spans.yaml
```diff
@@ -863,3 +870,145 @@ span_refinements:
...
+
+      - [Google ADK](https://adk.dev/skills/) `load_skill`;
+      - [OpenAI Agents](https://openai.github.io/openai-agents-python/sandbox/guide/)
+        `load_skill` for a lazy skill source;
+      - [Strands Agents](https://strandsagents.com/docs/user-guide/concepts/plugins/skills/)
+        `skills`;
+      - [Agno](https://docs.agno.com/skills/overview)
+        `get_skill_instructions`.
+
+      Otherwise, the generic `execute_tool` span applies.
```
lmolkova, line 893: nit:

```suggestion
```

having second thoughts about it - otherwise another refinement or generic execute tool can apply, but also it does not really convey any important information

(same for other refinements)

model/gen-ai/spans.yaml
```diff
@@ -863,3 +870,145 @@ span_refinements:
...
+
+        The skill name SHOULD NOT be appended without the resource name.
+    note: |
+      This refinement applies when a tool represents command execution,
+      including the execution of skill's script.
+
+      Examples include:
+
+      - [OpenAI Agents](https://openai.github.io/openai-agents-python/sandbox/guide/)
+        `exec_command`, which runs both general commands and scripts from
```
lmolkova, line 971: would it be possible to add reference scenario for this one or langchain to demonstrate process.* attributes on this refinements? Not for skills but for generic command

lmolkova (approved):
still several very cosmetic changes, but LGTM in general!

model/gen-ai/spans.yaml
filip-kopyt: I will apply this suggestion, but I feel like this is slightly contradictory with [one of the previous comments](https://github.com/open-telemetry/semantic-conventions-genai/pull/498#discussion_r3984546101). Also I  updated the `[refinement]` link to `gen-ai-agents-spans.md#execute-tool-span`, the suggested version reads like all specialized tools should follow agent skills refinements.

model/gen-ai/registry.yaml
filip-kopyt: > nit: Use > [!WARNING] (uppercase) to match conventions

FYI, grepping the registry suggests it is close to even split for both versions.

model/gen-ai/spans.yaml
filip-kopyt: Done

model/gen-ai/spans.yaml
filip-kopyt: Done

filip-kopyt:
CI failing due to github issues. `make generate-all` doesn't report changes

model/gen-ai/spans.yaml
RKest: Out of the four:
1. `execute_tool {gen_ai.tool.name}`
2. `load_skill {gen_ai.skill.name}`
3. `execute_tool {gen_ai.tool.name} {gen_ai.skill.name}` 
4. `{gen_ai.tool.name} {gen_ai.skill.name}`

(going ~backwards)

3. I like the least. I feel Filip's intuition for "too long" names is correct.
4. Means the name of the span can be different across frameworks. It may make telemetry consumers' lives more difficult than it needs be. If we wanted the tool name for skill loading to be consistent in this case, it would require changing this in the framework code, and it may change how the framework behaves (even though it would be a minor should-be-inconsequential change, but you never know with LLMs).

Between 1 & 2, it's a toss-up for me. The latter is a backwards incompatible, but now would be the best time to have a backwards incompatible change.

Issue #477, opened 2026-08-26
Agent vs workflow


lmolkova:
**Problem**:

**TL;DR: Callers and consumers treat agentic workloads as black boxes; internal architecture refactors should not break alerts and and dashboards for you or your downstream.**

We have invoke_agent and invoke_workflow operations. They are defined this way to match agent framework API shape:

- invoke_agent is used to cover something like `agent.run(messages)`
- invoke_workflow covers crews, graphs, workflows - any API that agentic framework exposes that's more complex than invoking a single agent.

Both workflow and agents can be nested under each other within one process, but usually we'd expect workflow to encompass multi-agent run.

From inside the application / framework the distinction makes sense - different API shapes, slightly different input / output params.
From outside the application it stops making sense:
 
- hosted / managed agents: the fact that the hosted agent is a complex combination of tools / subagents / workflows / transformations / validations is not important to the caller. Callers talk to it as if it was a single agent. If internal implementation changes from top-level-workflow to top-level-agent API, nothing needs to change for the caller; but with the current design telemetry would change completely

- harnesses: we don't know what's inside coding agents we use, but we're comfortable calling them agents. We don't care how they are modeled internally (graph, orchestrator, dynamic or static workflow, raw llm calls with tools, etc), we'd expect the same telemetry from them regardless

- no-code agent builders like [Microsoft Copilot Studio](https://www.microsoft.com/en-us/microsoft-365-copilot/microsoft-copilot-studio) or [Gemini Agent Studio](https://docs.cloud.google.com/gemini-enterprise-agent-platform/agent-studio/design-agents) - same thing - their internal implementation details are irrelevant and are subject to change.

----
**Proposal**:

1. Merge invoke_agent and invoke_workflow span and metric definitions. 
2. Allow operation name to be either `invoke_agent` or `invoke_workflow` (similarly to inference where we use `chat` or `generate_content` depending on the provider)
   - Long term: once span type becomes available over OTLP https://github.com/open-telemetry/opentelemetry-specification/pull/5233, we can relax operation name further and allow it to be anything idiomatic to the API and framework
3. Use span hierarchy and attributes to differentiate nested agents from the main one - see https://github.com/open-telemetry/semantic-conventions-genai/pull/385

DylanRussell:
sounds reasonable.. are the expected attributes on the 2 spans already exactly the same ?

RKest:
[...]

PR #500, opened 2026-09-04
 Add `gen_ai.skills` to the invoke agent internal span

filip-kopyt:
### Description

An opt-in attribute listing the skills whose instructions are in an invocation's context, each flagged with whether it is still held in full or has been replaced by a summary, recorded as `[{"name": "code-review", "compacted": true}, {"name": "pdf-processing", "compacted": false}]`.

Builds on #499.

### Motivation

Explaining context size and model behavior. The `execute_tool` attributes from #498 record the act of loading a skill. They say nothing about the invocations that follow, and those are the ones running with the instructions in context. When an agent behaves oddly ten turns after the load, the question is which skill bodies were in that prompt and whether compaction had already swapped them for a summary. That is also most of the answer to why a context grew.

`compacted` is reported for every entry, unlike `gen_ai.conversation.compacted`, which instrumentations set only when true. The difference is deliberate. The skill set is closed, so an instrumentation that can determine compaction for one skill can determine it for all of them, and an omitted flag would be ambiguous where an explicit `false` is not.

Opt-in, because the value grows with the number of loaded skills and carries skill names that some deployments treat as internal.

### Prototype

The ADK scenario reads this from session state rather than from bookkeeping it keeps for itself. A `load_skill` function response says whether the name resolved, and a compaction event declares the timestamp range its summary replaced, so a load whose response event falls inside that range is in context in compacted form. The conversation runs long enough for compaction to fire, so the final invocation observes one skill compacted and another still held in full, which is the state this attribute exists to describe.

### Checklist


model/gen-ai/metrics.yaml
```diff
@@ -248,6 +248,112 @@ metrics:
...
+      framework's load-skill tool: one sample each time the agent activates a
+      skill and reads its instructions into context. Counts every load the
+      instrumentation observes, including the ones that failed and the ones that
+      resolved a skill already loaded in the same invocation.
+
+      A load that resolved no skill is counted with `error.type` and without
+      `gen_ai.skill.name`, so a name the model invented cannot enter the
+      dimension. The name the call asked for is on the span.
+
+      `gen_ai.execute_tool.duration` is not dimensioned by `gen_ai.skill.*`, so
```
lmolkova, line 266: why not? a good reason not to would be if more than one skill can be loaded with one tool call, otherwise it seems straightforward to reuse existing metric and refine it similarly to how we refine span.

model/gen-ai/metrics.yaml
```diff
@@ -248,6 +248,112 @@ metrics:
...
+      its count cannot answer which skills an agent loads; keeping the skill
+      dimensions on a separate instrument holds that cardinality off the
```
lmolkova, line 268: is cardinality a real problem? how many skills do we expect on one agent?

model/gen-ai/registry.yaml
```diff
@@ -1039,3 +1039,80 @@ attributes:
...
+      call or instrumented library state. It may be a skill-relative path when
+      the framework uses paths.
+    examples: ["references/review_policy.md", "assets/report_template.html"]
+    stability: development
+  - key: gen_ai.skill.script.exited_with_error
+    type: boolean
+    brief: Whether the skill script terminated with a non-zero exit code.
+    note: >
+      A low-cardinality companion to `process.exit.code`, for use as a
+      metric dimension where the exit code itself would be unbounded.
```
lmolkova, line 1095: we don't usually record attributes that can be trivially derived from other attributes

model/gen-ai/metrics.yaml
```diff
@@ -248,6 +248,112 @@ metrics:
...
+    instrument: histogram
+    unit: "{skill}"
+    stability: development
+    attributes:
+      - ref: gen_ai.agent.name
+        requirement_level: recommended
+    annotations:
+      code_generation:
+        metric_value_type: int
+  - name: gen_ai.invoke_workflow.skill.loads
```
lmolkova, line 336: is workflow configured with a list of skills? I'd prefer not to add this one until we figure out https://github.com/open-telemetry/semantic-conventions-genai/issues/477

model/gen-ai/metrics.yaml
```diff
@@ -248,6 +248,112 @@ metrics:
...
+        requirement_level:
+          conditionally_required: If the name resolves to a skill available to the agent.
+    annotations:
+      code_generation:
+        metric_value_type: int
+  - name: gen_ai.skill.script.executions
+    requirement_level: recommended
+    brief: The number of times a skill's script was executed.
+    note: |
+      Recorded at the same boundary as the `execute_tool` span for the
```
lmolkova, line 287: I believe it should be tool metric refinement - same span, same instrumentation point, same metric with new dimenstions

model/gen-ai/metrics.yaml
```diff
@@ -248,6 +248,112 @@ metrics:
...
+      framework's run-skill-script tool.
+
+      `error.type` and `gen_ai.skill.script.exited_with_error` report different
+      outcomes: the first says the execution itself failed, the second says the
+      script ran and exited non-zero. An execution that never reached the script
+      has no `gen_ai.skill.script.exited_with_error`, and neither does one whose
+      framework runs the script without reporting the status it exited with. That
+      keeps the script's own outcome low-cardinality; the exit code itself is
+      recorded on the span as `process.exit.code`.
```
lmolkova, line 296: pretty much all semantic conventions put domain-specific status code on metrics. In practice the cardinality is not that high - it's likely that a specific script or program returns just a few possible exist codes.

So the semconv way would be to put `process.exit.code` on the metric and remove `gen_ai.skill.script.exited_with_error` altogether. Moreover, the value of `error.type` usually would be set to exit code when operation fails because this code was received,

If exit code cardinality is a concern in practice, we can consider some other measures (e.g. buckets)

model/gen-ai/gen-ai-skills.json
```diff
@@ -0,0 +1,32 @@
...
+                },
+                "compacted": {
+                    "description": "Whether the skill's instructions have been replaced by a summary.",
+                    "title": "Compacted",
+                    "type": "boolean"
+                }
+            },
+            "required": [
+                "name",
+                "compacted"
```
lmolkova, line 20: it can't be required - reference scenario reconstructs it by scanning the entire session event history for load_skill calls and cross-referencing timestamps with EventCompaction actions. Generic instrumentation cannot do this.

It seems without it this attribute becomes not quite useful - a list of bare skill names in a map.

model/gen-ai/registry.yaml
```diff
@@ -1039,3 +1039,80 @@ attributes:
...
+    examples: [true]
+    stability: development
+  - key: gen_ai.skills
```
lmolkova, line 1098: In my mind this is similar to `gen_ai.tool.definitions` - is there information about skills that won't be in the tool definitions? should we put this info into `gen_ai.skills` or better `gen_ai.skill.definitions`? 
This is where I'd expect to see skill names, descriptions, URIs and maybe even info about scripts and contents of the skill.md (opt-in).

That would naturally capture skill names, descriptions, source URIs, and optionally script definitions or contents.

If the goal is specifically to track dynamically loaded skills per turn or agent, that sounds like `gen_ai.agent.loaded_skills`. Given that individual skill compaction is not exposed by SDK runtimes today, I suggest keeping that tracking custom/experimental to ADK until a common pattern emerges across frameworks.

PR #518, opened 2026-09-15
Allow conversation ID on tool execution spans

### LINKED ISSUE #507: Allow `gen_ai.conversation.id` on `execute_tool` spans
### Problem

`gen_ai.conversation.id` is allowed on `gen_ai.inference.client`, `gen_ai.invoke_agent.client`, `gen_ai.invoke_agent.internal`, and `gen_ai.invoke_workflow.internal`. It is not allowed on `gen_ai.execute_tool.internal`.

So filtering telemetry by conversation returns an agent's model calls but not its tool calls. Which conversations hit failing tool calls, how much of a conversation's latency was tools rather than inference, and which tools a conversation actually ran are all unanswerable from the identifier, even though tool execution is where agent turns commonly break.

Without the attribute, a consumer has to recover the identifier from an ancestor span. How many levels up it sits depends on the framework, so grouping tool executions by conversation is a traversal rather than a `GROUP BY`, and most query surfaces cannot express one. When the application drives the tool loop itself, the identifier is on a sibling `chat` span and no ancestor carries it at all.

### Motivating example

One turn of a conversation:

```text
└── invoke_agent travel_planner
      gen_ai.conversation.id = conv_abc
      ├── chat
      │     gen_ai.conversation.id = conv_abc
      └── execute_tool search_flights
            cannot record conv_abc
```

Filtering on `gen_ai.conversation.id = conv_abc` returns the agent and inference spans and no tool execution.

### Proposed change

Add `gen_ai.conversation.id` to `gen_ai.execute_tool.internal`, with the requirement level used verbatim on the four spans that already carry it:

```yaml
      - ref: gen_ai.conversation.id
        requirement_level:
          conditionally_required: If and only if the instrumented library has one readily available, or the user application provides one through OpenTelemetry context or library-specific mechanisms.
```

Existing guidance still applies. Instrumentations must not synthesize a fallback from a trace ID, a UUID, or a request hash when the library exposes none, and tools invoked outside a conversation omit the attribute.

Spans only. `gen_ai.conversation.id` is high cardinality and does not belong on the `execute_tool` duration metric.

### Capturability

Google ADK, at the pinned `google-adk==2.8.0`, already creates this span. `record_tool_execution(tool, agent, function_args, invocation_context)` opens `execute_tool {tool.name}` (`telemetry/_instrumentation.py:495-507`) and forwards `invocation_context` to `trace_tool_call` (`:529-534`), which already reads that object to stamp the span, including `gen_ai.agent.name` from `invocation_context.agent` (`telemetry/tracing.py:273-276`). `invocation_context.session.id` is one attribute access away on the same object. ADK already maps that value to `gen_ai.conversation.id` on its other spans: `runners.py:602-604` passes `conversation_id=session_id` into `record_invocation`, and `telemetry/node_tracing.py:151` sets it from `context.session.id`. The identifier is the same one; only the span it may land on differs. The reference scenario reads it as `tool_context.session.id`, since `ToolContext` aliases `google.adk.agents.context.Context` and `session` is a public property.

OpenAI Assistants. `thread_id` is a required keyword of `submit_tool_outputs` and a required field on `Run`, so an application cannot execute the tool and continue the run without holding it. The scenario already records `thread.id` as `gen_ai.conversation.id` on the agent span, and emits the `execute_tool` span inside that same run loop.

### Related issues

- #409 allowed this attribute on `invoke_workflow` spans, for the same reason. This asks for the same thing one span over.
- #356 marks the root span of each conversation turn. It lets a backend list a conversation's turns, where this lets it see what ran inside one, so the two are complementary rather than overlapping.
- #91, #300 and #94 also want parent-scoped context on child spans, and are open on whether instrumentation should set it or baggage should carry it. This proposal does not wait on that: `gen_ai.conversation.id` is already a per-span attribute on four spans, and adding it to a fifth needs no new mechanism and no metric dimension.

sfc-gh-zeningchen:
## Description

Allow `gen_ai.conversation.id` on `execute_tool` spans, with the requirement level used verbatim on the four spans that already carry it. Spans only — it is not added to the `gen_ai.execute_tool.duration` metric.

Fixes #507.

## Motivation

**User journey:** incident triage. Filtering telemetry by conversation returns an agent's model calls but not its tool calls, so which conversations hit failing tool calls, and how much of a conversation's latency was tools rather than inference, cannot be answered from the identifier. Recovering it from an ancestor span is a traversal at a framework-dependent depth rather than a `GROUP BY`, and when an application drives the tool loop itself the identifier sits on a sibling `chat` span, so no ancestor carries it at all.

**Prior art:** Google ADK creates the `execute_tool` span itself, already stamps `gen_ai.agent.name` from the same `invocation_context` that holds the session, and already maps that `session.id` to `gen_ai.conversation.id` on its other spans. The OpenAI Assistants API exposes the thread id as a required parameter of the same run loop that executes tools.

## Prototype

Two reference scenarios now record the attribute on their `execute_tool` span from a library-owned identifier: `google-adk` from `tool_context.session.id`, and `openai-assistants` from the run loop's `thread_id`. Both already recorded the same value on their agent spans, so the tool span now groups with them. Coverage report, docs, `data.json` and a Towncrier fragment regenerated.

Validation: `uv run run-scenario google-adk`, `uv run run-scenario openai-assistants`, `make check-policies`, `make generate-all`.

## Checklist


model/gen-ai/spans.yaml
```diff
@@ -608,6 +608,9 @@ spans:
         sampling_relevant: true
       - ref: gen_ai.agent.name
         sampling_relevant: true
+      - ref: gen_ai.conversation.id
+        requirement_level:
+          conditionally_required: If and only if the instrumented library has one readily available, or the user application provides one through OpenTelemetry context or library-specific mechanisms.
```
lmolkova: maybe just 

```suggestion
          conditionally_required: If available.
```

?

lmolkova (approved):
Thank you!

model/gen-ai/spans.yaml
sfc-gh-zeningchen: sounds good! do we need to apply this to the other spans as well?

singankit (approved):

PR #525, opened 2026-09-18
feat(gen-ai): require provider-supplied tool call ID on execute_tool (#489)

### LINKED ISSUE #489: gen_ai.tool.call.id: should it carry only a provider-supplied identifier?
The execute_tool span lists `gen_ai.tool.call.id` as `Recommended` "If available", described as "The tool call identifier." (`docs/gen-ai/gen-ai-spans.md`, commit `ac46a5d`.)

Neither the level nor the description says *whose* identifier it is, or what makes one "available." I think that matters, because the attribute's only real use is correlating a tool execution back to the model request that asked for it ,and that only works if the id came from the model.

**Under the current text, a framework-synthesized id is conformant.** A framework that doesn't receive an id from the provider can generate its own, put it in this attribute, and be within spec. But such an id correlates nothing: it's unique to the process that made it up, and the consumer can't tell it apart from an attested one.

**This isn't hypothetical - a framework just hit it.** CrewAI is adding tool call ids to its telemetry events (crewAIInc/crewAI#7189). It needs a local fallback id to pair assistant and tool messages internally, and it also emits ids for telemetry. Having no guidance from the spec, it decided the two must not be the same thing, and wrote that into a docstring:

> Return a provider-supplied tool call ID, if one is available.
>
> CrewAI still needs a local fallback ID to pair assistant and tool messages, but that fallback is not a stable provider correlation key for telemetry.

Its events now carry the provider id or `None`, never the fallback. I think that's the right call ,but a framework had to reason it out independently, and a different framework could reasonably decide the opposite and still conform.

**Why the fallback specifically is a poor correlation key**, as an illustration of the failure mode: CrewAI's is `f"call_{id(tool_call)}"`. In CPython, `id()` returns the object's address, which is reused once the original object is collected - so a later tool call can be assigned an id identical to an earlier, already-freed one. A consumer treating that as a key mis-pairs in precisely the retry and parallel-call cases the attribute exists to disambiguate.

**Proposal.** Clarify in the attribute description that `gen_ai.tool.call.id` carries the identifier supplied by the model or provider, and should be omitted when none was supplied - making "if available" mean "if the provider gave you one." That's a description change, no new attribute.

If there's a reason to allow synthesized ids, then the alternative is a companion attribute recording provenance so consumers can weight it. But I'd expect omission to be simpler and more useful.

Happy to write the YAML for either.

Disclosure: I maintain an open-source trace analyzer that consumes these spans. It distinguishes explicit ids from mined and sequential fallbacks, and can't tell a synthesized id from an attested one ,which is what surfaced this. 
Raising it because the ambiguity affects any consumer doing causal attribution.

1aifanatic:
## Description

Fixes #489.

This PR clarifies the semantics and requirement level of `gen_ai.tool.call.id`:
- Clarifies in the attribute definition that `gen_ai.tool.call.id` MUST carry the identifier generated and supplied by the model or provider, and MUST NOT be populated with synthesized fallback identifiers (such as object memory addresses or sequential indices).
- Promotes the attribute on `execute_tool` spans from `Recommended: If available.` to `Conditionally Required: If the model or provider supplied a tool call identifier.`.

## Motivation

- **User journey:** Causal attribution, auditability, and incident triage in multi-step and parallel agentic execution. In enterprise agentic systems, correlating an `execute_tool` span back to the exact model generation that requested it is essential for tracing decision causality, verifying tool parameters, and auditing actions against system-of-record updates.
- **Why absence must be load-bearing:** At `Recommended`, omitting the attribute is ambiguous: it could mean the provider didn't supply an ID, the operator disabled recommended attributes, or the instrumentation hasn't implemented it. Promoting to `Conditionally Required: If the model or provider supplied a tool call identifier.` gives absence deterministic meaning: an omitted attribute reliably indicates the provider did not supply one, eliminating the temptation for frameworks to manufacture synthetic fallback IDs.
- **Failure mode of synthetic IDs:** When frameworks synthesize local fallback IDs—such as CrewAI's previous `f"call_{id(tool_call)}"`—CPython address reuse after garbage collection causes collision across retries and parallel tool calls, silently corrupting causal joins in downstream observability platforms.
- **Prior art:** CrewAI (crewAIInc/crewAI#7189) explicitly documented that local pairing fallbacks must not be emitted as provider correlation keys in telemetry. The condition pattern mirrors standard semconv precedents like `http.route`.

## Prototype

10 existing reference scenarios in this repository already emit `gen_ai.tool.call.id` (`agent-framework`, `autogen`, `google-adk`, `google-genai`, `langchain`, `llamaindex`, `openai-agents`, `openai-assistants`, `pydantic-ai`, `vertexai`). In all cases, they extract and emit the model/provider-attested identifier directly from the provider response. All scenarios and Weaver policy checks conform cleanly.

## Checklist

- [x] Towncrier fragment added under `changelog.d/489.clarification.md`


SAMurai-16:
lgtm

Changing `gen_ai.tool.call.id` to `Conditionally Required` provides the necessary strictness around absence semantics:

- `MUST` be populated if the LLM/provider returns a tool call ID.
- `MUST NOT` be synthesized or fabricated when absent.

This resolves the issue of frameworks generating colliding synthetic IDs and gives trace backends deterministic behavior to rely on.

PR #529, opened 2026-09-21
Keep explicit zero output tokens in the Google GenAI usage attributes

AgentGymLeader:
## Problem

In `reference/scenarios/google-genai/scenario.py`, `_usage_attributes` folded the output token count into a truthiness check:

```python
output_tokens = (um.candidates_token_count or 0) + reasoning_tokens
if output_tokens:
    attrs["gen_ai.usage.output_tokens"] = output_tokens
```

An explicit `candidates_token_count == 0` (and/or `thoughts_token_count == 0`) is therefore treated the same as usage the SDK did not report at all: the `gen_ai.usage.output_tokens` attribute is never set. Any metric derived from that attribute loses the zero observation. Recording the observations `0` and `10` through the usage metric recorder yields an output-token histogram with count 1 / sum 10 instead of count 2 / sum 10 — the sum is right, the count and distribution are not.

The `google-genai` SDK (2.22.0) keeps presence: `GenerateContentResponseUsageMetadata.candidates_token_count` / `thoughts_token_count` are `Optional[int]` with default `None`; an omitted JSON field arrives as `None` and an explicit `"candidatesTokenCount": 0` arrives as `0`. The reference scenario collapsed that distinction.

## Fix

Only usage the SDK did not report (`None` for both fields) leaves the attribute unset. An explicit zero is reported as `0`; positive values and the reasoning addition are unchanged.

```python
if candidates_tokens is not None or thoughts_tokens is not None:
    attrs["gen_ai.usage.output_tokens"] = (candidates_tokens or 0) + reasoning_tokens
```

One incidental difference: `candidates_token_count` is now read with `getattr(..., None)` like the neighbouring fields, so a usage object without that attribute is treated as "not reported" instead of raising. The SDK model always carries the attribute, so no observable change for real responses.

Out of scope on purpose (same truthiness pattern, separate follow-ups if wanted): `gen_ai.usage.reasoning.output_tokens` (still omitted when 0), the input-token branch, and the streaming path around `last_chunk.usage_metadata.candidates_token_count`.

## Regression test

`reference/scenarios/google-genai/test_usage_attributes.py` covers `(0, 0)`, `(0, None)`, `(None, 0)`, `(10, None)`, `(10, 5)` → the expected value, `(None, None)` → attribute absent, and reasoning added once. It imports `scenario.py` from the scenario's own environment (the shared `reference/tests` environment does not carry OpenTelemetry):

```bash
cd reference/scenarios/google-genai
MOCK_LLM_URL=http://unused uv run --with pytest pytest -q test_usage_attributes.py
```

On unpatched `main` 3 of 7 cases fail; with this change all 7 pass. `ruff check` and `ruff format --check` pass with the repository configuration.

Related: #374 (usage metrics; reads the attribute with `is not None` and is unaffected by this change — the defect is upstream of it).


AgentGymLeader:
CI doesn't run any of the reference pytest files today (reference/tests/test_metrics.py is in the same situation), so I kept this one runnable from the scenario's own environment rather than adding CI wiring in the same PR. If you'd like it in CI, I can add a pytest step to the reference-python-lint job as a follow-up.

reference/scenarios/google-genai/scenario.py
```diff
@@ -118,9 +122,8 @@ def _usage_attributes(um):
         attrs["gen_ai.usage.input_tokens"] = um.prompt_token_count + tool_use_tokens
     # Google reports "thoughts" (reasoning) tokens separately from the visible
     # candidates, so include them in the output token total.
-    output_tokens = (um.candidates_token_count or 0) + reasoning_tokens
-    if output_tokens:
-        attrs["gen_ai.usage.output_tokens"] = output_tokens
+    if candidates_tokens is not None or thoughts_tokens is not None:
```
aabmass, line 125: `prompt_token_count` above (line 120), `total_output_tokens` (line 243), and streaming `candidates_token_count` (line 295) still use truthy checks.

aabmass (commented):
1. The spec doesn't say whether `0` should be recorded or omitted yet, and #374 is reworking the token usage metrics and attributes. Should we decide how to handle `0` there first rather than encoding it in one scenario?

2. Regarding the PR description:
> The `google-genai` SDK (2.22.0) keeps presence: `GenerateContentResponseUsageMetadata.candidates_token_count` / `thoughts_token_count` are `Optional[int]` with default `None`; an omitted JSON field arrives as `None` and an explicit `"candidatesTokenCount": 0` arrives as `0`.

That's only true of the client-side Pydantic model, not the Gemini API wire payload. In the public API definition ([`google/ai/generativelanguage/v1beta/generative_service.proto` (`UsageMetadata`)](https://github.com/googleapis/googleapis/blob/f822db2f84fd93f8a3a59df23222fd3ab39baa0b/google/ai/generativelanguage/v1beta/generative_service.proto#L628-L650)), `int32 candidates_token_count = 2;` and `int32 thoughts_token_count = 10;` are plain `proto3` scalars (unlike `optional int32 max_output_tokens`, etc. earlier in the same file). Because `proto3` JSON serialization omits default `0` fields on the wire, a real Gemini response with `0` output tokens returns `{"promptTokenCount": 5, "totalTokenCount": 5}` with `candidatesTokenCount` omitted, which the Python SDK deserializes as `(candidates_token_count=None, thoughts_token_count=None)`. So I think a real `0`-output response from the API still hits the `(None, None)` path here rather than `(0, None)`.

reference/scenarios/google-genai/test_usage_attributes.py
```diff
@@ -0,0 +1,67 @@
+"""Regression tests for `_usage_attributes` in the Google GenAI reference scenario.
```
aabmass, line 1: `reference/scenarios/` only runs `uv run run-scenario` in CI, so a standalone `pytest` file in the scenario folder won't run.

aabmass: Ah I see it was in your comment. Is there any precedent for this in the repo already? I don't have a strong preference beyond making sure it does't bit rot from not being run.

reference/scenarios/google-genai/scenario.py
AgentGymLeader: Same pattern. Once #374 settles it I'll fix all of them together.

reference/scenarios/google-genai/test_usage_attributes.py
AgentGymLeader: Only `reference/tests/test_metrics.py`, and CI doesn't run that either. Once the behaviour is decided, the mock-driven `run-scenario` path is the right place to pin it; I'll drop this file with the draft.

AgentGymLeader:
Right — `GenerateContentResponse.UsageMetadata.candidates_token_count` is a plain proto3 `int32`, so a real zero arrives as `(None, None)` and this diff only matters for inputs that carry an explicit 0 (the conformance mock does, the API doesn't). Whether an omitted count with `usageMetadata` present means 0 or "not reported" is a spec question; let's settle it in #374. Converting this to a draft until then.

